PR #1, opened 2014-09-16
[ADD] module price_multicompany

bealdav:


sale_check_multicompany/__openerp__.py
```diff
@@ -0,0 +1,50 @@
...
+#
+##############################################################################
+
+{
+    'name': 'sale_check_multicompany',
+    'version': '0.1',
+    'category': '',
+    'sequence': 10,
+    'description': """
+Check if datas in quotation are linked to the same company
```
rvalyi: data is already plural -> no 's' (singular is datum)

price_multicompany/__openerp__.py
```diff
@@ -0,0 +1,52 @@
...
+
+{
+    'name': 'price_multicompany',
+    'version': '0.1',
+    'category': '',
+    'sequence': 10,
+    'summary': "Changes for multicompany prices",
+    'description': """
+This module should be used if your companies have their currencies non shared.
+Currencies can't be shared because 'res.currency.rate' haven't company_id field
```
rvalyi: Currencies cannot be shared because 'res.currency.rate' doesn't have a company_id field

price_multicompany/currency.py
```diff
@@ -0,0 +1,41 @@
...
+class ResCurrency(orm.Model):
+    _inherit = 'res.currency'
+
+    def name_get(self, cr, uid, ids, context=None):
+        """ If currencies are not shared and they have the same name,
+            then display have to distinguish them"""
+        result = []
+        for currency in self.browse(cr, uid, ids, context):
+            cpny = currency.company_id and currency.company_id.name
+            result.append((currency.id, currency.name + ' (' + cpny + ')'))
```
rvalyi: doing a string substitution like  "currency.name (%s)" % (cpny,) is more efficient than such concatenation. It's a good habit to take. See http://stackoverflow.com/questions/376461/string-concatenation-vs-string-substitution-in-python

pedrobaeza:
I think there are some overlaps here. We have some alternatives for this that we have to clean:
- https://github.com/OCA/product-attribute/tree/7.0/product_multi_company
- https://github.com/OCA/product-attribute/tree/7.0/product_multi_price
- Another one from @jgrandguillaume about product price history that I can't find now (we need an indexer...) and also add multi-company property.

bealdav:
sale_check_multicompany might be in a better place in OCA/sale-workflow.
@pedrobaeza it concerns sale, not product
What do you think ?

pedrobaeza:
Well, this module covers multi-company, so it can fit here, and the models covered (res.currency, product.price.type, etc) are very general (not only for sale).

The other topic that I raised concerned me more about overlappings. The third module I didn't remember was product_price_history in margin-analysis repo (https://github.com/OCA/margin-analysis/tree/7.0/product_price_history).

bealdav:
Hi @pedrobaeza 
I have renamed my module. Now, the use case is more precise,  i think you'll understand there no overlaps. Feature in my module should be use without features of the modules you mentionned

Thanks

sale_check_multicompany/sale.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class SaleOrder(orm.Model):
+    _inherit = 'sale.order'
+
+    def _check_pricelist(self, cr, uid, ids):
+        for elm in self.browse(cr, uid, ids):
+            if elm.pricelist_id:
+                if isinstance(elm.company_id, orm.browse_null) \
+                        or not elm.company_id:
```
guewen: Why do you check if the type is orm.browse_null? 
`browse_null.__nonzero__` always returns False so `elm.company_id` will be Falsy if it is a `browse_null`.

sale_check_multicompany/sale.py
```diff
@@ -0,0 +1,39 @@
...
+                    company_id = self.pool['res.company']._company_default_get(
+                        cr, uid, 'sale.order')
+                    company = self.pool['res.company'].browse(
+                        cr, uid, company_id)
+                else:
+                    company = elm.company_id
+                if elm.pricelist_id.company_id.id != company.id:
+                    raise orm.except_orm(
```
guewen, line 29: This exception is not translated. By the way, I think the context should be added to the arguments.

unshared_currency_multicompany/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+##############################################################################
+#
+#  licence AGPL version 3 or later
+#  see licence in __openerp__.py or http://www.gnu.org/licenses/agpl-3.0.txt
+#  Copyright (C) 2014 Akretion (http://www.akretion.com).
+#  @author David BEAL <david.beal@akretion.com>
+#
+##############################################################################
+
+import currency
```
guewen: `from . import currency`

unshared_currency_multicompany/__openerp__.py
```diff
@@ -0,0 +1,53 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Unshared Currency Multicompany',
+    'version': '0.1',
+    'category': '',
```
guewen: Isn't the category mandatory?

unshared_currency_multicompany/currency.py
```diff
@@ -0,0 +1,41 @@
...
+class ResCurrency(orm.Model):
+    _inherit = 'res.currency'
+
+    def name_get(self, cr, uid, ids, context=None):
+        """ If currencies are not shared and they have the same name,
+            then display have to distinguish them"""
+        result = []
+        for currency in self.browse(cr, uid, ids, context):
+            cpny = currency.company_id and currency.company_id.name
+            result.append((currency.id, '%s (%s)' % (currency.name, cpny)))
```
guewen: If a currency has no company, its name will be something like `Euro (False)`.

sale_check_multicompany/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+##############################################################################
+#
+#  licence AGPL version 3 or later
+#  see licence in __openerp__.py or http://www.gnu.org/licenses/agpl-3.0.txt
+#  Copyright (C) 2014 Akretion (http://www.akretion.com).
+#  @author David BEAL <david.beal@akretion.com>
+#
+##############################################################################
+
+import sale
```
guewen: `from . import sale`

guewen:
What do you mean by

> I have renamed my module. Now, the use case is more precise, i think you'll understand there no overlaps. Feature in my module should be use without features of the modules you mentionned

? It is incompatible with the aforementioned modules?

bealdav:
@guewen thanks for your really interesting inline comments. 
Sorry for my bad english i would like say:  Feature in my module CAN be use without features of the modules you mentionned

pedrobaeza:
OK, I understand now the purpose of the module, but there are two modules in this PR. Please split them in separate PRs.

bealdav:
replaced by #4 and #5

PR #4, opened 2014-12-30
[ADD] module unshared_currency

bealdav:
Comes from this PR: split in 2 modules
https://github.com/OCA/multi-company/pull/1

### DIFF
--- a/unshared_currency_multicompany/currency.py
+++ b/unshared_currency_multicompany/currency.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#  licence AGPL version 3 or later
+#  see licence in __openerp__.py or http://www.gnu.org/licenses/agpl-3.0.txt
+#  Copyright (C) 2014 Akretion (http://www.akretion.com).
+#  @author David BEAL <david.beal@akretion.com>
+#
+##############################################################################
+
+from openerp.osv import orm, fields
+
+
+class ResCurrency(orm.Model):
+    _inherit = 'res.currency'
+
+    def name_get(self, cr, uid, ids, context=None):
+        """ If currencies are not shared and they have the same name,
+            then display have to distinguish them"""
+        result = []
+        for currency in self.browse(cr, uid, ids, context):
+            cpny_info = ''
+            if currency.company_id:
+                cpny_info = ' (%s)' % currency.company_id.name
+            result.append((currency.id, '%s%s' % (currency.name, cpny_info)))
+        return result
+
+
+class ProductPriceType(orm.Model):
+    _inherit = 'product.price.type'
+
+    _columns = {
+        'currency_id': fields.property(
+            'res.currency',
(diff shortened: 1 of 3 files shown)


bealdav:
Failed solved by this PR https://github.com/OCA/multi-company/pull/6.
Is somebody with rights could rebuild this PR. How to do it ?
Thanks

pedrobaeza:
You have to rebase your PR.

bealdav:
I do not understand how runbot works, and how see what failed. :scream_cat:

pedrobaeza:
You don't have to take into account runbot. It's not working properly. Just take Travis information.

pedrobaeza:
A better name would be _base_currency_multicompany_

bealdav:
There 2 ways to use currencies in multi-company environnement:
- unshare currencies: the purpose of this module
- share currencies: see https://github.com/odoo/odoo/issues/641

This module deal with unshared currencies only.

unshared_currency_multicompany/currency.py
```diff
@@ -0,0 +1,43 @@
...
+    _inherit = 'res.currency'
+
+    def name_get(self, cr, uid, ids, context=None):
+        """ If currencies are not shared and they have the same name,
+            then display have to distinguish them"""
+        result = []
+        for currency in self.browse(cr, uid, ids, context):
+            cpny_info = ''
+            if currency.company_id:
+                cpny_info = ' (%s)' % currency.company_id.name
```
rvalyi: sorry if that's a dumb question, but why use that 1st interpolation here (' (%s)' %) ?
can you not just do:
result.append((currency.id, '%s %s' % (currency.name, currency.company_id.name)))
?

rvalyi:
@pedrobaeza for me a name like base_\* is for modules which are "virtual" in the sense that they only become useful when extended by other "concrete" modules. This is not the case here as this module is already usable alone. Also the name "unshare_" is quite explicit. So the name is okay for me, what do you think?

:+1:  LGTM

pedrobaeza:
It's more like an standard rule for names. Look at my comment in https://github.com/odoomrp/odoomrp-wip/issues/431#issuecomment-66614814, where I try to explain an homogeneus rule, but I'm not going to block this PR for this reason.

bealdav:
Thanks @pedrobaeza and @rvalyi, could I have your position on
https://github.com/odoo/odoo/issues/641
because to know what to put in homogeneus rules on multicurrencies, we have to find the optimal position. On my side I think, currencies should be shared. EUR is EUR, if your in US, pakistan, or in Argentina. Only currency rates evolve with time

yvaucher:
I agree with @rvalyi that `base_` prefix should be used by module which are abstract module giving base for other module as dependency.

bealdav:
Maybe It can be merged now ?

Viggor:
Seems good for me :+1:

yvaucher:
:+1:

PR #5, opened 2014-12-30
[ADD] module sale_check

bealdav:
Comes from this PR: split in 2 modules
https://github.com/OCA/multi-company/pull/1

### DIFF
--- a/sale_check_multicompany/__init__.py
+++ b/sale_check_multicompany/__init__.py
@@ -1,0 +1,11 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#  licence AGPL version 3 or later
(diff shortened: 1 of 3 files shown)


pedrobaeza:
I don't understand the motivation of this module, because pricelist rule access already limit the possibility to select a different company. The only scenario I see where can apply is that you have a parent company and you try to make a sale order for a child one, but pricelist is the smallest problem you are going to have.

Viggor:
Can you close and re open the PR for trigger a new travis test? (the last test is 3 months old)
The rest of this PR is all right.

elicoidal:
I agree with @pedrobaeza here. I dont see the point of this module.
You'd rather add a domain control or a security rule.

Issue #10, opened 2015-03-11
Module product_autocompany breaks travis ci test


duanyp1991:
The module: product_autocompany makes company_id on model(product_template) required 
however the demo data in sale/sale_demo.xml:326 have one record of product with empty company_id.

  This actually is quite normal in a real project, some customized modules might change the standard workflow which make the standard tests fail.

pedrobaeza:
That can be solved providing a migration script (for v7) or post-hook for v8 that fills with one company each empty record.

duanyp1991:
right! I'll create a migration script for that.

duanyp1991:
@pedrobaeza  I just checked the code of odoo,（ migration.py）, it seems that only when we update the module and the version is higher, the migration script will be executed. However in this case , we wanna run the migration script when the module is first installed. Please let me know if I miss anything. Thanks! :)

pedrobaeza:
For v7, if you check for `version is None`, you can execute it in first install. For v8, you have to use a pre or post-init hook. See an example here:

https://github.com/odoomrp/odoomrp-wip/blob/8.0/quality_control_stock/__openerp__.py#L42
https://github.com/odoomrp/odoomrp-wip/blob/8.0/quality_control_stock/__init__.py#L10

duanyp1991:
@pedrobaeza  Thanks very much for you answer. I get how it works for v8.
But for v7, What do you mean by 'version is None'. I checked the code again, the migration only works for the module whose state is "to upgrade" (no 'to install' state). I tried putting the version as none, not working.

pedrobaeza:
Here you can see an example of a migration script with `if not version:` condition that fits for performing at first install (not upgrade):

https://github.com/OCA/account-financial-reporting/blob/7.0/account_financial_report_webkit/migrations/7.0.1.0.2/post-migration.py

PR #11, opened 2015-03-16
[FIX] fix issue:Module product_autocompany breaks travis ci test #10

duanyp1991:
by adding a migration script.
 Changes to be committed:
    new file:   product_autocompany/migrations/7.0.0.0/post-migration.py

### DIFF
--- a/product_autocompany/migrations/7.0.0.0/post-migration.py
+++ b/product_autocompany/migrations/7.0.0.0/post-migration.py
@@ -25,7 +25,7 @@ def migrate(cr, version):
     if not version:
         # this is run only when the module is first installed
         # fill in the default company for products whose company is empty
-        cr.execute("update product_template"
+        cr.execute("update product_template "
                    "SET company_id="
                    "   (SELECT id from res_company limit 1)"
                    "    WHERE company_id is NULL;")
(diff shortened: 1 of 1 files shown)


product_autocompany/migrations/7.0.0.0/post-migration.py
```diff
@@ -0,0 +1,30 @@
...
+##############################################################################
+
+
+def migrate(cr, version):
+    if not version:
+        # this is run only when the module is first installed
+        # fill in the default company for products whose company is empty
+        cr.execute("update product_template"
+                   "SET company_id="
+                   "   (SELECT id from res_company limit 1);")
```
pedrobaeza: `WHERE company_id is NULL`

duanyp1991:
Hi @pedrobaeza  I added a migration script for this module, but I have a problem that: it's hard to determine which company to assign to product as default.

This module doesn't work well for multi-company environment if one product is shared among different companies. I think for product, the idea should be the same as my module: partner_multi_company: adding a new field: company_ids.

pedrobaeza:
Yeah, I know, but at least this script provides technical (but not business) correctness. Take into account that the cases where company_id is not set are few.

And indeed, ideally we should tend to a module like _partner_multi_company_.

yvaucher:
I think this fix is ok to keep it simple and stupid. When installing this module you should be aware that you cannot share your products anymore.

So :+1:

PR #12, opened 2015-04-13
product_name_unique_per_company converted for 8.0

franksl:


franksl:
This module works when creating a product with the sae name of another one that is already in db. It does not work when MODIFYING the name of a product to the name of another one already on db.
When I modify the name of a product (even without variants) the change is applyied to product_product.name_template field instead of product_template.name. How can this be solved?

gurneyalex:
@franksl be very careful as the name of a product is a translated field, and the implementation of the version you ported is broken in my opinion: a simple unique index is incorrect. 

@max3903 do you have any opinion on this?

Currently I'm :-1: on this PR (and in favor of removing the module from our code base if it is installable in 7.0 or 6.1)

gurneyalex:
@franksl did you sign the OCA CLA? I have one Frank in the OCA database, but cannot confirm this is really you...

franksl:
Hi, thanks for the answer, sorry, my fault, I didn't take translation into account, I needed unique names as requested by a customer who doesn't handle translations.
Is there a way to have unique names? Maybe by using constraints() method instead of a db index?

Actually I didn't sign the CLA, I'll do as soon as I have the time to read all the papers,
Thanks!

gurneyalex:
@franksl CLA ping?

franksl:
I'll do that as soon as possible

max3903:
@franksl ping CLA ?

@gurneyalex What is the problem with the translation ?

gurneyalex:
@max3903 the problem is that product name is a translated field. Adding a UNIQUE index on the database column will only ensure uniqueness of the "source" translation, not of the displayed value.

lasley:
Hi @franksl - any update on this?

franksl:
Hello, sorry but I do not work with Odoo anymore so I cannot maintain any changes.

PR #13, opened 2015-07-13
[ADD] module inter_company_rules

chafique-delli:
This module has been backported from Odoo v9.

PR #14, opened 2015-07-16
[add] multi_company_holding_invoicing module

chafique-delli:


chafique-delli:
This PR depends of https://github.com/odoo/odoo/pull/7780.

pedrobaeza:
Is this a kind of ICP (inter-company rules)? Isn't better to backport current saas-6 inter-company rules?

sebastienbeau:
Hi Pedro we already have backported the inter-company-rule here : https://github.com/OCA/multi-company/pull/13
This module have a different behaviour and it's complementary to the inter-company-rule.

Our case is "simple". We have a holding company and X "worker company". Depending of the "channel of sale" we may need to invoice the sale to the holding and then the holding will invoice the customer, or directly invoice the customer. Invoicing is based on shipping.

So a picking will be invoiced
- by the "worker company" to the "holding company" (the module intercompany will generate the supplier invoice into the holding)
  - by the holding to the customer

The invoice a generate by a cron and depending of the customisation done on the domain, you may have case where the picking is already invoiced by the holding but not by the worker and some case there it's the opposite.

Hope my explication help you, we should improve the description

multi_company_holding_invoicing/stock.py
```diff
@@ -0,0 +1,291 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp import fields, models, api, _
+from openerp.exceptions import Warning
```
moylop260: Can you change this warning?
see https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#idioms

multi_company_holding_invoicing/stock.py
```diff
@@ -0,0 +1,291 @@
+# -*- encoding: utf-8 -*-
```
moylop260: Use # coding: utf-8

moylop260:
Could you see pylint beta messages?
checks without change status-1

pedrobaeza:
Is there still interest in this?

PR #16, opened 2015-09-10
[ADD] stock_production_lot_multi_company

agaldona:
New module which makes serial numbers multi company aware.

### DIFF
--- a/stock_production_lot_multi_company/models/stock_production_lot.py
+++ b/stock_production_lot_multi_company/models/stock_production_lot.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Ainara Galdona - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import fields, models
+
+
+class StockProductionLot(models.Model):
+
+    _inherit = 'stock.production.lot'
+
+    def _default_company_id(self):
+        company_model = self.env['res.company']
+        company_id = company_model._company_default_get('stock.production.lot')
+        return company_model.browse(company_id)
+
+    company_id = fields.Many2one(
+        comodel_name='res.company', string='Company', change_default=True,
+        default=_default_company_id)
--- a/stock_production_lot_multi_company/views/stock_production_lot_view.xml
+++ b/stock_production_lot_multi_company/views/stock_production_lot_view.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+
+        <record id="view_production_lot_form" model="ir.ui.view">
+            <field name="name">stock.production.lot.company.form</field>
+            <field name="model">stock.production.lot</field>
+            <field name="inherit_id" ref="stock.view_production_lot_form"/>
+            <field name="arch" type="xml">
+                <field name="ref" position="after">
+                    <field name="company_id" groups="base.group_multi_company"/>
+                </field>
+            </field>
+        </record>
+
+        <record id="view_production_lot_tree" model="ir.ui.view">
+            <field name="name">stock.production.lot.company.tree</field>
+            <field name="model">stock.production.lot</field>
+            <field name="inherit_id" ref="stock.view_production_lot_tree"/>
+            <field name="arch" type="xml">
+                <field name="create_date" position="after">
+                    <field name="company_id" groups="base.group_multi_company"/>
+                </field>
+            </field>
+        </record>
+
+    </data>
+</openerp>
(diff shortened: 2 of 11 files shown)


stock_production_lot_multi_company/models/stock_production_lot.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Ainara Galdona - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import models, fields
```
ajite: Import should be alphabetically ordered ([source](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports))

`from openerp import fields, models`

stock_production_lot_multi_company/models/stock_production_lot.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class StockProductionLot(models.Model):
+
+    _inherit = 'stock.production.lot'
+
+    company_id = fields.Many2one(comodel_name='res.company', string='Company')
+
+    _defaults = {
+        'company_id': lambda self, cr, uid, c:
```
ajite: Would not it be better to use the Odoo 8 API `default` parameter in the field declaration?

stock_production_lot_multi_company/views/stock_production_lot_view.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+
+        <record id="stock_production_lot_company_form" model="ir.ui.view">
```
ajite: The record `id` should be `view_production_lot_form`  see the [OCA Guidelines on inherited xml](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#inherited-xml)

stock_production_lot_multi_company/views/stock_production_lot_view.xml
```diff
@@ -0,0 +1,28 @@
...
+            <field name="model">stock.production.lot</field>
+            <field name="inherit_id" ref="stock.view_production_lot_form"/>
+            <field name="arch" type="xml">
+                <field name="ref" position="after">
+                    <field name="company_id" groups="base.group_multi_company"/>
+                </field>
+            </field>
+        </record>
+
+        <record id="stock_production_lot_company_tree" model="ir.ui.view">
```
ajite: The record `id` should be `view_production_lot_tree`  see the [OCA Guidelines on inherited xml](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#inherited-xml)

moylop260:
Excellent module.
+1 with @ajite comments

stock_production_lot_multi_company/models/stock_production_lot.py
agaldona: When you call to the method _company_default_get using Odoo 8 API it raises an error. We use the old API to avoid it.

pedrobaeza: It seems to be an error that has been introduced in last days in Odoo. I'm investigating it, but until them, this solution works.

agaldona:
I just changed the ids of inherited views and the order of imports.

stock_production_lot_multi_company/models/stock_production_lot.py
moylop260: Can you add a comment with link reference to odoo issue to future refactory change it to `default=` param?

moylop260:
:+1:

ajite:
LGTM technically.

pedrobaeza:
@agaldona, for making more new-API oriented, you have to put the following:

```
class StockProductionLot(models.Model):

    _inherit = 'stock.production.lot'

    def _default_company_id(self):
        company_model = self.env['res.company']
        company_id = company_model._company_default_get('stock.production.lot')
        return company_model.browse(company_id)

    company_id = fields.Many2one(
        comodel_name='res.company', string='Company', change_default=True,
        default=_default_company_id)
```

stock_production_lot_multi_company/models/stock_production_lot.py
pedrobaeza: I have investigated the issue and it's an unfortunate chain of consequences difficult to solve: original model is on old API, and this field is in new API, leading to a confusion in the compatibility layer. It's better to solve it as I propose below and let this question, because old API is to extinguish.

agaldona:
Changes applied.

pedrobaeza:
Can you please make only one commit for this?:

```
git reset c39cf6c7c75b3bf9a28eee3ec7067abc38f7601c
git commit --all --amend
git push -f
```

agaldona:
Done!

pedrobaeza:
:+1:

PR #32, opened 2016-11-22
[8.0] account_invoice_inter_company

chafique-delli:


pedrobaeza:
Isn't better to call this `base_inter_company`, as is the base for the other modules as stated in https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#modules?

chafique-delli:
@pedrobaeza ,
The name 'base_inter_company' is not precise enough about the function of the module.

pedrobaeza:
Well, you're right that it's not enough precise. OK, let's go for this one, but the README needs to be redone with new split

sebastienbeau:
It's ok for me, tested and running in production :+1:

bealdav:
:+1:

PR #35, opened 2017-01-13
[10.0][MIG] product_multi_company

lasley:
Migration of `product_multi_company` to v10
* Rename manifest
* Change openerp references to odoo
* Bump version

This depends on
- [x] #39


pedrobaeza:
It seems there are some problems with the tests.

lasley:
Ahhhh yup I have these modules installed in my local. Fix incoming, simple simple!

Weird runbot didn't fail though.

lasley:
Am I reading [v10 product security right](https://github.com/odoo/odoo/blob/10.0/addons/product/security/ir.model.access.csv)? There's no way for a non-admin user to edit a product without depending on `sales_team`? 

Adding a dependency simply for a test seems like overkill IMO. I almost feel like it would be better to create a new group in the test with proper permissions. Or maybe just edit an existing group. Thoughts?

lasley:
Hooray editing the security seems to have worked & I don't think will have any side effects due to transaction rollback. This is ready for review.

product_multi_company/hooks.py
```diff
@@ -0,0 +1,44 @@
...
+            'domain_force': (
+                "['|', ('company_ids', 'in', user.company_id.ids),"
+                " ('company_id', '=', False)]"
+            ),
+        })
+        # Copy company values
+        template_model = env['product.template']
+        groups = template_model.read_group([], ['company_id'], ['company_id'])
+        for group in groups:
+            if not group['company_id']:
```
pedrobaeza: Put here `# pragma: no cover` as the next line is not important to test and you'll achieve full coverage.

pedrobaeza (approved):
Little nitpicking

lasley:
This is now WIP, I'm about to strip some logic out and move it into `base_multi_company` #39

Cedric-Pigeon:
@lasley @pedrobaeza Hello I totally disagree with this PR. From my point of view, it should not depends on https://github.com/OCA/multi-company/pull/39. For example, in my project I use product-multi-company but I am not interested at all to have active per company.
The suggestion is really interesting but should not be imposed to every one.
I will suggest a new PR to simply migrate product-multi-company as it was, not less, not more.

pedrobaeza:
@Cedric-Pigeon, if you don't include active mechanism too, you will have usability problems. I have already found that problems, but minimize them with the users, covering the underlying problem.

What are your concerns? Performance?

Cedric-Pigeon:
@pedrobaeza My concern is that my customer don't want this active mechanism. He want a single flag common to all companies.
Yes I am afraid we will have a lot of problems. My customer has more than 80.000 products and performance were a real day-to-day fight until now and I don't want to take any additional risk. 2 separated PR let the choice to use it or not.

pedrobaeza:
Maybe an opt-in/opt-out mechanism inside the same module? What I want to avoid is to have partner_multi_company, partner_multi_company_active, product_multi_company, product_multi_company_active and so on.

@lasley what do you think?

Cedric-Pigeon:
@pedrobaeza I understand your position but I still think we are too far from a simple migration, adding features should be done separatly otherwise backports will be totally impossible.

pedrobaeza:
Well, if someone do the migration for you, that's not a problem. I also understand your position about solving your specific problem and thus not complicating with other not needed features, but if the architecture is good (which is), migration for both modules (product_multi_company and base_multi_company) is not going to be a problem. Even more, you will reduce the needed code, as part of the logic is shared on base_multi_company.

Cedric-Pigeon:
@pedrobaeza Sorry for me the architecure is not good. You will multiply queries by 3. 
And, as I said backport will be totally impossible.

lmignon:
@pedrobaeza The proposed implementation in #39 would have a serious problem at scale!!!!! (https://github.com/OCA/multi-company/pull/39#issuecomment-296563772, https://github.com/OCA/multi-company/pull/39#issuecomment-296762391)

lmignon:
@pedrobaeza IMO 'visibility by company' and 'active by company' desserves different purposes and we must be able to choice one and/or other.  As commented the proposed architecture will lead to serious performance issue. Since product_multi_company is a requirement for our client and is a central module of our solution we propose to migrate the current implementation from 9.0 to 10.0. The current PR could be rebased on the migrated module once a solution to avoid performance issues and to put in place your opt-in/opt-out mechanism will be found. This approach will let us move forward and benefits to others users of the product_multi_company module.

lasley:
I recommend you take a look at how much code I have reduced in this module before stating that a 1:1 migration is the best option. This module and all like it duplicate a bunch of code, and 100% need a base module.

I see your point that active by company could not be desired in some instances. I can remove that feature from the #39 base, then just add it into another module. We could still avoid the cascading module issue mentioned by Pedro, because we'd be using a common base and would only need one module to change its operation. 

TBD whether an enable/disable option is feasible once I come up with a new implementation, but I'll keep that in mind as well. If that works, it can just be an upgrade to the base instead of a new module.

pedrobaeza:
OK, for me let's go for base_multi_company + product_multi_company depending on it with a lot of duplication removed, and then an extra module base_multi_company_active that adds automatically on all dependent modules (partner_multi_company, product_multi_company...) the active by company feature.

lasley:
Just as a note, once the base merges in I will also upgrade `partner_multi_company` to v10 using it too.

Cedric-Pigeon:
@lasley @pedrobaeza great for me if you separate active per company in another module. Thanks for your consideration.

Cedric-Pigeon:
@lasley Could you please remove base_multi_company module form this PR ?

lasley:
@Cedric-Pigeon - This PR is dependent on `base_multi_company` and will not work without it. I will rebase once #39 is merged, which will remove `base_multi_company`

Cedric-Pigeon:
@lasley Yes I know I tested it on my own branch with the PR `base_mutli_company` and it works very well.

lasley:
This has been rebased and squashed now that #39 has been merged, and is now ready for review!

product_multi_company/tests/test_product_multi_company.py
```diff
@@ -0,0 +1,94 @@
...
+        self.assertTrue(company.id in product.company_ids.ids)
+
+    def test_company_none(self):
+        self.assertFalse(self.product_company_none.company_id)
+        # All of this should be allowed
+        self.product_company_none.sudo(self.user_company_1.id).name = "Test"
+        self.product_company_none.sudo(self.user_company_2.id).name = "Test"
+
+    def test_company_1(self):
+        self.assertEqual(self.product_company_1.company_id, self.company_1)
```
lasley: I think I missed some part of the functionality of this module. Looking here we are operating on the `product.product`, but we have implemented on `product.template`.

Is the intent of this module to distribute the companies to the variants too? In my case this is less than ideal, because the variants are owned by the companies themselves. Are there cases I am not considering, in which the variants would be owned by multiple companies?

lmignon: @lasley yes we have cases where the variants are owner by multiple companies...

lasley: @lmignon The same companies in all instances?

product_multi_company/hooks.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015-2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from odoo.addons.base_multi_company import hooks
```
elicoidal, line 10: Isnot this redundant with the dependency in manifest?

elicoidal (approved):
LGTM

product_multi_company/tests/test_product_multi_company.py
Cedric-Pigeon: @lasley I agree with @lmignon. I think that by default the sharing of variants must be identical than the  sharing of their template.

Cedric-Pigeon:
@lasley Could you check why tests are failing ?

product_multi_company/tests/test_product_multi_company.py
lasley: Yeah I have to design this. My use case is not quite so simple, and requires isolated company sets. I've set some time next week for this.

lasley:
@Cedric-Pigeon - tests are failing due to the product variant discussion [here](https://github.com/OCA/multi-company/pull/35#discussion_r118763250)

product_multi_company/tests/test_product_multi_company.py
Cedric-Pigeon: @lasley I investigated deeper your problem. This line has no chance to succeed because the user on `self.product_company_1` is admin and its company is not `self.company_1` but main company. I think the test should be `self.assertEqual(self.product_company_1.sudo(self.user_company_1).company_id, self.company_1)` because in base_multi_company, the `company_id` field is computed based on the user of `self.env`.

Another way to solve it is to create `product_company_1` with `self.user_company_1` in the setUp method.

product_multi_company/tests/test_product_multi_company.py
```diff
@@ -0,0 +1,94 @@
...
+        # All of this should be allowed
+        self.product_company_1.sudo(self.user_company_1).name = "Test"
+        self.product_company_both.sudo(self.user_company_1).name = "Test"
+        # And this one not
+        with self.assertRaises(AccessError):
+            self.product_company_2.sudo(self.user_company_1).name = "Test"
+
+    def test_company_2(self):
+        self.assertEqual(self.product_company_2.company_id, self.company_2)
```
Cedric-Pigeon: Same issue here, the test should be  `self.assertEqual(self.product_company_2.sudo(self.user_company_2).company_id, self.company_2)`

lasley: Ahhh yes how could I forget the fancy new computation?! Good call 😄

product_multi_company/hooks.py
lasley: It is- but if for some reason the module isn't there, Odoo will fail to start.

lasley:
This PR should be good to go now.

Note - I am now proposing a module rename to `product_template_multi_company` in order to more closely align with this module's functionality. 

I will be submitting a new module `product_product_multi_company`, which matches the functionality I need and thought was provided here (companies at the variant level).

Cedric-Pigeon (approved):
Good for me

pedrobaeza:
Why don't you put both functionalities in this module?

lasley:
@pedrobaeza - In this module, the template and variant operate in identical fashion in respect to the companies. It was stated that the product variants and templates need to be aligned in [this convo](https://github.com/OCA/multi-company/pull/35#discussion_r118881205). While it was mentioned that this need only be the default operation, I need something that is fairly at odds with this module.

In my case, I require the template companies to be completely separate from the variant companies. A company with access to the variant will only receive read access to the template, unless that company is also in the template's companies.

This will effectively allow for management of the template by Company A, while the variants are managed by their individual companies (Variant 1 by Company B, Variant 2 by Companies C and D). Think of something like amazon.com's marketplace, where the UPCs are primarily controlled by Amazon, but some aspects of the posted product are controlled by the vendor.

In order to make this operate, my solution is to simply implement `base_multi_company` at the variant level + add some security rules. If I were to do that in this module, a significant amount of logic would be required in order to provide a flip switch - similar to what you did in  `product_multi_image`. While this is definitely feasible, it adds a layer of complexity that I don't feel is required here.

Cedric-Pigeon:
@pedrobaeza I completely agree with @lasley I think it should be split to a more specific module

lasley:
@Cedric-Pigeon @lmignon - How are you guys currently handling pricelists? Looking at this code, I feel like there should be massive security issues:
``` xml
    <record model="ir.rule" id="product_pricelist_comp_rule">
        <field name="name">product pricelist company rule</field>
        <field name="model_id" ref="model_product_pricelist"/>
        <field name="global" eval="True"/>
        <field name="domain_force"> ['|',('company_id','=',user.company_id.id),('company_id','=',False)]</field>
    </record>
    <record model="ir.rule" id="product_pricelist_item_comp_rule">
        <field name="name">product pricelist item company rule</field>
        <field name="model_id" ref="model_product_pricelist_item"/>
        <field name="global" eval="True"/>
        <field name="domain_force"> ['|',('company_id','=',user.company_id.id),('company_id','=',False)]</field>
    </record>
```

_Edit: added the code, whoops_

lasley:
Oh snap nevermind, seems they work just fine. Makes sense in hindsight, they just restrict by company & there would be no reason for multiple companies to sell them!

So yeah, I think this is good for merge - unless there's disagreement on the name change?

I'll squash after confirmation & before merge - there's a few logical commits in here that we don't want to lose.

pedrobaeza:
I prefer then to keep this one as _product_multi_company_, which is the usual way of working, and call the other module _product_variant_multi_company_, following same convention as the one used in https://github.com/OCA/product-variant

lasley:
@pedrobaeza - renamed back to product_multi_company

pedrobaeza (approved):
Please squash at your taste and I'll merge.

Cedric-Pigeon:
@lasley I think it should  be a good idea to include record rules on product_template and product_product don't you ?

lasley:
@Cedric-Pigeon - record rules are applied and removed by the [base hooks](https://github.com/OCA/multi-company/blob/10.0/base_multi_company/hooks.py)

@pedrobaeza - squashed!

pedrobaeza:
@lasley the product.product rule is missing as @Cedric-Pigeon says.

lasley:
@pedrobaeza - I don't see a product.product rule to update. [`product_comp_rule` is the only security rule for products in core](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/addons/product/security/product_security.xml#L43). Am I misunderstanding something?

lasley:
Wait thinking about it further - the company_id and company_ids is being implemented on the template. Wouldn't the security rules implicitly pass through due to the inheritance? I will need to create a new rule in my product.product module, but I think we're fine here. You guys have been using this previously with no security issues, yes?

pedrobaeza:
Yeah, you're right, I didn't remember that part. Indeed the security rule is inherited. Merging...

Issue #38, opened 2017-04-17
[RFC] partner_multi_company_active: Active per Company


lasley:
This module will allow for a partner to be assigned to multiple companies, but only active in a few of them. It could also easily be added directly into `partner_multi_company`, but I'm not sure if the functionality would be wanted by all.

Currently if a partner is assigned to two companies (C-A & C-B), is deactivated by C-A, it is also deactivated for C-B. This is not ideal, and could cause for partners to drop out of disparate company systems without notice.

This feature will be accomplished by overriding the `active` field on `res.partner` to be computed. We will add a new model `res.partner.deactivate` which will be used to store the activity flag.

It will also add a `deactivated_company_ids` o2m to `res.partner`, which will inverse to `company_id` on `res.partner.deactivate`.

The compute method of the `active` field will return `False` if there are records in `deactivated_company_ids` for the partner and the company that the user is logged in under. Otherwise, `True`.

It would be nice if we could somehow store this computation, or maybe just use the cache. This will be a major performance hit without, I think.

The inverse method of the `active` field will add a new new record into `res.partner.deactivate` if `active` is set to `False` and there is not already an existing, active `res.partner.deactivate` record for the partner and current company. If `active` is `True`, any active `res.partner.deactivate` record for the partner and the current company will be deactivated.

Note that the above plan is utilizing deactivation of associated records, instead of unlink, because this will create an audit log of partner activations/deactivations that could be potentially useful.

The plan looks pretty simple I think. My main question here is whether we should make the independent module `partner_multi_company_active`, or just add into `partner_multi_company`? I personally think that the `toggle_active` feature is inconsistent with intent while `partner_multi_company` is installed, but who am I to judge? Anyone have thoughts on this?

cc @pedrobaeza

pedrobaeza:
I prefer this to be part of partner_multi_company.

:+1: to the reasons why this is needed.

lasley:
@pedrobaeza Sweet I was honestly hoping you would say that. Another module seemed excessive

lasley:
Ok so as I was playing around with other modules in multi company context, I see this being a recurring theme. I went ahead and made a PoC for a `base_multi_company` module including a deactivation mixer following a similar plan as to what I laid out here - #39.

I will leave this open still though because it's not implemented where we need it yet.

PR #39, opened 2017-04-20
[10.0][ADD] base_multi_company: Create new module

lasley:
* Create new module to provide base multi company logic and mixin
* Add deactivation by company mixin

This is a WIP PoC for a base to provide #38.


base_multi_company/models/company_active_mixin.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class CompanyActiveMixin(models.AbstractModel):
+
+    _name = 'company.active.mixin'
+    _description = 'Company Active Mixin'
+
+    active = fields.Boolean(
+        compute='_compute_company_active_status',
+        inverse='_inverse_company_active_status',
```
lasley: Note these are out of compliance with OCA guidelines for method naming in order to avoid potential name clashes during inheritance. I'm open to other suggestions for naming though.

lasley:
@pedrobaeza - I'm actually going to rip out some of your logic from `product_multi_company` for the `company_id` vs `company_ids` handling too. Seems that this looks nearly identical to that in `partner_multi_company`.

pedrobaeza:
OK, no problem. Indeed, they share the same field. Anyway, record rules should be provided/changed by model.

lasley:
Yeah I can't quite figure out how to make the record rules abstract, so that'll have to be left to dependent modules. I'll put something for implementation in the readme though to make it easier.

Any other ideas of logic we need for multi-company handling outside of:
* Company assignment
* Activate/Deactivate by company
* Domain_force in security

lasley:
Hah I can make a hook helper for the securiy and centralize a bunch of that too! Dependent modules would just need to feed in a model name & rule to edit.

pedrobaeza:
One of the things I wasn't able to get in my tries was to was able to select other companies apart from the user current one (that are not childs) in the many2many company_ids field. This way, only admin can change allowed companies in one step.

lasley:
Ah yes I've run into this with multiple companies - You would expect that if someone is assigned to multiple companies, they would be able to at least see the other companies they represent in the dropdowns. This is a pain even without the multi-company modules if you ever have to assign a partner to a different company for some stupid reason.

I think the solution here would lie in updating the [existing domain_force](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/odoo/addons/base/ir/ir_rule_view.xml#L93) to be include a:
```
[('company_id', 'child_of', user.company_ids.ids)]
```

That seems almost too simple to have thwarted you though. Did you try something like this, and do you remember the results?

pedrobaeza:
The problem is that you don't want this to act in all Odoo, as this way, you will be able to change the field company_id for a sale.order, having the rest of the data referred to the current company.

lasley:
Oh crap you're right. So we need to act specifically on the `company_ids` fields that are provided by our mixin. Hmmmmmmm we can't add security on `AbstractModels` can we? That would help I think.

pedrobaeza:
The only way I can think about is to provide a virtual copy of the companies (maybe an SQL view over res_company) and a new model, with the record rule you have mentioned, for the field company_ids. Then, the record rule can be the same, as the IDs are going to be the same.

lasley:
We could do this with delegate inheritance I think?

```
class ResCompanyAssignment(models.Model):
    _name = 'res.company.assignment'
    _inherits = {'res.company': 'company_id'}
    company_id = fields.Many2one(
        string='Company',
        comodel_name='res.company',
        required=True,
        ondelete='cascade',
    )
```

Then just make the rules for `res.company.assignment`?

lasley:
Actually I think your SQL view idea is significantly more efficient on double-thought. Do you know of anything doing something similar in Odoo/modules? I figure it's just a matter of creating the PSQL view via the cursor, but it's always nice to see something existing 😉

pedrobaeza:
No, it should be with an SQL view, or you won't have same IDs for the many2many.

```
class ResCompanyAssignment(models.Model):
    _name = 'res.company.assignment'
    _auto = False

    def __init__(...
        # Create SQL view from res_company
```

lasley:
Whoah this `_auto = False` is a TIL. I can totally run with this, thanks @pedrobaeza

pedrobaeza:
OK, here you have a complete `__init__` example: https://github.com/odoo/odoo/blob/852940f1ba1219b106590efc580ec2b5fdfa1740/addons/account/report/account_invoice_report.py#L164

lasley:
I added in `product_multi_company` so I can play around with it in this testbed. Will strip out and move to the appropriate PR when done 🚀

base_multi_company/hooks.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015-2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
```
lasley, line 4: @pedrobaeza - some of this code was licensed under AGPL in your `partner_multi_company`. I created this module as LGPL, so it got included here as such. Are you good with switching the parts I scavenged from AGPL or would you like me to switch this module?

pedrobaeza: Ok, no problem

base_multi_company/models/company_activate.py
```diff
@@ -0,0 +1,100 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class CompanyActivate(models.Model):
+
+    _name = 'company.activate'
+    _description = 'Company Activations'
+
+    is_active = fields.Boolean(
```
elicoidal: @lasley I am wondering why you create a specific model while the property table is the perfect fit for that?

pedrobaeza: You're right, but it can be a possible side effect, as I have found that some triggers are launched with sudo rights, making this not work for properties. Maybe this is not happening in 10.0. It worths to be checked.

lasley: Is it bad that I have no idea what you guys are talking about? What properties table?

pedrobaeza: The suggestion is to just put active = fields.Boolean (company_dependent=True) and Odoo handles this automatically in the properties table. It can work...

lasley: Ohhhhhh cool yes I didn't know that was called the properties table, thanks for the elaboration. I have also run into some issues with sudo, but I will try again in this context and version just to make sure.

lasley:
@elicoidal @pedrobaeza - I switched over to company dependent, and it does not look like this is working for either the admin or non-admin users. Maybe something with the way that active is treated internally?

* [The Travis](https://travis-ci.org/LasLabs/multi-company/jobs/224394014#L559)
* [The Relevant Tests](https://github.com/LasLabs/multi-company/blob/1724096ddc21bec29934b6a36f81cdaba045a2a2/base_multi_company/tests/test_multi_company_abstract.py#L92)
* [The Field Definition](https://github.com/LasLabs/multi-company/blob/1724096ddc21bec29934b6a36f81cdaba045a2a2/base_multi_company/models/multi_company_abstract.py#L13)

It's kind of weird that the one which explicitly does a `.sudo(demo_user)` doesn't have the appropriate effect, I was honestly only expecting the admin one to fail. It could also be a fluke with how tests run, but does illustrate the issue described by Pedro.

lasley:
Ok so it seems that there is some sort of hidden default mechanism for fields that are company dependent - the field is actually False before any of the tests, even with a default of True. Setting it to True on creation only seems to do the trick for the first company too.

It looks like this behavior is due to the [`_default_company_dependent`](https://github.com/odoo/odoo/blob/10.0/odoo/fields.py#L446) handling for the fields, which just returns a new [`ir.property`](https://github.com/odoo/odoo/blob/10.0/odoo/fields.py#L612) for the field name and model. Looking in that model, it seems that the [`.get` method just returns `False` if no property is found](https://github.com/odoo/odoo/blob/44276dfdc3b69cd9b834bfa7037671558e6d0078/odoo/addons/base/res/ir_property.py#L130). [`.get_by_record` does the same thing](https://github.com/odoo/odoo/blob/44276dfdc3b69cd9b834bfa7037671558e6d0078/odoo/addons/base/res/ir_property.py#L119).

I think I can get around this pretty easily. Any idea if this model is used for things other than field definitions? My plan is simply to search `ir.model.fields` for the `name` and `model` that are provided to `get` and `get_by_record`, then just compute the field default as a return instead of the `False`.

lasley:
Hah sorry for the spam - seems I just needed to learn how to use the properties model. I just needed to insert a property record to act as the default for the active field. I did that in the `init` of the abstract model, if a record doesn't already exist.

I added an explicit test for the active flag using the admin user, and it seems that this case is now handled properly. 

So yeah I think this works. I'm going to finalize tests and stuff, let me know if there are any other suggestions!

pedrobaeza:
Is this out of WIP?

lasley:
I went ahead and removed `product_multi_company` from this PR, and set #35 to the new code + base. 

The only thing changing here now will be the addition of the tests for hook, so yup I'd say this is good to review!

elicoidal:
I had an error testing the feature. 
Logged in as demo user (mycompany): inactivate Agrolait
Logged in as company2 user (company2) and when arriving to the sales menu I get this:
TypeError: this.widget.active_view is null (with long js gibberish)

Looked related to your feature.
Agrolait was not visible either (which is not what I was expecting)

elicoidal:
You can log in runbot (company2 pw is 1)

elicoidal:
By the way, some simple demo data would be great :)

lmignon:
@lasley IMO you should avoid the use of company dependend field to avoid
performance issue. In my experience this can of field is unusable with a
large amount of records.

Le 21 avr. 2017 5:37 PM, "Dave Lasley" <notifications@github.com> a écrit :

@elicoidal <https://github.com/elicoidal> @pedrobaeza
<https://github.com/pedrobaeza> - I switched over to company dependent, and
it does not look like this is working for either the admin or non-admin
users. Maybe something with the way that active is treated internally?

   - The Travis
   <https://travis-ci.org/LasLabs/multi-company/jobs/224394014#L559>
   - The Relevant Tests
   <https://github.com/LasLabs/multi-company/blob/1724096ddc21bec29934b6a36f81cdaba045a2a2/base_multi_company/tests/test_multi_company_abstract.py#L92>

It's kind of weird that the one which explicitly does a .sudo(demo_user)
doesn't have the appropriate effect, I was honestly only expecting the
admin one to fail. It could also be a fluke with how tests run, but does
illustrate the issue described by Pedro.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/OCA/multi-company/pull/39#issuecomment-296225225>, or mute
the thread

lasley:
@lmignon - Do you remember at what point it becomes a problem? Is it when a large amount of records are in the model implementing the company dependent field's table, a large amount of records are in the properties table, or a large amount of records are actually selected to utilize?

The alternative here is to utilize [a specific model](https://github.com/LasLabs/multi-company/blob/d4729b88641bfcf9b924c19786d54dd2b07021e3/base_multi_company/models/company_activate.py) paired with a [related field and a computation](https://github.com/LasLabs/multi-company/blob/d4729b88641bfcf9b924c19786d54dd2b07021e3/base_multi_company/models/multi_company_abstract.py#L19), but this is a very similar strategy to what the properties model is doing.

lmignon:
@lasley First of all, a boolean in a company dependent field doesn't work since Odoo will never store the value into the properties table if the value is False. Therefore it's not possible to search with a domain `'active', '=', 'False'` IOW, it's not possible to write record rules based on this field.
You'll also encounter performance problems. All the company dependent values are stored into the same table. On large database this table will become the bottleneck. Each time you'll  read this field (record rule, tree view, form view, ...) Odoo will genereate an 'SQL select' statement with a `res_id in (list of ids) and res_model=...` on this large table.

Cedric-Pigeon:
I agree with @lmignon I tried to do it for a customer and it was not a success at all.

lasley:
@lmignon - The `False` issue you described was circumvented [here](https://github.com/LasLabs/multi-company/blob/a50b290947be501b30cabba2a3035e654f8259a6/base_multi_company/models/multi_company_abstract.py#L43) by adding in a `True` record with `False` for the `res_id` and a `fields_id` of the associated field. This seems to be a default mechanism that Odoo SA provided, so it was just a matter of making each inherited model add in a default if one doesn't exist. It seems that setting the value from the default will generate a new record, regardless of what the value is - [my tests look to back up this hypothesis](https://github.com/LasLabs/multi-company/blob/a50b290947be501b30cabba2a3035e654f8259a6/base_multi_company/tests/test_multi_company_abstract.py#L95), but I'll add some search ones just to make sure.

We're stuck between a rock and a hard place here. Any solution implemented will need an EAV table such as that implemented in `ir.property`. 

I'm having a bit of trouble understanding why this is really a problem though, assuming that the searched fields are indexed (they are). Databases are built for this, and I have personally used Postgres with tables millions of records deep. EAV is a commonly accepted architecture for the implementation of features such as this, and while I personally don't like it, I also don't know of a viable alternative in this instance.

Do you have any suggestions of how to implement this without `ir.property`/EAV? I thought about cache for a bit, but I think that's going to be more overhead than it's worth.

lasley:
@lmignon Ok so digging into this `ir.property`, I do see a bunch of stuff that's not really needed in this context. If I'm following this code right, each query to the property table is actually three queries; so this isn't simply a matter of a join against a monstrous table IMO. Thanks for letting me know about this issue - this would have been a serious problem at scale.

I'll move back to the original method of using [a custom-built EAV model](https://github.com/LasLabs/multi-company/blob/d4729b88641bfcf9b924c19786d54dd2b07021e3/base_multi_company/models/company_activate.py), but would like to know if there are alternate suggestions before I drop any more time on that.

lmignon:
@lasley I wonder if the best way to support these company dependent values would not be to create a new kind of field that would be stored into a json column...

lasley:
@lmignon well now that's an interesting thought - basically creating a NoSQL document inside of the SQL. I didn't know about the JSON PSQL type, thank you for showing this to me.

It looks like these columns are being stored as text in the db, which could affect performance I'm thinking. Interestingly though, it looks like you [can index keys](https://stackoverflow.com/questions/17807030/how-to-create-index-on-json-field-in-postgres-9-3) so it may be feasible to come up with a data structure allowing for something like this that isn't simply adding a new index every time a new company is created. 

I'm going to play around & benchmark. This might be a solid option!

lasley:
@Cedric-Pigeon @lmignon - I have removed company activation functionality from this PR as mentioned in https://github.com/OCA/multi-company/pull/35#issuecomment-298950788. Let's continue the discussion here please instead, as it is relevant to the base module and not `product_multi_company`.

base_multi_company/models/multi_company_abstract.py
```diff
@@ -0,0 +1,37 @@
...
+        string='Company',
+        comodel_name='res.company',
+        compute='_compute_company_id',
+        store=True,
+    )
+    company_ids = fields.Many2many(
+        string='Companies',
+        comodel_name='res.company.assignment',
+        default=lambda s: s._default_company_ids(),
+        validate=False,
```
lmignon: @lasley `validate=False` ??? I can't find this parameter into the odoo api.

base_multi_company/models/multi_company_abstract.py
```diff
@@ -0,0 +1,37 @@
...
+class MultiCompanyAbstract(models.AbstractModel):
+
+    _name = 'multi.company.abstract'
+    _description = 'Multi-Company Abstract'
+
+    company_id = fields.Many2one(
+        string='Company',
+        comodel_name='res.company',
+        compute='_compute_company_id',
+        store=True,
```
lmignon: @lasley IMO this one must not be stored. But you must specify a search parameter.

```python

company_id = fields.Many2one(
    ...
    search='_search_company_id'
)

@api.model
def search_company_id(self, operator, operand):
    return [('company_ids', operator, value)]
```

base_multi_company/models/res_company_assignment.py
```diff
@@ -0,0 +1,22 @@
...
+class ResCompanyAssignment(models.Model):
+
+    _name = 'res.company.assignment'
+    _auto = False
+
+    name = fields.Char()
+
+    @api.model_cr
+    def init(self):
+        tools.drop_view_if_exists(self.env.cr, self._table)
```
lmignon: @lasley The SQL view should never be dropped otherwise the foreign key into the Many2many field from `multi.company.abstract` will be set to null... '

lmignon (changes requested):
@lasley How ca we be sure that the company_id defined into the abstract model will override the one defined on concrete models as in https://github.com/OCA/multi-company/pull/35/files#diff-3a1117111ff3330d3ec1fc41ab7692b8R9 ?

lasley:
@lmignon - Thanks for the review, I'll fix up the inlines sometime today.

Regarding the `company_id` override in concrete models - assuming that the concrete model directly inherits the abstract, why would it not override?

base_multi_company/models/res_company_assignment.py
lmignon: @lasley This view should be created only once into an _init_hook. But I still wonder how the orm will play with this view. AFAIK it's not possible in SQL to have references on a VIEW.

pedrobaeza: This is the same strategy as "BI" views. Why is not valid? I don't get the reference part.

FYI, this is only a trick for allowing to select any of the allowed companies in a user that is logged on only one company.

lmignon: @pedrobaeza AFAIK the init is called at install and each time you update the module... the init_hook only the first time the module is installed.

pedrobaeza: Yeah, but the same principle would apply to the rests of the SQL views. There should be a reason why Odoo recreates them each time...

lmignon: @pedrobaeza 

> There should be a reason why Odoo recreates them each time...

Before odoo 8, it was not possible to declare init hooks.  The creation of views was only possible by an override of the `init` method. This approach lead to serious performance issues when you update (-u all) an odoo instance with views declared on large tables.

The only advantage is that you don't have to provide a migration script if the view is modified....

pedrobaeza: Well, that's true, but in version 10 there's already the same mechanism... Let's move then this creation here and see if we find any problem later.

pedrobaeza: One of the reasons I see now for this mechanism in Odoo BI views is the extensibility via inheritance, that can't be handled with init hooks, which I don't think it's the case here (or should we provide it the same?)

lmignon: @pedrobaeza The inheritance of the view is just a matter of `CREATE OR REPLACE VIEW ...` into the init hook of the module that extend another one.

pedrobaeza: Well, I don't think so, as you can't use here Odoo inheritance method, with the consequent problem on SaaS/multi-DBs instances.

base_multi_company/models/multi_company_abstract.py
lasley: Why not? It's stored in the other modules and seems to be working fine:

* https://github.com/OCA/multi-company/blob/9.0/product_multi_company/models/product_template.py#L22
* https://github.com/OCA/multi-company/blob/9.0/partner_multi_company/models/res_partner.py#L20

pedrobaeza: This is the eternal conflict between performance on writing and performance on searching. In this case, as this field hardly changes, I prefer to store the value for having a better performance on searching. If not, each query will need to link at least 2 tables.

lasley: That was my thought - I generally try to weigh my writes against my reads. Anything >=4x the reads typically gets a stored unless there are specific circumstances to warrant against that.

I think maybe an index is in order though?

One thing that computing it would do - the search could be dynamic against the multiple companies, but as a singleton. Example:

Current:
```
>>> record.id
10
>>> record.company_id.id
1
>>> record.company_ids.ids
[1, 2]
>>> self.search([('company_id', '=', 1), ('id', '=', 10)])
my.model(10)
>>> self.search([('company_id', '=', 2), ('id', '=', 10)])
my.model()
```

With computed and some domain alterations:
```
>>> record.id
10
>>> record.company_id.id
1
>>> record.company_ids.ids
[1, 2]
>>> self.search([('company_id', '=', 1), ('id', '=', 10)])
my.model(10)
>>> self.search([('company_id', '=', 2), ('id', '=', 10)])
my.model(10)
```

lmignon: @pedrobaeza @lasley  my main concern here is that the stored value is not predictable and not right for all users (always the last one into the list of available company). Now that it's possible to all the users to see the list of companies (company_ids -> 'res.company.assignment')  where the product is available even if they are not part of one of these companies, the value stored into this company_id field could be one company not allowed for the current user.... IMO this field should definitely never be stored to avoid unpredictable side effect

lasley: @lmignon It's the first available, but I see your point. How would just changing to computed solve this though, particularly because it seems like this would be a permissions issue if the user doesn't have access to view the company?

pedrobaeza: @lmignon, you're right, as this depends on a dynamic value like the user current company, we can't store it. Thanks for your pointer.

pedrobaeza: Thinking twice, this field is only provided for compatibility reasons for any model looking for the field, but it shouldn't be used for access rules. If we don't store it, maybe we have some problems with any SQL view that uses it...

lasley: Well my thought with the access rules is as follows:

* User A is assigned to Company B
* Product A is assigned to company_id A and company_ids A,B
* User A navigates to Product A
* Security issue when displaying company_id (Company A), which is included in the product view

pedrobaeza: Well, I haven't tried that. First impulse is to hide that field from views. Second one is to not hide, but see adding `compute_sudo=True`. The only doubt is how name_get() calls are done, which is the possible read errors that can happen. Have you tried?

Other option is to change also the comodel_name to the new SQL view, which won't have access problems at all.

lasley: Ok I just tried on Runbot, and it looks like the `name_get` is not a problem. There is a security issue if you click the company though. Take a look at Agrolait on the [v9 Runbot](http://3286807-9-0-7f3092.runbot1.odoo-community.org/) under demo/demo user. 

There is still an inconsistency when searching `company_id` for a company that isn't the primary though. Not sure how often that's actually used though.

pedrobaeza: > There is a security issue if you click the company though.
That's something expectable, as you can't actually access the information.

> There is still an inconsistency when searching company_id for a company that isn't the primary though.
I can't imagine right now any standard code in this use case...

@lmignon, what is your opinion? Performance with this assumptions? Maximum compatibility?

base_multi_company/models/res_company_assignment.py
lmignon: IMO if you provides the search method the company_id will be usable... and the compute should be replaced as follow
```python
    @api.multi
    @api.depends('company_ids')
    def _compute_company_id(self):
       user_companies = self.env['res.company'].search([])
        for record in self:
            record.company_id = False
            for company in record.company_ids:
                if company_id.id in user_companies.ids:
                    record.company_id = company_id
                    break
```

lasley:
@lmignon @pedrobaeza - Alright I have switched over to the compute implementation & moved the view creation into a hook. Seems it works. I tried a few restarts of the server and upgrades too with no issues. I'll take it!

Just need to add some tests for the new inverse & search methods, then I'll be done here. Any other suggestions?

lasley:
I actually spoke too soon sorry, Travis said something different than my local.

```
2017-05-04 19:34:22,841 5593 ERROR openerp_test odoo.modules.registry: Model res.company.assignment has no table.
```

base_multi_company/models/multi_company_abstract.py
```diff
@@ -0,0 +1,49 @@
...
+            (6, 0, Companies._company_default_get().ids),
+        ]
+
+    @api.multi
+    @api.depends('company_ids')
+    def _compute_company_id(self):
+        for record in self:
+            for company in record.company_ids:
+                if company.id in self.env.user.company_ids.ids:
+                    record.company_id = company.id
```
lmignon, line 39: @lasley 'break'

lasley:
@lmignon good catch with the break, updated!

base_multi_company/models/multi_company_abstract.py
```diff
@@ -0,0 +1,50 @@
...
+        string='Company',
+        comodel_name='res.company',
+        compute='_compute_company_id',
+        inverse='_inverse_company_id',
+        search='_search_company_id',
+    )
+    company_ids = fields.Many2many(
+        string='Companies',
+        comodel_name='res.company.assignment',
+        default=lambda s: s._default_company_ids(),
```
lmignon, line 23: @lasley @pedrobaeza IMO we should add `auto_join=True` to generate JOINs upon search through that field. This will improve the performances and it's safe since record rules doesn't apply/matter on this relation.

pedrobaeza: :+1:

lasley: TIL `autojoin`. Thanks @lmignon

base_multi_company/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    'version': '10.0.1.0.0',
+    'author': "LasLabs, Tecnativa, Odoo Community Association (OCA)",
+    'category': 'base',
+    'website': 'https://laslabs.com',
+    'license': 'LGPL-3',
+    'installable': True,
+    'application': False,
+    'pre_init_hook': 'create_company_assignment_view',
+    'depends': [
+        'base',
```
lmignon: @lasley base is no more required in Odoo 10

base_multi_company/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    'name': 'Multi Company Base',
+    'summary': 'Provides a base for adding multi-company support to models.',
+    'version': '10.0.1.0.0',
+    'author': "LasLabs, Tecnativa, Odoo Community Association (OCA)",
+    'category': 'base',
+    'website': 'https://laslabs.com',
```
lmignon, line 11: The web site must be ""website": "https://odoo-community.org/"

base_multi_company/models/res_company_assignment.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import fields, models
+
+
+class ResCompanyAssignment(models.Model):
+
```
lmignon: @lasley It would be nice to describe the meaning of this special model: the motivation, how it works and where the view is created. Indeed it will be hard for a developer how has not followed the discussion in this PR to understand why this model exists etc....

lmignon (approved):
@lasley Great Job! a few remaining issues but the code looks very good to me now. (code review)

Cedric-Pigeon (approved):
@lasley Great job. Thanks.

lmignon:
@lasley Can you wait before the merge.. I would like to check an hypothesis...

lmignon:
@lasley I've improved your PR to allow the use of `child_of` and `parent_of` operator on the `company_id` and `company_ids` fields https://github.com/LasLabs/multi-company/pull/4

lmignon:
@lasley @pedrobaeza IMO this one will be ready to merge once LasLabs#4 is merged.

lasley:
Alright LasLabs#4 is merged and I have squashed commits by author as best as possible.

Note that Github says 88272b5 is after 7e72709 in the timeline, but when rebasing it ends after 2c9777e instead. Due to this, we seem to have two consecutive commits by me that I can't wizard around.

lasley:
I think I've attended to everything - does anyone have any more suggestions, or are we good to merge?

lmignon:
:+1:

Cedric-Pigeon:
:+1:  Good to merge for me.

pedrobaeza:
Merged. Great example of community work!!

PR #40, opened 2017-04-20
[10.0][IMP] Update Travis File

lasley:
I just noticed that this repo is building using sudo. I don't see any reason why, so this PR changes that.

* Remove sudo from travis file
* Match Travis file more closely to sample travis

I'll submit for the other versions once this gets merged. This seems fairly intentional, so I'm sure I'm missing something.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,12 @@
-sudo: required
-cache: pip
+language: python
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Fast-tracking

PR #41, opened 2017-04-20
[9.0][IMP] Update Travis File

lasley:
* Remove sudo from travis file
* Match Travis file more closely to sample travis

re #40


pedrobaeza (approved):
FT

PR #47, opened 2017-06-14
[10.0][MIG] partner_multi_company

lasley:
* Bump versions
* Rename manifest
* Rename openerp to odoo
* Implement base_multi_company


mstuttgart:
I find the follow error when try install it:

>>> TypeError: Model 'res.partner' inherits from non-existing model 'multi.company.abstract'

To fix it, I added to `base_multi_company` in `depends` key in manifest file. This module works fine :+1:

lasley:
Good call @mstuttgart - fixed & rebased to include the current base_multi_company

lasley:
Alright this should be 🍏  now

pedrobaeza (approved):
You can improve coverage putting `_commercial_fields` method with `# pragma: no cover`

lasley:
> You can improve coverage putting `_commercial_fields` method with `# pragma: no cover`

Cheater! I went ahead and added a test instead 😉

pedrobaeza:
hehe, you are cheating also, as you are not testing the real thing: create a parent partner, create a child one, and see that the field `company_ids` is propagated to the child.

lasley:
Ahhh but then we get into the philosophical discussion of what is a unit test - in this instance, if we were testing the `company_ids` being present in a child partner, wouldn't we instead be breaching into integration/acceptance territory because we're not slicing off our piece of logic & instead testing the Odoo core at the same time? 

It's my understanding that in a unit test, one should assume all interfaces outside of the current logic are properly tested and working in order to minimize scope and duplicate logic.

On the devil's advocate side though - this is testing implementation not result, which is also bad because the interface could technically shift out from underneath us without a validation. We don't write integration/acceptance tests usually, so in a migration this could bite us.

All that said, I didn't even know what the `_commercial_fields` method did until you just told me 😆

pedrobaeza:
Well, terminology apart, what I consider that should be tested in Odoo is the feature you want to get. That said, imagine in version 11 Odoo uses another technique for knowing what are the fields that are propagated to the childs. When migrating this module to v11, you won't notice the problem if the other method is still present by any chance.

This is the thermometer I use for making the tests. I don't call them unit tests nor integration test. They are Odoo tests (with the surname "Pedro's flavor" if you prefer :wink: ).

mstuttgart (approved):
It's work. Thanks :+1:

pedrobaeza:
@lasley please squash at your taste for merging.

lasley:
Squashed!

Issue #48, opened 2017-06-23
Multi Company Hooks - Inactive Records


lasley:
All of our multi company hooks are susceptible to one simple issue - domains are implicitly forced to only include `active=True` items. This is a problem when it comes to pre-existing, yet deactivated items.

I have a quick fix, but am still in console on a production database. Didn't want to forget this, so please ping me if I do (week or twoish)

lasley:
Alright I think we're good here, I'm going to close to track in the PRs. Looks like I wrote around this in v10 `base_multi_company`, so we're implicitly good there I think too.

PR #49, opened 2017-06-23
[8.0][FIX] partner_multi_company: Fix hook on deactivated partners

lasley:
* Add domain to include both active and inactive partners into the install hook

Re #48 

Note - I'm having a bit of trouble figuring out how to actually test this. Is there a way to insert/change a record in the db before the module install?

### DIFF
--- a/partner_multi_company/hooks.py
+++ b/partner_multi_company/hooks.py
@@ -16,7 +16,7 @@ def post_init_hook(cr, registry):
                              "('company_ids', 'in', user.company_id.id), "
                              "('company_id', '=', False)]")
         # Copy company values
-        partner_model = env['res.partner']
+        partner_model = env['res.partner'].with_context(active_test=False)
         groups = partner_model.read_group([], ['company_id'], ['company_id'])
         for group in groups:
             if group['company_id']:
--- a/partner_multi_company/README.rst
+++ b/partner_multi_company/README.rst
@@ -56,6 +56,7 @@ Contributors
 
 * Oihane Crucelaegui <oihanecruce@gmail.com>
 * Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+* Dave Lasley <dave@laslabs.com>
 
 Maintainer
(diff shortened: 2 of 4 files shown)


pedrobaeza (approved):

StefanRijnhart:
Just curious, why do you prefer adding domain elements instead of calling the search with context active_test=False?

lasley:
@StefanRijnhart - because I've never heard of `active_test=False`. Does that just disable the implicit domain force?

StefanRijnhart:
Exactly: https://github.com/odoo/odoo/blob/8.0/openerp/models.py#L4488

lasley:
TIL, thanks @StefanRijnhart. Done!

partner_multi_company/hooks.py
```diff
@@ -16,11 +16,11 @@ def post_init_hook(cr, registry):
                              "('company_ids', 'in', user.company_id.id), "
                              "('company_id', '=', False)]")
         # Copy company values
-        partner_model = env['res.partner']
+        partner_model = env['res.partner'].with_context(active_test=False)
         groups = partner_model.read_group([], ['company_id'], ['company_id'])
         for group in groups:
             if group['company_id']:
-                partners = partner_model.search(group['__domain'])
+                partners = partner_model.search(domain)
```
StefanRijnhart: Is the change in this line intentional?

lasley: Nope not now that I am not editing the domain. Excellent catch

StefanRijnhart (approved):
Thanks!

StefanRijnhart:
If you **really** want to test this, does this inactive partner from the demo data help? https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/res/res_partner_demo.xml#L365

lasley:
It did, thanks! The init hook wasn't tested at all, so I added one and used the deactivated partner in order to nail two birds with one stone.

PR #50, opened 2017-06-23
[8.0][FIX] product_multi_company: Fix hook on deactivated records

lasley:
* Add domain to include both active and inactive records into the install hook

Re #48 

Same note from #49 - I'm having a bit of trouble figuring out how to actually test this. Is there a way to insert/change a record in the db before the module install?

### DIFF
--- a/product_multi_company/tests/test_product_multi_company.py
+++ b/product_multi_company/tests/test_product_multi_company.py
@@ -2,6 +2,8 @@
 # (c) 2015 Serv. Tecnol. Avanzados - Pedro M. Baeza
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
+from ..hooks import post_init_hook
+
 from openerp.tests import common
 from openerp.osv.osv import except_orm as AccessError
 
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):

StefanRijnhart:
Can we have active_test=False in this one as well?

lasley:
Updated to new strategy and added a test. Will port into the 9.0 PR once we get a CI 🍏

product_multi_company/tests/test_product_multi_company.py
```diff
@@ -77,3 +79,9 @@ def test_uninstall(self):
         domain = (" ['|',('company_id','=',user.company_id.id),"
                   "('company_id','=',False)]")
         self.assertEqual(rule.domain_force, domain)
+
+    def test_init_hook(self):
+        deactivated_product = self.env.ref('product.product_product_4d')
```
StefanRijnhart, line 84: While the product is inactive, its template is not inactive so it does not test the problem that was fixed in this change. How about Pedro's suggestion, to inject some data at this point and then call the hook method again?

lasley: The SQL in that suggestion is quite intense due to M2M relation. I'll look into it though yeah

pedrobaeza: Well, I think it's not very difficult:
```python
product = env['product.product'].create({
    'name': 'Test',
    'active': False,
})
env.cr.execute(
    "DELETE FROM <table_rel> "
    "WHERE product_id = %s",
    (product.id, ),
)
<call_hook>()
# Check that product.company_ids is filled.
```

pedrobaeza:
Closing this as not completed and very old.

PR #51, opened 2017-06-23
[9.0][FIX] product_multi_company: Fix hook on deactivated records

lasley:
* Add domain to include both active and inactive records into the install hook

Re #48


pedrobaeza (approved):

pedrobaeza:
Merging this old one as proper fix

PR #52, opened 2017-06-23
[9.0][FIX] partner_multi_company: Fix hook on deactivated partners

lasley:
* Add domain to include both active and inactive partners into the install hook

Re #48

### DIFF
--- a/partner_multi_company/README.rst
+++ b/partner_multi_company/README.rst
@@ -52,6 +52,7 @@ Contributors
 
 * Oihane Crucelaegui <oihanecruce@gmail.com>
 * Pedro M. Baeza <pedro.baeza@tecnativa.com>
+* Dave Lasley <dave@laslabs.com>
 
 Maintainer
 ----------
--- a/partner_multi_company/__openerp__.py
+++ b/partner_multi_company/__openerp__.py
@@ -6,7 +6,7 @@
 {
     "name": "Partner multi-company",
     "summary": "Select individually the partner visibility on each company",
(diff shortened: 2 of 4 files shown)


pedrobaeza (approved):

lasley:
Suggestions in  #49 have been brought in here. Forgive the squash, it was easier to just bring the direct commit over instead of hashing out the diff just to squash again

StefanRijnhart (approved):

pedrobaeza:
Travis is failing as demo data is not working. What I do for testing hooks is to set initial situation by SQL, and then call the hook.

elicoidal (approved):

pedrobaeza:
@elicoidal please don't merge if I have commented that there's an error.

elicoidal:
Oups sorry: I saw the 2 approved green and didnot see the comment.
May I suggest that if you have a blocking point you revert the approval to red in the future?

Do you want me to revert?

pedrobaeza:
The best indicator is that Travis is red, and you shouldn't merge anything in red unless it's very justified, so no need for me to add a "Request changes" point. Just see that section. As the problem is in the tests, a new PR solving it is enough.

StefanRijnhart:
Tests fixed here: https://github.com/OCA/multi-company/pull/70

lasley:
It's just tests, I'll submit a PR today. I think that'd be cleaner than a revert

lasley:
Orrrr. refresh the page and @StefanRijnhart already did it for me. Woot!

PR #55, opened 2017-07-07
[10.0][add] mcfix_base

JordiBForgeFlow:
Mcfix Base
==========

This module is part of a set of modules that are intended to make sure that
the multi-company functionality is consistent.

It adds a computed field `current_company_id` to the `res.partner`, so that
property fields associated to this model, defined in other modules, will only
allow users to select values that are consistent with the company that the
user is currently working on.

### DIFF
--- a/mcfix_base/README.rst
+++ b/mcfix_base/README.rst
@@ -1,0 +1,79 @@
+.. image:: https://img.shields.io/badge/licence-lgpl--3-blue.svg
+   :target: http://www.gnu.org/licenses/LGPL-3.0-standalone.html
+   :alt: License: LGPL-3
+
+==========
+Mcfix Base
+==========
+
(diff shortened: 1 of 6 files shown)


gurneyalex:
@jbeficent why is this on res.partner rather than res.users?

[edit] sorry I got it.

gurneyalex (approved):

JordiBForgeFlow:
Because it is used in other modules that extend the res.partner, so that fields that are company dependent can be restricted on the user's company.
https://github.com/Eficent/multicompany-fixes/blob/10.0/mcfix_account/views/res_partner_view.xml#L9

JordiBForgeFlow:
We're no longer maintaining for 10.0. Superseded by https://github.com/OCA/multi-company/pull/92

PR #56, opened 2017-07-07
[10.0] [add] mcfix_product

JordiBForgeFlow:
Mcfix Product
=============

This module is part of a set of modules that are intended to make sure that
the multi-company functionality is consistent.

It adds a computed field `current_company_id` to the `product.product` and
`product.template`, so that property fields associated to these models, defined
in other modules, will only allow users to select values that are consistent
with the company that the user is currently working on.

It also adds constraints to ensure consistency company-wise:

* Supplier Info must belong to the same company as the product's and the
supplier's.


gurneyalex (approved):

JordiBForgeFlow:
We're no longer maintaining for 10.0. Will send a separate PR.

PR #57, opened 2017-07-19
[10.0] porting stock_production_lot_multi_company

eLBati:


tafaRU (approved):

hurrinico:
:+1:  IMHO

hurrinico (approved):

stock_production_lot_multi_company/__manifest__.py
```diff
@@ -1,13 +1,12 @@
...
-#############################################################################
+# Copyright 2015 Ainara Galdona - AvanzOSC
+# Copyright 2017 Lorenzo Battistini - Agile Business Group
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
     "name": "Stock Production Lot Multi Company",
     "summary": "Make serial numbers multi-company aware",
-    "version": "8.0.1.0.0",
+    "version": "10.0.0.1.0",
```
pedrobaeza: 10.0.1.0.0

pedrobaeza (changes requested):
Please change first commit text, which is not correct (I don't remember why I mixed that)

stock_production_lot_multi_company/__manifest__.py
eLBati: @pedrobaeza brief explanation: we are using setuptools-odoo to build "unstable" modules from open PRs. So we use versions < 1.0.0 for modules to be released. Now, assuming the module to be ready for merging, I change version to 1.0.0

Otherwise, we could directly use 1.0.0 for new PRs, then increase the version (1.0.1 ...) for fixes done during the review process.

What do you suggest?

eLBati:
> Please change first commit text, which is not correct (I don't remember why I mixed that)

@pedrobaeza done
Also see https://github.com/OCA/multi-company/pull/57#discussion_r136263566

pedrobaeza:
I don't know too much setuptools-odoo, but isn't there any way to force to reinstall modules? 

I recommend you to check our Docker project: https://github.com/Tecnativa/docker-odoo-base, and its way to indicate source modules: you can point directly to a PR to be included (https://github.com/Tecnativa/docker-odoo-base/blob/8b3582fc2635b69a940b6ff7e934a6cba72ed9af/odoo/custom/src/repos.yaml#L16). It's totally another world and you will forget all the deployment problems...

pedrobaeza (approved):

eLBati:
@pedrobaeza we were using buildout before setuptools-odoo and  looking at ```repos.yaml``` reminds me ```buildout.cfg```.
One of the main differences between systems like buildout and setuptools-odoo is that, with buildout you are versioning the repository, while with setuptools you are versioning the module.
Indeed, you could install this single module in this specific version with
```
pip install odoo10-addon-stock-production-lot-multi-company==10.0.0.1.0
```

I must say, moving to setuptools-odoo I forgot several deployment problems   :smiley:

pedrobaeza:
But you still have to deal with the rest of the deployment details and its cumbersome edges (wkhmtltopdf, ruby, less, sass...). With this, you have all packaged.

Another question that you can't deal easily with setuptools is if you have several PRs with patches for the same module (which is not the rule, but sometimes happen). With our system you just need:

```yaml
./repository:
    ...
    merges:
        - oca $ODOO_VERSION
        - oca refs/pull/pr_1_number/head
        - oca refs/pull/pr_2_number/head
```
and this without the need of doing anything for having available that intermediate version (I suppose you need to publish it in any way).

We have been working with both systems (buildout and our new Docker system) and assure you ours it's more straightforward and powerful.

PR #60, opened 2017-08-26
[FIX] button 'Propagate taxes' does not work when called from company…

sergiocorato:
… which have not created the product - clean and simpler code for better readability


pedrobaeza:
Travis is not happy with the change. Please check.

sergiocorato:
I close it, I missed some logic

PR #62, opened 2017-09-01
[ADD] website_multi_company: Isolate websites by company

lasley:
This module allows for websites to be isolated by company.

 It does this by automatically maintaining a unique public user for each website & synchronizing the company between the two.


lasley:
Note this is totally a WIP. The reason I'm submitting so early is to gather ideas of areas other than the products where companies are relevant in e-commerce. I added a checklist in the main comment of areas that need to be covered.

pedrobaeza:
But are you sure this is needed? What I do is to configure one public user with the company I want to use and link the website to that user. With that and some patches to Odoo core is enough (you will need that patches anyway with this I think).

lasley:
@pedrobaeza - honestly I'm not. This is part of the reason I submitted so early.

How are you assigning a public user to the website?

I'd like to implement this without patching Odoo core if possible. Do you by any chance have a reference of what you had to patch when you did this?

pedrobaeza:
There's a field called `user_id` binded to each `website` record. You have to assign it at your convenience. The problem is that the field is not directly shown in website settings... but very simple to solve it.

Then, this patch that hasn't been merged is needed at least for multi-company: https://github.com/odoo/odoo/pull/19153

lasley:
Whoa ok awesome this totally makes things easier. The public user will be assigned to a company, which will implicitly limit the record selection based on those ACLs. 

Luckily in my instance this is healthcare, so I don't even have to worry about any of the taxing (although I approved your incredibly sane PR - hope it gets merged).

Would you be opposed to me just adding the `user_id` into the [config settings provided in the multi-theme module](https://github.com/OCA/website/blob/10.0/website_multi_theme/wizards/website_config_settings_view.xml)? Maybe even a note in the readme about it?

pedrobaeza:
I don't think that's the best place for it. Maybe what you can do is to keep this module which adds this field to the interface, and also on `website` record creation, creates a copy of the default user.

lasley:
Ok sounds like a plan. I'll alter it to `website_multi_company` then I think too.

lasley:
(I particularly like your idea of linking the public user creations with the website)

lasley:
Conceptually this is complete. I ended up not even putting the `user_id` in the view, and instead opted for a completely automated management strategy.

I still need to functionally test this in the e-commerce context, but the users that are being created are in the correct company so I'm thinking we're going to have good results.

lasley:
Reporting back - this still doesn't work in E-Commerce context because the products are not filtered based on the user's company. 

The proper filtering is added in `product_multi_company` though, so I'll submit a glue module shortly.

website_multi_company/hooks.py
```diff
@@ -0,0 +1,32 @@
...
+
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        websites = env['website'].search([])
+        public_user = env.ref('base.public_user')
+        used_existing = False
+
+        for website in websites:
+
+            if all((
+                website.company_id == public_user.company_id,
```
pedrobaeza, line 20: Isn't this line redundant?

website_multi_company/hooks.py
```diff
@@ -0,0 +1,32 @@
...
+                website.user_id == public_user,
+                not used_existing,
+            )):
+                used_existing = True
+                continue
+
+            user = public_user.copy()
+            user.write({
+                'company_id': website.company_id.id,
```
pedrobaeza: You have to write also `company_ids` field.

website_multi_company/models/res_users.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo import api, models
+from odoo.http import request
+
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    @api.model
+    def _signup_create_user(self, values):
+        values.update({
```
pedrobaeza: Why is this needed? If this is not working on core, shouldn't this be a patch upstream?

website_multi_company/models/website.py
```diff
@@ -0,0 +1,68 @@
...
+class Website(models.Model):
+    _inherit = 'website'
+
+    @api.multi
+    @api.constrains('user_id')
+    def _check_user_id(self):
+        """Do not allow same public user on two websites.s"""
+        for record in self:
+            existing = self.search([('user_id', '=', record.user_id.id)])
+            if len(existing) > 1:
```
pedrobaeza, line 18: `if existing:` is enough

website_multi_company/models/website.py
```diff
@@ -0,0 +1,68 @@
...
+                ))
+
+    @api.model
+    def create(self, vals):
+        """Assign a unique public user per website and align the company."""
+        if not vals.get('user_id'):
+            user = self._get_new_public_user()
+            company_id = vals.get('company_id') or self.env.user.company_id.id
+            if user.company_id.id != company_id:
+                user.company_id = company_id
```
pedrobaeza: You need to change `company_ids` also.

website_multi_company/models/website.py
```diff
@@ -0,0 +1,68 @@
...
+            vals['user_id'] = user.id
+        user = super(Website, self).create(vals)
+        # This is required to circumvent duplicate keys
+        user.login = 'public-%d' % user.id
```
pedrobaeza: What about user's name?

website_multi_company/models/website.py
```diff
@@ -0,0 +1,68 @@
...
+        """Keep the user company synced with website & don't allow editing."""
+        if vals.get('user_id') and not self.env.context.get('write_user'):
+            raise ValidationError(_(
+                'The website public user is maintained automatically and '
+                'cannot be set.',
+            ))
+        if vals.get('company_id'):
+            for record in self:
+                if record.user_id.company_id.id != vals['company_id']:
+                    record.user_id.company_id = vals['company_id']
```
pedrobaeza: `company_ids` also.

pedrobaeza (changes requested):

website_multi_company/models/website.py
lasley: This is for the public user login itself. The user defined login is used when the actual user is created.

Without this fix, we can only have two websites. The third website (and subsequent third public user) will yield a duplicate key error on the login field for `public (copy)`

website_multi_company/models/website.py
lasley: We're going for a dup check here, so I'm checking for >1

website_multi_company/hooks.py
lasley: If there are already multiple websites existing in the db, this is used to make sure that they all have the managed user. Otherwise, we start in an unknown state.

website_multi_company/models/res_users.py
lasley: If there are already multiple websites existing in the db, this is used to make sure that they all have the managed user. Otherwise, we start in an unknown state.

(oops I put this on the wrong line first time sorry)

website_multi_company/hooks.py
lasley: Redundant with which other line?

website_multi_company/models/res_users.py
lasley: I don't know if this could be considered as not working in core. The design of core is that a [template user](https://github.com/odoo/odoo/blob/6921da00b3a253ea5d0cdd135a8629ef9cb14ca7/addons/auth_signup/models/res_users.py#L81) is used to create the users that are created by `auth_user`. 

The fix in core would be to have one template user per website, but that would require a new module joining `auth_user` and `website`. In the instance of that, I would just submit the module to OCA so that it is guaranteed to stay Open Source.

I don't think the added logic is necessary at this time, although it would be a nice to have allowing for the auto-manage of both the public and template users at the website level. If we're not exposing the template user to the administrator for editing though, it doesn't seem like that much of an advantage over simply injecting the values.

website_multi_company/models/website.py
lasley: I need to drink coffee before things. Ok so yeah `name` doesn't have a unique constraint, but I can see how the canonical would be good here. Updating.

lasley:
Thanks for the review @pedrobaeza - most comments attended to, some responded to. Ignore my early morning lack of ☕️  stupidity in some of them please 🚀

lasley:
Alright CI should go 🍏  now

website_multi_company/README.rst
```diff
@@ -0,0 +1,53 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
+
+============================
+Website - Multi-Company
+============================
```
pedrobaeza: Length should be the same as the title

pedrobaeza (changes requested):

website_multi_company/hooks.py
pedrobaeza: If the following one. If you have already sync `company_id` with the company of the user, you don't need this check.

website_multi_company/models/res_users.py
pedrobaeza: OK, please investigate further is this is really needed (this is one of the uncovered methods in tests), and if so, put a docstring explaining its usage.

website_multi_company/models/website.py
pedrobaeza: OK, as the check is done also on write, you can have in the search the current record.

treviser:
@lasley Not only products should be isolated, but website public categories too. But these are not multi-company yet. Would it be possible to make them multi-company aware using the module base_multi_company?

lasley:
Ahhh good call @treviser. I'll add that into #63 and expand the scope a bit, calling it `website_sale_multi_company` instead of `website_product_multi_company`

lasley:
Whoah actually... I could implement multi company on [`website.seo.metadata`](https://github.com/odoo/odoo/blob/10.0/addons/website/models/website.py#L670) instead, which is inherited by all things website (including the [public categories](https://github.com/odoo/odoo/blob/10.0/addons/website_sale/models/product.py#L53)). The standard security rule could then be added, which would restrict everything properly.

I wonder how much this will break things? This is a rather far-reaching change in one fell swoop.

website_multi_company/hooks.py
lasley: This is covering the normal case of a 1:1 `website`:`public_user` relationship. In this case, no user is created and no synchronization occurs. 

The `used_existing` part is just in case there are 2 websites assigned to the 1 public user. This would happen if someone maybe created a website for use in `website_multi_theme`, then decided to install this one to add the company isolation.

In all cases other than the standard 1 website to 1 company to 1 user, a new public user is created and assigned to the website.

There is one edge case in this logic that could cause the default public user to not be assigned to the default website, which would be if the company in either has been desynchronized manually for some reason.

lasley:
Well crap I knew that was too good to be true. This works alone, but not when we start installing the other modules (specific case is `product_multi_company`) due to a violation of [Python's Method Resolution Order (MRO)](https://www.python.org/download/releases/2.3/mro/).

Here's an example case illustrating the problem:

```python
class A(models.AbstractModel):
    _name = 'a'
class B(models.AbstractModel):
    _name = 'b'
    _inherit = 'a'
class C(models.Model):
    _name = 'c'
    _inherit = ['a', 'b']
```

Or, more simply (without Odoo):

```python
>>> class A(object):
...  pass
... 
>>> class B(A):
...  pass
... 
>>> class C(A, B):
...  pass
... 
  File "<stdin>", line 1, in <module>
TypeError: Error when calling the metaclass bases
    Cannot create a consistent method resolution
order (MRO) for bases B, A
```

In the example case, both `B` and `C` extend from `A`. `C` also extends from `B`. The problem in resolution order comes in because Python is not able to determine the ancestry once we have the child class that inherits from the same class as the parent.

In the `product_multi_company` case, we are inheriting the `multi.company.abstract` on `website.seo.metadata`. Both of these models are inherited by `product.template`, however:

```
TypeError: Cannot create a consistent method resolution
order (MRO) for bases ProductTemplate, website.seo.metadata, ProductTemplate, multi.company.abstract, website.published.mixin, ProductTemplate, base, mail.thread, ProductTemplate, ProductTemplate, ProductTemplate, Model, rating.mixin
```

I think I'm stumped on this one, so I'm going to revert the change to `website.seo.metadata` in favor of the originally planned direct support on product categories. That is, unless someone has some ideas on how to manually define the resolution order?

lasley:
Oh additional note, you can technically override the MRO with metaclassing:

```python
>>> class Meta(type):
...  def mro(cls):
...   return (cls, B, A)
... 
>>> class C(A, B):
...  __metaclass__ = Meta
... 
```

That's a bit more difficult when Odoo comes into play though. 

We could easily make a dynamic resolution order that simply ignores the second case of an inheritance; the problem is where to inject that. I've played with overriding the metaclass before, and ended up in a world of pain.

treviser:
I would like to help you, but I'm not a developer and am still struggling defining the hooks for my website_category_multi_company module :-( ...see here: https://github.com/OCA/multi-company/pull/65

treviser:
@lasley Are you aware of this: https://github.com/OCA/website/issues/381 ?

PR #64, opened 2017-09-05
[FIX] user should be able to create in company assignment model

Cedric-Pigeon:
This is needed otherwise you won't be able to create a record on the model inherited.

For example, I have a sale manager which is not able to create a product.

He has the right to create on product.template but not on ResCompanyAssignment.


Cedric-Pigeon:
@lasley

lasley (approved):
Thanks @Cedric-Pigeon - I never considered this case admittedly

Please raise the module version

pedrobaeza:
But that model is for a copy of the companies. It has nothing to do with product.template itself...

lasley:
Hrmmm and as a view, isn't it fully managed by Postgres?

Cedric-Pigeon:
fake, this is not cause of my problem. Sorry.

lasley:
I was misunderstanding the problem though, I thought this was when a user was attempting to create a new company (for some reason at the same time as a product... dunno why). 

So yeah, I tested and the demo user can create a company without the write rule. The company is magically in the assignment view too. Figured it would be good to explicitly test the case I thought this was for 🚀

PR #65, opened 2017-09-09
[IMP] base_multi_company: Improve ReadMe

lasley:
Now that I've forgotten how to use this module, it was the perfect time to update the ReadMe. 

Mainly because I forgot how to use it, so I had to rediscover everything because I suck and didn't write proper instructions the first time around.

* Add implementation instructions to ReadMe

### DIFF
--- a/base_multi_company/README.rst
+++ b/base_multi_company/README.rst
@@ -6,31 +6,84 @@
 Multi Company Base
 ==================
 
-This module provides an abstract model to be inherited by models that need to implement multi-company functionality.
+This module provides an abstract model to be inherited by models that need
+to implement multi-company functionality.
 
-Installation
-============
+No direct functionality is provided by this module.
 
-To install this module, simply follow the standard install process.
+Implementation
+==============
 
-Configuration
-=============
+Multi Company Abstract
+----------------------
 
-No configuration is needed or possible.
+The `multi.company.abstract` model is meant to be inherited by any model that
+wants to implement multi-company functionality. The logic does not require a
+pre-existing company field on the inheriting model, but will not be affected
+if one does exist.
 
-Usage
-=====
+When inheriting the `multi.company.abstract` model, you must take care that
+it is the first model listed in the `_inherit` array
 
-Todo
+.. code-block:: python
+
+   class ProductTemplate(models.Model):
+       _inherit = ["multi.company.abstract", "product.template"]
+       _name = "product.template"
+       _description = "Product Template (Multi-Company)"
+
+The following fields are provided by `multi.company.abstract`:
+
+* `company_ids` - All of the companies that this record belongs to. This is a
+  special `res.company.assignment` view, which allows for the circumvention of
+  standard cross-company security policies. These policies would normally
+  restrict a user from seeing another company unless it is currently operating
+  under that company. Be aware of apples to oranges issues when comparing the
+  records from this field against actual company records.
+* `company_id` - Passes through a singleton company based on the current user,
+  and the allowed companies for the record.
+
+Hooks
+-----
+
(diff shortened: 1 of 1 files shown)


rafaelbn (approved):
Great! :+1

sebalix (approved):

pedrobaeza (approved):

treviser:
@lasley In order to improve the description, because I'm still confused, but I have to confess I'm not a developer, so maybe my remarks are useless:
As I'm trying to create a module website_category_multi_company, I still do not understand properly what is meant by 

> +A module implementing these hooks would need to first identify the proper rule
 +for the record (`product.product_comp_rule` in the above example).

I can not find "product.product_comp_rule".
1. What exactly do I have to do: create a corresponding rule in "Settings/Technical/Security/Record Rules" with the name "product.product_comp_rule", but there is already a rule with the name "Product multi-company"?
2. Does the actual version of [hooks.py](https://github.com/OCA/multi-company/blob/10.0/product_multi_company/hooks.py) in product_multi_company work out of the box (or is only a boilerplate) or why does @pedrobaeza 's version look so different: https://github.com/pedrobaeza/multi-company/blob/3723d3e265873364b8f70db60350606a14174c14/product_multi_company/hooks.py ?
3. How would such hooks look like for example for the model product.public.category?

lasley:
Hey @treviser - So in your case, you actually won't use the hooks provided. These are so that you can manipulate a pre-existing single-company security (such as in the instance of [`product_comp_rule`](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/addons/product/security/product_security.xml#L43)

You just want a simple security rule in this case, because the categories are completely naive of the company. You can see an example [here](https://github.com/OCA/multi-company/pull/62/commits/8d39b578b3b0d6fe6f38fba13185f7fc336f56e0#diff-0262fb648a0ec2bac543bf7fe8a79dde).

That said, I'm wondering if multi-company is the incorrect approach with this. I think maybe we should instead define a `website_id` on the categories so that they can be assigned that way. Otherwise, I think we're limiting use case quite a bit - vs. this way we could complement `website_multi_theme` as well

treviser:
You are right with assigning them to the website_id, that's much better.
...and thanks for the explanation with the security rule, now I understand it.

treviser:
@lasley Should we make the public product company simply website aware or should we also do something like the base_multi_company module as an abstract module for multiple website aware models? If I think about blogs, forum, slides and so on...?

lasley:
@treviser - I tried to make a fell-swoop website aware in the [example I linked earlier](https://github.com/OCA/multi-company/pull/62/commits/8d39b578b3b0d6fe6f38fba13185f7fc336f56e0#diff-0262fb648a0ec2bac543bf7fe8a79dde). The problem is described [here](https://github.com/OCA/multi-company/pull/62#issuecomment-328601752) and the commit reverted [here](https://github.com/OCA/multi-company/pull/62/commits/1b20b8d5ba4f33da147651e063fae35c0f3062ec). _Edit: oops this was company aware_

Regarding an abstract model - what logic would we need to centralize? I'm thinking all that's needed for implementation of website-aware is a `website_id` key that defaults to the main website, and a `domain_force` security rule right? I think abstraction of that would be more trouble than worth in this instance.

lasley:
@treviser - Actually looking further, it may make sense to just add the `website_id` handling to `website.published.mixin` or `website.seo.metadata` to cover all of website. No additional modules would be required

treviser:
@lasley If I understand it correctly, we would nevertheless need a backend view in order to manage the manytomany relationship between a multi-website object and the website_id, similar to multi-company, at least for usability purposes.

PR #66, opened 2017-09-12
[8.0] purchase_sale_inter_company: fix and extract the search domain of users from the _check_intercompany_product method 

chafique-delli:
- Extract the search domain of the users from the _check_intercompany_product method to facilitate overloading
- Fix the search domain of the users to ensure that the user with whom the product is tested will have the necessary rights on 'purchase.order.line'
- Add the field 'note' in the list of fields to be sent to the sales order
- Fix french translate


bealdav (approved):

mourad-ehm:
:+1:

mourad-ehm (approved):

PR #67, opened 2017-09-12
[8.0] account_invoice_inter_company: fix and extract the search domain of users from the _check_intercompany_product method

chafique-delli:
- Extract the search domain of the users from the _check_intercompany_product method to facilitate overloading
- Fix the search domain of the users to ensure that the user with whom the product is tested will have the necessary rights on 'account.invoice.line'
- Add the field 'comment ' in the list of fields used to create  the invoice


bealdav (approved):

mourad-ehm:
:+1:

mourad-ehm (approved):

PR #74, opened 2017-11-14
[8.0] [FIX] account_invoice_inter_company

chafique-delli:
fix search domain for the inter-invoice

### DIFF
--- a/account_invoice_inter_company/models/account_invoice.py
+++ b/account_invoice_inter_company/models/account_invoice.py
@@ -288,6 +288,10 @@ class AccountInvoice(models.Model):
                 for inter_invoice in self.sudo().search(
                         [('auto_invoice_id', '=', invoice.id)]):
                     inter_invoice.signal_workflow('invoice_cancel')
+                    inter_invoice.write({
+                        'supplier_invoice_number': '',
+                        'origin': invoice.company_id.name +
+                        _(' Canceled Invoice: ') + str(invoice.number)})
         return super(AccountInvoice, self).action_cancel()
 
 
--- a/account_invoice_inter_company/models/account_invoice.py
+++ b/account_invoice_inter_company/models/account_invoice.py
@@ -87,8 +87,10 @@ class AccountInvoice(models.Model):
         self._check_intercompany_product(dest_company)
         # if an invoice has already been genereted
         # delete it and force the same number
-        inter_invoice = self.search(
-            [('auto_invoice_id', '=', self.id)])
+        inter_invoice = self.search([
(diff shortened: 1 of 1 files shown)


bealdav (approved):

account_invoice_inter_company/models/account_invoice.py
```diff
@@ -286,6 +288,10 @@ def action_cancel(self):
                 for inter_invoice in self.sudo().search(
                         [('auto_invoice_id', '=', invoice.id)]):
                     inter_invoice.signal_workflow('invoice_cancel')
+                    inter_invoice.write({
+                        'supplier_invoice_number': '',
```
bealdav: s/'supplier_invoice_number': '',/'supplier_invoice_number': False,

bealdav (approved):
ok except typo

account_invoice_inter_company/models/account_invoice.py
chafique-delli: @bealdav ,
It's done.

hparfr (approved):

account_invoice_inter_company/models/account_invoice.py
```diff
@@ -87,8 +87,10 @@ def _inter_company_create_invoice(
         self._check_intercompany_product(dest_company)
         # if an invoice has already been genereted
         # delete it and force the same number
-        inter_invoice = self.search(
-            [('auto_invoice_id', '=', self.id)])
+        inter_invoice = self.search([
+            ('auto_invoice_id', '=', self.id),
+            ('company_id', '=', dest_company_id)
```
sebastienbeau, line 92: Just to understand, why do you need to had the filter on company_id?
Only one intercompany invoice will be generated so filetring on auto_invoice_id is enought?

chafique-delli: @sebastienbeau ,
case of an inter-company automatic invoicing (account_invoice_inter_company module):

in a company A (supplier) have created a customer invoice with a partner company B.When validating the customer invoice, this is created in company B, a supplier invoice with partner = company A and auto_invoice_id = id of customer invoice of the company A.
If we decide to cancel the customer invoice in company A to put it back in draft and modify the partner, then the supplier invoice in company B will be canceled but always keep the auto_invoice_id.
when we try to validate the customer invoice that has been updated with partner = company C, well we can not because it's trying to recreate an intercompany invoice by keeping the same internal number that's on the canceled invoice of the company B.

sebastienbeau (approved):

bealdav:
you can delete the old branch

PR #79, opened 2017-12-15
[MIG] *_multi_company: OpenUpgrade migration scripts

pedrobaeza:
@Tecnativa


pedrobaeza:
I merge this for having a proper migration. For the rest, it doesn't affect.

PR #86, opened 2018-02-23
[FIX] for _compute_company_id

sharonshentc:
So it takes the company the user is logged on instead of the first company that is allowed in the record and allowed in the user.


pedrobaeza (approved):

pedrobaeza:
Superseed by #91, as it has better documentation and tests.

PR #89, opened 2018-03-12
[10.0][FIX] account_invoice_inter_company: payment_term_id and reference

chafique-delli:


mourad-ehm (approved):
Code reviw

account_invoice_inter_company/models/account_invoice.py
```diff
@@ -173,13 +173,15 @@ def _prepare_invoice_data(self, dest_inv_type, dest_journal_type,
         }
         dest_partner_data = self.env['account.invoice'].play_onchanges(
             dest_partner_data, ['partner_id'])
+        if not dest_partner_data.get('payment_term_id'):
+            dest_partner_data['payment_term_id'] = self.payment_term_id.id
```
pedrobaeza: Payment terms might not shared between companies, so this doesn't serve as is.

pedrobaeza (changes requested):
And you have to change module version number

account_invoice_inter_company/models/account_invoice.py
chafique-delli: @pedrobaeza , You're right, I was too fast in my PR.
I'm sorry, I correct.

chafique-delli: @pedrobaeza , I wanted to have your opinion to share the terms of payment. I thought to add a boolean field on account.payment.term to say which one wants to share. And with an access rule, I authorize those who have this boolean checked ?

pedrobaeza: In the spirit on inheritability, I think you can test if there are read permissions for the target company, and if affirmative, then assign it. This way, you can implement any strategy for sharing them (or use the default of not assigning company).

chafique-delli: @pedrobaeza , ok, thank you.

alexis-via: I agree with Pedro. It's surprising that there are no record rules by default on payment term, which is a bit surprising. I made a PR to add it to see how the editor considers it https://github.com/odoo/odoo/pull/23607 But checking for read perm in the target company via the generic method for that and then add it to destination invoice or leave the field empty. If the field payment_term_id is left empty, then the due_date should be copied.

account_invoice_inter_company/models/account_invoice.py
```diff
@@ -173,13 +173,15 @@ def _prepare_invoice_data(self, dest_inv_type, dest_journal_type,
...
         return {
             'name': self.name,
             'origin': _('%s - Invoice: %s') % (self.company_id.name,
                                                self.number),
             'type': dest_inv_type,
             'date_invoice': self.date_invoice,
-            'reference': self.reference,
+            'reference': _('Invoice: %s') % self.number,
```
alexis-via: I think we should just put "self.number" in that field, without 'Invoice' prefix.

chafique-delli:
@pedrobaeza , Hi, I made the changes after your comments and those of @alexis-via , can you tell me what you think. Thank you.

chafique-delli:
@alexis-via , Hi, I made the changes after your comments and those of @pedrobaeza , can you tell me what you think. Thank you.

account_invoice_inter_company/models/account_invoice.py
```diff
@@ -173,13 +173,20 @@ def _prepare_invoice_data(self, dest_inv_type, dest_journal_type,
...
+                    dest_user).check_access_rule('read'):
+                dest_partner_data['payment_term_id'] = self.payment_term_id.id
         return {
             'name': self.name,
             'origin': _('%s - Invoice: %s') % (self.company_id.name,
                                                self.number),
             'type': dest_inv_type,
             'date_invoice': self.date_invoice,
-            'reference': self.reference,
+            'reference': self.number,
```
pedrobaeza, line 189: Why this change?

chafique-delli: @pedrobaeza ,I think it's clearer to have the original invoice number in the 'supplier reference' field.
I know it can duplicate with the field 'origin'.
By the way, the field 'origin', I could leave it empty.

pedrobaeza: OK, thanks for the clarification. Indeed it's a good option. Leave both.

pedrobaeza (approved):

Issue #72, opened 2017-10-31
partner_multi_company: user's partner random existence


pedrocs-exo:
Hi,
I found a weird bug on the partner_multicompany module: a user's partner is only visible to its Current Company users. Steps to reproduce:
1. Create User 1 and User 2 and set Company A and Company B on Allowed Companies for both. As partners for these users are automatically created, you may as well associate them to Company A and B on the partner form.
2. Log in as User 2, select Company B and navigate to the partners list. You should see a partner named User 1, right? Wrong, it's not there!
3. On a different browser log in as User 1 and select Company B as well. Guess what: if you refresh the browser you used on step 2 you will now see User 1.

Can you please confirm? I may be missing something here. FYI I've tested on both my system and the last runbot.

pedrobaeza:
Version?

pedrocs-exo:
Oops, forgot to mention that. Version 10.

pedrobaeza:
@lasley have you experimented this?

lasley:
This one's new to me.

The key is the partner itself. The `company_ids` field is resetting to only the company selected as the primary in the user form. If you navigate to the partner, you can re-add the other company (in the case of Runbot `fourth`), and the partner appears again.

This doesn't happen in console, so I'm assuming there is an `onchange` in play.

lasley:
It's in the [`res.users.write`](https://github.com/odoo/odoo/blob/c7a6409fac80a2d33006ab56fe40d27f2acd2cfd/odoo/addons/base/res/res_users.py#L356). Naturally it's nested deep 😑

PR #90, opened 2018-03-24
Fix/partner multicompany

agb80:
This PR include two commits that fix some miss behaviours that we notice on a recent implementation. Both issues are present for module on all versions but as currently we are using 8.0 for this customer I'm sending this PR against 8.0 and later when approved will push from other versions.

First commit is for solve error reported on issue #72. 

Second commit is for allow to create a new res.partner but enforcing any security rule that have being added on database. The exactly user case that we are solving is for a customer that want to stop most of the users to be able to create res.partner records with `is_company` field set on `True`

### DIFF
--- a/partner_multi_company/models/res_partner.py
+++ b/partner_multi_company/models/res_partner.py
@@ -39,6 +39,7 @@ class ResPartner(models.Model):
         return vals
 
     @api.model
+    @api.returns('self', lambda value: value.id)
     def create(self, vals):
         """Neutralize the default value applied to company_id that can't be
         removed in the inheritance, and that will activate the inverse method,
@@ -48,7 +49,10 @@ class ResPartner(models.Model):
         # We must suspend security during this creation because it fails due to
         # all the mail stuff in between that confuses security rules
         obj = getattr(self, 'suspend_security', lambda: self)()
-        return super(ResPartner, obj).create(vals)
+        new_record = super(ResPartner, obj).create(vals)
+        # Call check_access_rules on new record to honor security rules
+        self.browse(new_record.id).check_access_rule('create')
+        return new_record
 
     @api.multi
     @api.depends('company_ids')
--- a/partner_multi_company/models/res_users.py
+++ b/partner_multi_company/models/res_users.py
@@ -14,7 +14,7 @@ class ResUsers(models.Model):
         if 'company_ids' in vals:
             res.partner_id.company_ids = vals['company_ids']
         if 'company_id' in vals:
-            res.partner_id.company_ids = [(6, 0, [vals['company_id']])]
+            res.partner_id.company_ids = [(4, vals['company_id'])]
         return res
 
     @api.multi
@@ -22,5 +22,5 @@ class ResUsers(models.Model):
         res = super(ResUsers, self).write(vals)
         if 'company_id' in vals:
             for user in self.sudo():
-                user.partner_id.company_ids = [(6, 0, [vals['company_id']])]
+                user.partner_id.company_ids = [(4, vals['company_id'])]
         return res


agb80:
The CI error is from job with this configuration: 

```
"env": "TESTS=\"1\" ODOO_REPO=\"odoo/odoo\" EXCLUDE=\"product_multi_company\" LINT_CHECK=\"0\"",
```
as I understand is checking all other modules except the one I modify, do I need to fix that errors?

partner_multi_company/models/res_partner.py
```diff
@@ -48,7 +49,10 @@ def create(self, vals):
         # We must suspend security during this creation because it fails due to
         # all the mail stuff in between that confuses security rules
         obj = getattr(self, 'suspend_security', lambda: self)()
-        return super(ResPartner, obj).create(vals)
+        new_record = super(ResPartner, obj).create(vals)
+        # Call check_access_rules on new record to honor security rules
+        self.browse(new_record.id).check_access_rule('create')
```
pedrobaeza, line 54: I don't see this needed

partner_multi_company/models/res_users.py
```diff
@@ -14,13 +14,13 @@ def create(self, vals):
         if 'company_ids' in vals:
             res.partner_id.company_ids = vals['company_ids']
         if 'company_id' in vals:
-            res.partner_id.company_ids = [(6, 0, [vals['company_id']])]
+            res.partner_id.company_ids = [(4, vals['company_id'])]
```
pedrobaeza, line 17: Don't change this. It's on purpose.

pedrobaeza:
I don't think there's a problem at all. The issue is reported for version 10, not version 8 and the code is very different on both.

partner_multi_company/models/res_partner.py
agb80: You're right this belong to other pourpose

partner_multi_company/models/res_users.py
agb80: Would you mind explain why?

pedrobaeza: Because we want to overwrite full list, not to add one value to the current list.

agb80:
We are facing exactly the same problem that @pedrocasi reported on issue #72 on version 8 and I think even the rest of the code base is different two things remain equally and are the origin of the bug. 

The first are the security rules and the second is the write function for `res.users` object. Let me try to clarify our current user case with more detail

1. There are Company A and Company B totally independents among them.
2. There is User 1 whom is able to work on Company A and on Company B.
3. There is User 2 whom is only able to work on Company B.
4. For segment partner's visibility according to company the belong we are using a global rule on res.partner object with this domain: `['|','|',('company_ids.child_ids','child_of', [user.company_id.id]), ('company_ids','child_of',[user.company_id.id]), ('company_id','=',False)]`

So, when you create User 1 the `create` function found [here](https://github.com/OCA/multi-company/blob/8.0/partner_multi_company/models/res_users.py#L15) writes on `company_ids` field for related partner Company A and Company B and when you create User 2 same function set on `company_ids` field for User 2 partner Company B.

Until here everything is ok and when you login with User 2 on Company B you are going to be able to see User 1 partner.

The problem is when you login with User 1 and start to switch from Company A to Company B and back.  Every time user switch company `write` function found [here](https://github.com/OCA/multi-company/blob/8.0/partner_multi_company/models/res_users.py#L21) is called and as a result of `6, 0, vals` command it's replacing `company_ids` content with current company for User 1.

When User 1 switch from Company A to Company B there is no problem for User 2 because User 1 partner's company set on `company_ids`: Company B, it's on the companies available for him but as son as the User 1 switch back to Company A the User 1 partner's company set on `company_ids`: Company A now, is no longer under the companies available for User 2 and User 2 is not going to be able to see User 1 partner.

partner_multi_company/models/res_users.py
agb80: Would you mind check mi last post and let me know what you think?

agb80:
ping, any other comment about this?

pedrobaeza:
But Travis error is still there. This can't be merged while you don't fix that.

PR #91, opened 2018-03-27
[FIX][10.0][multi_company_abstract] Fix company_id computation issue (+ add unit test)

acsonefho:
**Steps to reproduce (actual) error:**
- Configuration: `product.template` inherit `multi.company.abstract`
- Log in (not with Admin) with a user set with at least 2 `company_ids`
- Browse a product with at least 2 `company_ids`
- Check the `company_id` of the product
- If the `company_id` of the product is the same than the current company, stay on the same product but switch your current company (top right button)
- The `company_id` of the product is not your actual company: you will have some issue (for example to use pricelist (price list set to `without_discount` as `discount_policy`) on `sale.order` because this model use the `product.company_id.currency_id` => BOOM, `AccessError` exception because you don't have rights to read another company than the current one).

**Fix (this MR):**
- To compute the `company_id` field of a `multi.company.abstract`, give the priority to the company of the current user. If this company is not into `company_ids` of the targeted model, pick the first company (actual behaviour)

### DIFF
--- a/base_multi_company/models/multi_company_abstract.py
+++ b/base_multi_company/models/multi_company_abstract.py
@@ -33,7 +33,13 @@ class MultiCompanyAbstract(models.AbstractModel):
 
     @api.multi
     def _compute_company_id(self):
+        user_company = self.env.user.company_id
         for record in self:
+            # Give the priority of the current company of the user to avoid
+            # access right error
+            if user_company.id in record.company_ids.ids:
+                record.company_id = user_company.id
+                continue
             for company in record.company_ids:
                 if company.id in self.env.user.company_ids.ids:
                     record.company_id = company.id
--- a/base_multi_company/tests/test_multi_company_abstract.py
+++ b/base_multi_company/tests/test_multi_company_abstract.py
@@ -2,7 +2,7 @@
 # Copyright 2017 LasLabs Inc.
 # License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
 
-from odoo import fields, models
+from odoo import exceptions, fields, models
 from odoo.tests import common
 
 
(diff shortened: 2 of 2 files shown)


pedrobaeza:
This seems similar to #86, isn't it?

acsonefho:
@pedrobaeza Ah yes exactly same issue (so quite similar fix). Sorry, I didn't see it before

pedrobaeza:
No problem. Let's keep this one as better documented and with tests.

base_multi_company/__manifest__.py
```diff
@@ -1,12 +1,13 @@
-# -*- coding: utf-8 -*-
```
pedrobaeza: Don't remove this in v10.

base_multi_company/__manifest__.py
```diff
@@ -1,12 +1,13 @@
...
 # Copyright 2017 LasLabs Inc.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
 {
     'name': 'Multi Company Base',
     'summary': 'Provides a base for adding multi-company support to models.',
     'version': '10.0.1.0.1',
-    'author': "LasLabs, Tecnativa, Odoo Community Association (OCA)",
+    'author': "LasLabs, Tecnativa,"
+              "ACSONE SA/NV,"
```
pedrobaeza: This patch is not enough for granting you co-authorship. Put you in contributors in README instead.

pedrobaeza (changes requested):

base_multi_company/__manifest__.py
acsonefho: Ok no problem. My internal check becomes red due to the coding utf-8 and because I didn't add my company into the module. I'll correct this in few minutes

acsonefho:
@pedrobaeza The test is red due to another module. Seems ok for you now?
Thanks

pedrobaeza:
No, @acsonefho , Travis is failing due to this flake8 error in your changes:
```
base_multi_company/tests/test_multi_company_abstract.py:188:13: F841 local variable '_' is assigned to but never used
```

SimoRubi (approved):
Code review, very nice with tests too!

pedrobaeza (approved):

PR #95, opened 2018-04-15
[10.0][MIG] purchase_sale_inter_company

chafique-delli:

PR #97, opened 2018-04-25
 [MIG] purchase_sale_inter_company: Migration to 11.0 (from 8.0) WIP

carlosdauden:
@Tecnativa


pedrobaeza:
Why not starting from #95?

pedrobaeza:
Please squash commit history a bit.

Joaco1980:
After installing the module it leaves inter company, when trying to configure it when I give to save changes the configuration is not saved correctly and it is
captura

pedrobaeza:
@carlosdauden, @Joaco1980's comment is addressed?

Joaco1980:
After the last commits, it works for me correctly. what I do not understand because it has the dependence on the intercompany_invoice, because if you take out merchandise in the sale and invoices the delivery notes, in company B an invoice is generated but it is not related to the purchase order.

yajo:
I added a commit that will make Travis fail until https://github.com/OCA/multi-company/pull/110 is merged, given this was WIP already...

purchase_sale_inter_company/models/sale_order.py
```diff
@@ -0,0 +1,36 @@
...
+        readonly=True,
+        copy=False,
+    )
+
+    @api.multi
+    def action_confirm(self):
+        for order in self.filtered('auto_purchase_order_id'):
+            for line in order.order_line.sudo():
+                if line.auto_purchase_line_id:
+                    line.auto_purchase_line_id.price_unit = line.price_unit
```
chienandalu, line 24: Wouldn't it be interesting to update the line qty if it's changed in the SO? (and maybe log the change into the PO chatter)

purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -0,0 +1,163 @@
...
+            :param purchase_line : the origin Purchase Order Line
+            :rtype purchase_line : purchase.order.line record
+            :param dest_company : the company of the created SO
+            :rtype dest_company : res.company record
+            :param sale_order : the Sale Order
+        """
+        new_line = self.env['sale.order.line'].new({
+            'order_id': sale_order.id,
+            'product_id': purchase_line.product_id.id,
+            'product_uom': purchase_line.product_id.uom_id.id,
```
chienandalu: Shouldn't it be `purchase_line.product_uom.id`? 
E.g.: you order 1 dozen of eggs but the sale order registers 1 unit of eggs (default uom in the product itself) and the price in the PO is syncronized to price of that unit. So the result is the company selling serves just 1 egg.

carlosdauden: Of course. This should be purchase_line.product_uom.id

pedrobaeza:
@carlosdauden please fix tests of this module.

bealdav:
As said by @Joaco1980 I think dependency with account_invoice_inter_company is not mandatory to make intercompany flows more modular and quicker ported to upper version according to each scenarii.
I propose to port myself purchase_sale_intercompany in v12 removing this dependency.
What do you think ? @pedrobaeza

pedrobaeza:
OK if not needed, but are you sure? I think there's some hook there.

Joaco1980:
I said that it was not necessary because when you work with storable products because it does not relate to warehouse movements you have to generate the invoice from the purchase orders so you have 2 invoices one that creates the inter invoice without a connection with delivery notes and another that you create your to invoice the movements of warehouse.

chienandalu:
I'd say it's not a good idea, since the invoice could differ from SO to PO (e. g.: shipping costs in the SO that won't be in the original PO)

bealdav:
Please @chienandalu could you confirm what is or not a good idea ? Thanks

chienandalu:
Delivery charges could be one case, discount set in invoice could be another, multiple invoices from the same SO could be problematic as well.

anajuaristi:
I agree with @Joaco1980 at this point.
For me, the single intercompany point is Sale-purchase or purchase-sale. The rest of the flow can be followed for each company on their own site.
I mean

Sale order Comp A --> Delivery order comp A
                               --> invoice comp A
         If S.O not automatically created by comp B --> Purchase on Comp B


Purchase Comp B --> Receipt order comp B
                             --> Invoice comp B
         If P.O not automatically created by comp A --> Sale Order on Comp A

Having intercompany invoice would only apply if the invoices are direct, not coming from Sales or Purchase orders

rafaelbn (approved):

rafaelbn:
This modules was proposed migration to v12 in https://github.com/OCA/multi-company/pull/120

pedrobaeza:
The problem is that tests are still broken. @chienandalu did you find the root cause?

chienandalu:
@pedrobaeza Green tests

pedrobaeza (approved):
OK, let's merge this finally!

PR #103, opened 2018-06-12
[10.0][FIX] base_multi_company, partner_multi_company: Fix search and uninstall partner_multi_company

igallart:


partner_multi_company/models/res_partner.py
```diff
@@ -12,6 +12,9 @@ class ResPartner(models.Model):
     _name = 'res.partner'
     _description = 'Partners (Multi-Company)'
 
+    display_name = fields.Char(compute='_compute_display_name', store=True,
```
pedrobaeza, line 15: Why do you need to re-add this field?

base_multi_company/hooks.py
```diff
@@ -57,12 +57,17 @@ def post_init_hook(cr, rule_ref, model_name):
...
         column1 = model._fields['company_ids'].column1
         column2 = model._fields['company_ids'].column2
-        SQL = """
+        sql = """
             INSERT INTO %s
             (%s, %s)
-            SELECT id, company_id FROM %s WHERE company_id IS NOT NULL
-        """ % (table_name, column1, column2, model._table)
-        env.cr.execute(SQL)
+            SELECT id, company_id FROM %s AS tb1
```
pedrobaeza, line 63: Why is this needed?

pedrobaeza:
Please change each module in one commit, explaining in it why you make that changes

partner_multi_company/models/res_partner.py
igallart: This change is because in Sale/Customers no search by name. With this change yes work.

base_multi_company/hooks.py
igallart: That is because when uninstall the module partner_multi_company and after install the same module partner_multi_company this crash and is not is possible install.

pedroortega-punt:
Is correct? If is correct, please merge.

Thanks

igallart:
Closed by inactivity.

PR #110, opened 2018-09-14
[FIX] account_invoice_inter_company: Move demo data to test

yajo:
This demo data is quite dangerous by design, since it creates compenies, accounts, decides the companies country, etc. All of that by hand.

In a tested integration environment, this fails when installing account-related addons due to companies not having a chart of accounts.

The fix is simple: remove this from demo data and leave it just as test data.

@Tecnativa


yajo:
Travis' :x: is unrelated, and the main branch is already broken [as you can see](https://travis-ci.org/OCA/multi-company/builds/411572002).

You can also check that [this addon's tests pass fine](https://travis-ci.org/OCA/multi-company/jobs/428578869#L1213).

account_invoice_inter_company/tests/test_inter_company_invoice.py
```diff
@@ -2,37 +2,38 @@
...
-        self.user_company_a = self.env.ref(
+        cls.user_company_a = cls.env.ref(
             'account_invoice_inter_company.user_company_a')
-        self.user_company_b = self.env.ref(
+        cls.user_company_b = cls.env.ref(
             'account_invoice_inter_company.user_company_b')
 
-        self.chart = self.env['account.chart.template'].search([], limit=1)
-        if not self.chart:
+        cls.chart = cls.env['account.chart.template'].search([], limit=1)
```
pedrobaeza, line 31: This shouldn't be the method. You have a method `try_load_chart_template` or similar executed on company record for loading the proper chart template.

yajo: AFAICS this line can be removed, as it's not used. I don't want to refactor everything unless strictly necessary, TBH.

pedrobaeza: But you need to load the chart template any way. Previously they tricked the system adding as demo all the stuff that this method makes. You need also to look for demo XML-IDs used here in the test.

yajo: Yes, I'm just loading the XML data in the test itself. [It seems a common pattern for testing](https://github.com/odoo/odoo/search?l=Python&q=def+_load) when you need to lazy-load some demo data, or you need to ensure its state. This way I don't need to rewrite the tests.

You can see [it works fine](https://travis-ci.org/OCA/multi-company/jobs/428578869#L1213).

pedrobaeza: Well, it can cause side effects, as not all data is properly set (from the `account` module perspective, and this is an irregular way of having a chart of account. For example, field `chart_template_id` on company is not set - as there's no chart template -), so this can throw side effects, not right now, but on future changes or dependent modules.

pedrobaeza:
I didn't have to change the approach for this, so I fast-track it (although branch is red due to other failed test), and I can, I'll do it in a new PR.

PR #111, opened 2018-09-19
[11.0][IMP] account_invoice_inter_company: option to avoid product_id in invoice lines

astirpe:
For our multicompany environment we need an option to skip the product_id in the created invoice lines.

### DIFF
--- a/account_invoice_inter_company/models/account_invoice.py
+++ b/account_invoice_inter_company/models/account_invoice.py
@@ -43,6 +43,8 @@ class AccountInvoice(models.Model):
     @api.multi
     def _check_intercompany_product(self, dest_company):
         self.ensure_one()
+        if not dest_company.use_inter_company_products:
+            return
         domain = dest_company._get_user_domain()
         dest_user = self.env['res.users'].search(domain, limit=1)
         if dest_user:
@@ -250,6 +252,11 @@ class AccountInvoiceLine(models.Model):
                     % (self.env.ref('account.data_account_type_revenue').name,
                        dest_company.name, dest_company.id))
         tax_ids = dest_line_data.get('invoice_line_tax_ids', False)
+
+        product_id = False
+        if dest_company.use_inter_company_products:
+            product_id = self.product_id.id or False
+
         vals = {
             'name': self.name,
             # TODO: it's wrong to just copy the price_unit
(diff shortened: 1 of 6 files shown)


pedrobaeza (commented):
Curious use case, but modifications are correct and they don't mean any performance penalty. I think any way that this should be put at inter-company rule level instead, for better flexibility. Is it possible?

astirpe:
@pedrobaeza the point is that some users can have access to multiple companies, so products belonging to a different company than the one of the invoice could be used. In our environment we use different product for different companies. I cannot find a working solution by inter-company rule level.

pedrobaeza:
I'm talking about putting the "no product" option in the inter-company rule level, although a product mapping can be also a good addition.

astirpe:
I think I didn't understand then... What do you mean with inter-company rule? :thinking:

yung-wang:
@pedrobaeza I think the productmapping will be quite complex if you have a lot of products. We are working with 4 companies and mapping them would be a lot of work. For now I would suggest to keep it as simple as possible and leave it empty. 

Maybe in the future we can add a matrix with the products on the left side and on the top row the companies. And a good addition would be to have a flag on the products for intercompany usage, so the list of the to be mapped is shorter.

pedrobaeza:
OK, I see that there's no other place but company to put this. Then it's OK for me. You can mention in the roadmap the mapping part.

account_invoice_inter_company/models/account_invoice.py
```diff
@@ -250,6 +252,11 @@ def _prepare_invoice_line_data(self, dest_invoice, dest_company):
                     % (self.env.ref('account.data_account_type_revenue').name,
                        dest_company.name, dest_company.id))
         tax_ids = dest_line_data.get('invoice_line_tax_ids', False)
+
```
pedrobaeza: Please don't add empty lines inside methods code.

pedrobaeza (approved):

yung-wang (approved):
Works fine. All the lines are copied to the intercompany invoice without products

yung-wang:
Well it would be nice to add the possibility to create an invoice without any products. At this moment the system will raise an user error when I try confirm the invoice, and it wont create the inter-company invoice in the other company.
This is the user error I get:
```
The invoice line 'inter-company test invoice line' doesn't have a product. All invoice lines should have a product for inter-company invoices.
```

astirpe:
No runbot? :worried:

astirpe:
Failure of tests in travis seems not related with this module

PR #119, opened 2019-02-13
MIGR 12 invoice inter company

bealdav:


bealdav:
Please a review on it @astirpe @etobella @Yajo

account_invoice_inter_company/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2013-2014 Odoo SA
+# Copyright 2015-2017 Chafique Delli <chafique.delli@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Inter Company Invoices',
+    'summary': 'Intercompany invoice rules',
+    'version': '12.0.0.0.0',
```
astirpe:
```suggestion
    'version': '12.0.1.0.0',
```

account_invoice_inter_company/models/account_invoice.py
```diff
@@ -0,0 +1,282 @@
...
+    def _inter_company_create_invoice(self, dest_company):
+        """ create an invoice for the given company : it will copy
+                the invoice lines in the new invoice.
+            :param dest_company : the company of the created invoice
+            :rtype dest_company : res.company record
+        """
+        self.ensure_one()
+        # check intercompany product
+        self._check_intercompany_product(dest_company)
+        # if an invoice has already been genereted
```
astirpe:
```suggestion
        # if an invoice has already been generated
```

account_invoice_inter_company/models/account_invoice.py
```diff
@@ -0,0 +1,282 @@
...
+                ('company_id', '=', dest_company.id)], limit=1)
+            if not account:
+                raise UserError(_(
+                    'Please define %s account for this company: "%s" (id:%d).')
+                    % (self.env.ref('account.data_account_type_revenue').name,
+                       dest_company.name, dest_company.id))
+        tax_ids = dest_line_data.get('invoice_line_tax_ids', False)
+        product_id = False
+        if dest_company.company_share_product:
+            product_id = self.product_id.id or False
```
astirpe:
```suggestion
            product_id = self.product_id.id
```

account_invoice_inter_company/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,3 @@
+To configure this module, you need to go to the menu *Settings > General Settings*, select one of the companies and go to the tab *Multi-Companies* then check *Multi-companies*
```
astirpe:
```suggestion
To configure this module, you need to go to the menu *Settings > General Settings*, go to the tab *Multi-Companies* then check *Multi-companies*
```

account_invoice_inter_company/views/res_config_settings_view.xml
```diff
@@ -0,0 +1,25 @@
...
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="base_setup.res_config_settings_view_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//div[@id='multi_company']//field[@name='module_inter_company_rules']/../../div[2]" position="before">
+                <div class="o_form_label">Inter Company OCA features</div>
+            </xpath>
+            <xpath expr="//div[@id='multi_company']//field[@name='module_inter_company_rules']/../../div[2]" position="inside">
+                <span class="fa fa-lg fa-building-o" title="Values set here are company-specific." aria-label="Values set here are company-specific." groups="base.group_multi_company" role="img"/>
+                <div id="inter_company_invoice_validation">
+                    <field name="invoice_auto_validation" class="oe_inline" domain="[('company_id', '=', company_id)]"/>
```
astirpe: A domain on a boolean field is not useful.

```suggestion
                    <field name="invoice_auto_validation" class="oe_inline"/>
```

account_invoice_inter_company/views/res_config_settings_view.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="res_config_settings_view_form" model="ir.ui.view">
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="base_setup.res_config_settings_view_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//div[@id='multi_company']//field[@name='module_inter_company_rules']/../../div[2]" position="before">
```
astirpe, line 7: I tried to install this module locally and this view didn't show up in the configuration. But in Runbot is shown! I investigated a bit and the cause is that I locally had the module `remove_odoo_enterprise` installed (see https://github.com/OCA/server-brand/pull/12).

Not sure whether module `remove_odoo_enterprise` works the correct way. However, is it possible to make here the inheritance of this view not depending on field `module_inter_company_rules`? This way we avoid this kind of conflicts, due to the fact that the configuration depends on a field (`module_inter_company_rules`) that is directly connected with enterprise apps.

account_invoice_inter_company/models/res_company.py
```diff
@@ -7,14 +7,17 @@ class ResCompany(models.Model):
...
              "for this company, it will automatically validate it",
         default=True)
-    use_inter_company_products = fields.Boolean(
-        help="Use the same products when an invoice is created by "
-             "a multi company rule.",
-        default=True)
+
+    def _compute_share_product(self):
+        self.ensure_one()
+        product_rule = self.env.ref('product.product_comp_rule')
```
astirpe, line 19: In case the multicompany product rule is another one, different than `product.product_comp_rule`, the feature would not be working. Can't the old mechanism based on field `use_inter_company_products` be restored, instead of using this computed boolean `company_share_product`? Or perhaps another solution would be making the multicompany product rule selectable in the configuration, in case is other than `product.product_comp_rule`.

I have some doubts about this point :thinking:

astirpe (changes requested):
Only code review. I didn't test the functionality.

account_invoice_inter_company/models/res_company.py
bealdav: @astirpe thanks for your remark

If you have defined another rule to deal with product rule, I think the better for your project is to put in your custom code this field as readonly to avoid to break the erp behavior.
In this case no side effect.
You also may override the compute method to be compatible with the right rule : the purpose is just to active/unactive rules, then you can just do it without method or settings

I think it's better to plug into Odoo new settings than continue with an extra field for the same feature.

account_invoice_inter_company/views/res_config_settings_view.xml
bealdav: Thanks to notice this issue. I'll investigate

account_invoice_inter_company/models/res_company.py
astirpe: Of course I can override the compute methods and do whatever I want for my projects, but that's not the point. :wink: 

I was not thinking about adding an extra field, I was thinking about restoring the old mechanism and not using the `not bool(product_rule.active)` black magic at all. Could be confusing and is not even mentioned in the readme at the moment.

bealdav: Ok, I have no strong opinion here. 
I can mention in the readme, no problem
Let's see what others can think about that.

Please @pedrobaeza

pedrobaeza: Well, I see the intention of @bealdav removing that extra field for avoiding incoherences between one company and the other, and also reducing the number of configuration options. There's no special problem on the "hardcoded" rule, as you shouldn't have another multi-company rule, having this one marked as `noupdate` and being able to change it.

@astirpe if I don't remember bad, you were the one that added the inter company products feature. Is the granularity by company really needed?

astirpe: > Is the granularity by company really needed?

Could be possible in theory, it depends on the relations and agreements between companies I guess.

But I would prefer some functional expert to say an opinion about this (as I said, I have doubts about this part). 
I'm not blocking the PR because of this point. But at least I would like to see the new proposed behavior of this module explained in the readme.

account_invoice_inter_company/views/res_config_settings_view.xml
bealdav: @astirpe I tried to make modules compatible but I failed. I documented that in readme.
I think it's OK for you now. Thanks

account_invoice_inter_company/models/res_company.py
bealdav: @astirpe, as you know there'll have a lot of changes in Odoo v13 in multi-company, then I think we can move away with this code for now.
Time to merge ! ?
Don't you think ?

astirpe (approved):
Only code review (not tested).

bealdav:
@astirpe thanks

@pedrobaeza could you make a final review please.

Thanks

account_invoice_inter_company/models/account_invoice.py
```diff
@@ -0,0 +1,282 @@
...
+    @api.multi
+    def action_invoice_open(self):
+        """ Validated invoice generate cross invoice base on company rules """
+        res = super(AccountInvoice, self).action_invoice_open()
+        for src_invoice in self:
+            # do not consider invoices that have already been auto-generated,
+            # nor the invoices that were already validated in the past
+            dest_company = src_invoice._find_company_from_invoice_partner()
+            if not dest_company or src_invoice.auto_generated:
+                return res
```
mourad-ehm: s/return res/continue/g  use continue instead of return to allow managing multiple invoice validation (case when self contains inter company and simple invoice)

bealdav: @astirpe what do you think about @mourad-ehm proposal `continue` instead of `return res`

I think it's a good thing.

Thanks

astirpe: Yes indeed, must be continue. Thanks! :+1:

mourad-ehm:
```suggestion
                continue
```

mourad-ehm (approved):

Issue #109, opened 2018-08-30
[RFC] performances issues on product_multi_company used on huge database


Cedric-Pigeon:
Hi,

I am using product_multi_company on a very big database containing more than 110.000 products spread over 5 companies.

The problem I am facing is bad performance when accessing product.product model.
The record rule is based on a many2many field but when reading through an inherits relation it is not well optimized. We get huge subqueries with expression `id in`.

We are planning to add a technical field on product template as a string expression of the many2many content.

For exemple= companies_technical = "-1-, -3-, -13-"

The goal is to have the record rule like :

[('companies_technical', 'ilike', str('-' + user.company_id.id + '-')]

The childof in the current record rule will be managed by a search method on the technical field.

We tested this patch and it reduced response time by 6 on some processes.

Does anyone has an opinion on that ?  A better idea ?

I am currently preparing a PR with what we did.

Cedric-Pigeon:
@sbidoul @lmignon

pedrobaeza:
Can `auto_join` help?

lmignon:
@pedrobaeza The problem seems to be related to https://github.com/odoo/odoo/issues/26702
`company_ids` is already declared with `auto_join=True`

pedrobaeza:
OK, I see. Then it's OK to go to that patch. It's similar to the parent_left/parent_right deprecation done on master.

sbidoul:
@pedrobaeza @lmignon @Cedric-Pigeon there is a better solution I think. We can optimize the search for company_ids=False if we have a stored flag 'visible_for_all_companies' that is set to True when company_ids is empty, and use it in the record rule.

pedrobaeza:
OK, that can be another solution that seems cleaner.

PR #123, opened 2019-05-13
[12.0] migrage base multi company + improve performance

florian-dacosta:
Standard migration + improve performance applying idea proposed by @sbidoul in https://github.com/OCA/multi-company/issues/109

@Cedric-Pigeon


tbaden (changes requested):
please use the fragment [readme](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0)

florian-dacosta:
@tbaden There is no "Implementation" section and I don't really see a suitable section in the readme files for this part. Any idea?

pedrobaeza:
Put it inside usage section with `~~~~` underlining.

florian-dacosta:
@pedrobaeza Thanks for the hint, I'll do that

florian-dacosta:
@tbaden Readme has been split.

tbaden (approved):
LGTM 👍

LoisRForgeFlow (approved):
LGTM

base_multi_company/hooks.py
```diff
@@ -0,0 +1,85 @@
...
+
+    :param: env: Environment
+    :param: rule_ref: XML-ID of the security rule to change.
+    """
+    rule = env.ref(rule_ref)
+    if not rule:  # safeguard if it's deleted
+        return
+    rule.write({
+        'active': True,
+        'domain_force': (
```
kittiu, line 34: @florian-dacosta I am doing https://github.com/OCA/multi-company/pull/136 which has error from this rule writing. In partner_multi_company, it will error when I create a new company which also create res.partner.
Can you change to following, because normally we need to have term "or company_id = False" to avoid that error.
```suggestion
        'domain_force': (
            "['|', '|', ('company_ids', 'in', user.company_id.ids),"
            " ('visible_for_all_companies', '=', True), ('company_id', '=', False)]"
        ),

```

pedrobaeza: I think now that a better technique is to duplicate the rule, and deactivate the old one, and on uninstall, delete the new one, and activate the old.

kittiu: Allow me to confirm my understanding,
The domain_force in this **base_multi_company** should stay like this, and the changes per mentioned goes to **partner_multi_company**, correct?

kittiu: Also, this one can be merged soon to fix dependency? :)

florian-dacosta: @pedrobaeza I think the problem is not the modification of the rule, but the value in domain_force that raise an error when it should not, in a specific case.
It is True that it could be cleaner to do it this way (activate/deactivate) but it won't change this particular problem.

@kittiu I checked a little, I am not sure why it fails exactly, and only in this particular case.
It seems the `visible_for_all_companies` field is computed too late...
Anyway, I don't think that changing the domain for all models/modules is the good solution.
Indeed, the changes have been made to improve the poor performance of this rule, and adding `('company_id', '=', False)` once again revert the performance improvments.

So the domain could eventually be changed in partner_multi_company, to fix this very particular problem, but it will have an impact on performance.

Another ideal which could be implemented in this base module is to add a default value True to `visible_for_all_companies` field.
This way, the record rule won't fail during creation if `visible_for_all_companies` is not  computed yet.
And if it is computed, it will get the right value...
@pedrobaeza any comment about this?

kittiu: @florian-dacosta as you mentioned about default=True, I try add 2 lines in your abstract class, and it works fine.
image

florian-dacosta: @kittiu Actually the second line is not necessary.
I made the change

florian-dacosta:
Test are failing, but it seems it is not because of this module.
Still, I don't understand why...

kittiu:
@florian-dacosta I also can't find the exact problem. But I found that, if we run test_multi_company_abstract, we will get this error,
```
2019-06-24 05:34:21,915 5396 CRITICAL odoo12_base_multi_company odoo.service.server: Failed to initialize database `odoo12_base_multi_company`. 
  File "/home/kittiu/workspace/odoo12/odoo/service/server.py", line 1116, in preload_registries
    registry = Registry.new(dbname, update_module=update_module)
  File "/home/kittiu/workspace/odoo12/odoo/modules/registry.py", line 86, in new
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/home/kittiu/workspace/odoo12/odoo/modules/loading.py", line 497, in load_modules
    registry.init_models(cr, list(models_to_check), {'models_to_check': True})
  File "/home/kittiu/workspace/odoo12/odoo/modules/registry.py", line 301, in init_models
    func()
  File "/home/kittiu/workspace/odoo12/odoo/models.py", line 281, in _reflect
    self.env['ir.model']._reflect_model(self)
  File "/home/kittiu/workspace/odoo12/odoo/addons/base/models/ir_model.py", line 274, in _reflect_model
    ids = query_update(cr, self._table, params, ['model'])
  File "/home/kittiu/workspace/odoo12/odoo/addons/base/models/ir_model.py", line 65, in query_update
    cr.execute(query, values)
  File "/home/kittiu/workspace/odoo12/odoo/sql_db.py", line 147, in wrapper
    raise psycopg2.OperationalError(msg)
psycopg2.OperationalError: Unable to use a closed cursor.

```

florian-dacosta:
Well I was wrong, the failing was really because of this module.
I did not meet any error in local using pytests, that's why I did not see it.
Anyway, I have fixed the issue, I think this PR is now ready to merge.
@kittiu Could do make a review?

kittiu:
@florian-dacosta thanks for the fix. Now it works, and also works with partner_multi_company too https://github.com/OCA/multi-company/pull/136

florian-dacosta:
@kittiu 
Since you seem satisfied with the migration ,could you approve the PR?

kittiu:
@florian-dacosta how to approve, I don't have rights :p

florian-dacosta:
@kittiu Of course you have!
If you click on "file changed", on the top of the PR, to see the changes made in the PR, you will see a button named "Review changes".
Then you can comment, approve, or request changes, depending if you think it is all good or not.

kittiu (approved):

florian-dacosta:
@pedrobaeza Could you merge this?

amkarthik:
@pedrobaeza @florian-dacosta 
It seems that this fix is missing v13 branch
Any specific reason for that?

pedrobaeza:
This is not a fix, but a migration PR. Anyway, you can do a PR with anything you consider for improving the module.

amkarthik:
@pedrobaeza 
Thanks

PR #126, opened 2019-05-17
[12.0][MIG] account_multicompany_easy_creation

kittiu:
Work in Progress


kittiu:
@pedrobaeza , after looking in more detail. Seem that v12 are quite different than v11. I.e., wizard.multi.charts.accounts is also missing now.
As I am not very familiar with this feature, I would like to withdraw from this PR for now.

pedrobaeza:
OK

PR #140, opened 2019-07-10
[12.0][ADD] - Add multi-company management to mail templates

sbejaoui:


mail_template_multi_company/models/mail_template.py
```diff
@@ -0,0 +1,18 @@
...
+
+    _inherit = 'mail.template'
+
+    company_id = fields.Many2one(
+        'res.company',
+        'Company',
+        default=lambda self: self.env['res.company']._company_default_get(
+            'mail.template'
+        ),
+        ondelete='cascade',
```
sbidoul: I'd do `set null` instead of cascade.

mail_template_multi_company/models/mail_template.py
```diff
@@ -0,0 +1,18 @@
...
+class MailTemplate(models.Model):
+
+    _inherit = 'mail.template'
+
+    company_id = fields.Many2one(
+        'res.company',
+        'Company',
+        default=lambda self: self.env['res.company']._company_default_get(
+            'mail.template'
+        ),
```
sbidoul, line 16: We should not set the company by default on all existing mail templates when installing this module.
Because many standard template need not be specialized by company.

OTOH, when this module is installed it makes sense that *new* templates being created by users have a company set by default.

sbidoul (changes requested):

mail_template_multi_company/security/mail_template.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+    <data>
```
legalsylvain, line 6: I think that the default rule in Odoo is to set ``ir.rule`` defined in security folder as "noupdate="1"" to allow user to change it by the UI.

https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/security/base_security.xml#L7

mail_template_multi_company/views/mail_template.xml
```diff
@@ -0,0 +1,40 @@
...
+        <field name="inherit_id" ref="mail.email_template_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='model_id']" position="after">
+                <field name="company_id" options="{'no_open': True, 'no_create': True}" groups="base.group_multi_company"/>
+            </xpath>
+        </field>
+    </record>
+
+    <record model="ir.ui.view" id="mail_template_search_view">
+        <field name="name">mail.template.search (in mail_template_multi_company)</field>
```
legalsylvain, line 19: not a blocking point. ``name field`` is useless because autogenerated by Odoo by default.

mail_template_multi_company/views/mail_template.xml
```diff
@@ -0,0 +1,40 @@
...
+        <field name="inherit_id" ref="mail.view_email_template_search"/>
+        <field name="arch" type="xml">
+            <xpath expr="//group" position="inside">
+                <field name="company_id" groups="base.group_multi_company"/>
+            </xpath>
+        </field>
+    </record>
+
+    <record model="ir.ui.view" id="mail_template_tree_view">
+        <field name="name">mail.template.tree (in mail_template_multi_company)</field>
```
legalsylvain, line 30: same here.

legalsylvain (changes requested):
some minor remarks. LGTM otherwise.
thanks a lot !

RobinetDenisAcsone (approved):
LGTM

mail_template_multi_company/security/mail_template.xml
sbejaoui: Hi @legalsylvain ,
Thank you for your review,

User can archive an ir.rule and create an alternative.

mail_template_multi_company/views/mail_template.xml
sbejaoui: we prefer add the module name in the inherited view name, so we can easily identify it when debugging from the UI

sbejaoui:
@sbidoul and @legalsylvain , can you update your review please

qgroulard:
+1 :+1:

qgroulard (approved):

mail_template_multi_company/views/mail_template.xml
sbidoul: The module name is also easily visible in 12, because the xmlid is displayed both in the tree view and for view of ir.ui.view. So I see some value in letting Odoo computing the name itself.
Perhaps we should change the mrbob template.

mail_template_multi_company/security/mail_template.xml
sbidoul: @sbejaoui @legalsylvain is there a written OCA convention for noupdate in record rules?

legalsylvain: Hi @sbidoul. Well, not an OCA convention, AFAIK. for my side, just a recurrent Odoo rule, I observed. 
All the rule in Odoo are in noupdate="1" (and all the modules in OCA I use and / or developped).

ref : product module in V13 : (for exemple) 

https://github.com/odoo/odoo/blob/13.0/addons/product/security/product_security.xml#L32

My preference is to use this "convention". don't you think ?

I don't know who ping in Odoo team to know the reason of this choice.

PR #143, opened 2019-07-17
[FIX] account_invoice_inter_company: Compatibility with password_security

yajo:
`password_security` asks by default for 8 chars minimum in the password. If it happens to be installed before this module, it cannot be loaded.

It's just demo data, so I just use a longer password here.

@Tecnativa


pedrobaeza (approved):
Very straight-forward, so fast-tracking this

PR #148, opened 2019-08-06
[12.0][FIX] Fix travis (as per new template)

astirpe:


pedrobaeza:
It seems there's a problem.

astirpe:
@pedrobaeza perhaps another temporary issue? Is it possible to retry?

PR #149, opened 2019-08-08
[11.0][IMP]  base_multi_company: add hook in computing the record's company

MiquelRForgeFlow:
Let's have room to be able to customize better. Someone may want to filter the records by some specific models in which apply the updating and not do it in all of them.


MiquelRForgeFlow:
ping: @jbeficent

PR #150, opened 2019-08-08
[11.0][FIX]  partner_multi_company: update partner company_ids only if the partner is not global

MiquelRForgeFlow:
1. You create a user. Assign the user to several companies.
2. Go the the user's partner and delete the company_ids of the partner.
3. Log with the user and change the company of the user.
4. Error: Check again the company_ids of the user's partner and you find that the field have been filled again.

We should maintain the same behavior as odoo: https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/res/res_users.py#L369:
```python
# if partner is global we keep it that way
```

### DIFF
--- a/partner_multi_company/models/res_users.py
+++ b/partner_multi_company/models/res_users.py
@@ -21,5 +21,7 @@ class ResUsers(models.Model):
         res = super(ResUsers, self).write(vals)
         if 'company_id' in vals:
             for user in self.sudo():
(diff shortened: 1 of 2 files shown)


JordiBForgeFlow (approved):

HviorForgeFlow (approved):
Makes sense.

In which use case we would be benefited on this change?

PR #156, opened 2019-09-30
[ADD][10.0] web_company_from_domain_name

thomaspaulb:


company_from_url/models/ir_rule.py
```diff
@@ -0,0 +1,74 @@
...
+    _inherit = 'ir.rule'
+
+    ''' COMPUTE DOMAIN PER COMPANY'''
+
+    # /base/ir/ir_rule.py
+    # This did not work:
+    # @api.model
+    # @tools.ormcache('self.env.uid', 'self.env.user.company_id', 'model_name', 'mode') #UPDATED
+    # def _compute_domain(self, model_name, mode="read"):
+    #     return super(IrRule, self)._compute_domain(model_name, mode=mode)
```
thomaspaulb: @norlinhenrik I don't understand enough about `ormcache` to know how to make the above example work, but I think it's still worth trying. Maybe someone could assist us

company_from_url/models/ir_rule.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+    # https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/ir/ir_rule.py
+    # - @tools.ormcache('self._uid', 'model_name', 'mode')
+    # + @tools.ormcache('self.env.uid', 'self.env.context.get("company_id")', 'model_name', 'mode')
+    @api.model
+    @tools.ormcache('self.env.uid', 'self.env.user.company_id', 'model_name', 'mode')
```
thomaspaulb: @norlinhenrik Instead of `self.env.context.get("company_id")` I made this with `self.env.user.company_id`, which also seems to work, and then I could comment out the context-related code in `res_users.py`. Do you think that will lead to problems?

company_from_url/models/res_users.py
```diff
@@ -0,0 +1,87 @@
...
+
+    company_id = fields.Many2one(
+        compute='_compute_company_id',
+        inverse='_inverse_company_id',
+        string='Company (from URL)',
+        store=False, index=False, required=False)
+    stored_company_id = fields.Many2one(
+        'res.company', string='Company (stored)', readonly=True)
+
+    ''' PORTAL USER - rather use 'share' field... '''
```
thomaspaulb: @norlinhenrik What did you try to achieve here?

thomaspaulb:
@norlinhenrik I did some polishing of the code and removed the 'website' dependency. It now works for me, but I still would like to do some optimizations as discussed above, add a README, and also make the module for other versions of Odoo.

company_from_url/models/ir_rule.py
thomaspaulb: @hbrunn If you have time: my technical question here would be how to successfully override a value for `tools.ormcache` that was set on the parent function, in the child function. I have studied the code in `cache.py` in Odoo [link](https://github.com/OCA/OCB/blob/10.0/odoo/tools/cache.py#L33) and if I could somehow always clear the cache before the super() call, I think I would have solved my issue, ..... I'll keep on trying.

thomaspaulb: I found something that [works](https://github.com/OCA/multi-company/pull/156/commits/2f316b5e986f5a0b767d8478519c8216072e59bc), but it doesnt seem very efficient - it often recomputes the cache, IMO too often.

hbrunn: yes, you do this with every call, that's very inefficient, specifically because you clear all orm caches by clearing some ormcached function's cache if I read https://github.com/OCA/OCB/blob/10.0/odoo/tools/cache.py#L96 correctly. The net result then is that you have no caching at all for this function. Your specific case you can fix by just calling `_super._func_` (that's where the decorator lib that the cache uses puts a reference to the function the decorator wraps), as you want to completely circumvent the original caching. This can break if there's an override in between that's not decorated with the cache decorator at all, but you'll have to trust people don't break inheritance I guess.

To do this right, the framework should do the unwrapping when carving classes, but this also might be unwanted if we want super to be cached. Difficult.

company_from_url/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import api, exceptions, fields, models
+
+import logging
+_logger = logging.getLogger(__name__)
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    access_url = fields.Char(string="Direct access URL")
```
hbrunn: what you fill inhere is not actually an URL but a domain name. If you can, the addon itself should be renamed too

company_from_url/models/res_users.py
```diff
@@ -0,0 +1,87 @@
...
+                company = None
+        for this in self:
+            this.company_id = \
+                company or this.stored_company_id or this.company_ids[:1]
+
+    company_id = fields.Many2one(
+        compute='_compute_company_id',
+        inverse='_inverse_company_id',
+        string='Company (from URL)',
+        store=False, index=False, required=False)
```
hbrunn: this will cause you a lot of trouble because there will be 10k scripts/addons that expect the database column to exist. Better override `read` and if `request` is not falsy (that's much cheaper to check than catching the runtime error), set your company in the result dict. That's cached within the current environment which should be good enough for starters, you could also set this in some property in the request, then for every request you only search once.

hbrunn:
probably you'll need a more general solution for the caching, as many things will turn out to be company dependent. Better override https://github.com/OCA/OCB/blob/10.0/odoo/addons/base/ir/ir_http.py#L174 (this is where all request to Odoo go through), save on `request.env.registry` which company this worker has served before, if it's not the same, clear all caches.

Extra points for bending werkzeug's dispatching to prefer workers that have served the same company before, but that's going to be very hard next to impossible.

company_from_url/models/res_company.py
thomaspaulb: `company_from_domain_name` or `web_company_from_domain_name` should be better then.

thomaspaulb:
@hbrunn Thank you for the review, this is very helpful.

I agree with your last suggestion except that it should be "which uid-company combination this worker has served before" - different users should be able to click around in different companies without their cache invalidating all the time. The number of logged in users could well exceed the number of workers.

Smart choosing of workers should not be necessary, at least not in the use case that I foresee - when one user decides to switch to his other browser tab and click around in another company, he'll usually be sticking around there for enough requests to have justified changing all his cache entries to the other company.

company_from_url/models/ir_rule.py
thomaspaulb: I have resorted to the below suggestion to clear caches when a user suddenly decides to use a different access domain. This seems to work.

company_from_url/models/res_company.py
thomaspaulb: Have renamed this.

company_from_url/models/res_users.py
thomaspaulb: I have [tried](https://github.com/sunflowerit/multi-company/pull/1/files) to do this, but for some reason the cached values of user.company_id bite me in the case of using a `read` override, but not when using a computed field.

hbrunn: I'd suggest to debug this, there's no caching involved I'm aware of. But your approach also looks quite needslessly complicated to me: Ditch the `stored_company_id` field completely, and just do something like:

```python
result = super()
if fields and 'company_id' not in fields:
    return result
for values in result:
    values['company_id'] = this._get_company_id_from_domain()
return result
```

web_company_from_domain_name/models/ir_http.py
```diff
@@ -0,0 +1,33 @@
...
+class IrHttp(models.AbstractModel):
+    _inherit = 'ir.http'
+
+    @classmethod
+    def _dispatch(cls):
+        """ Clear cache when a user switches to another domain name """
+        uid = request.session.uid
+        domain_name = request.httprequest.host.partition(":")[0]
+        cached_domain_name = request.registry['res.users']. \
+            get_cached_domain_name(uid, domain_name)
```
hbrunn: you can just access `request.registry.cache`, and invent some unique key like `('ir.http', 'last_company', uid)` - if this exists and isn't your current company, you need to clear the cache

web_company_from_domain_name/models/ir_http.py
```diff
@@ -0,0 +1,33 @@
...
+        # Dont trigger on longpolling because this occurs too often
+        if domain_name != cached_domain_name and not \
+                'longpolling' in request.httprequest.path:
+            # TODO: only clear cache for this user, not all?
```
hbrunn: odoo ditched the concept of partial clearing of caches, all those functions clear the cache, because we only have one (`registry.cache`) - in older times, this was on the model object, that's where the function names come from

web_company_from_domain_name/models/ir_http.py
```diff
@@ -0,0 +1,33 @@
...
+            request.registry.clear_caches()
+            request.env.invalidate_all()
```
hbrunn: ah, you mean this. Makes sense to only invalidate envs with the current user's uid

web_company_from_domain_name/models/ir_http.py
```diff
@@ -0,0 +1,33 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class IrHttp(models.AbstractModel):
+    _inherit = 'ir.http'
+
+    @classmethod
+    def _dispatch(cls):
+        """ Clear cache when a user switches to another domain name """
+        uid = request.session.uid
```
hbrunn: note that this is not necessarily set

web_company_from_domain_name/models/ir_http.py
thomaspaulb: Would that be better than this? I thought this was kind of neat; `get_cached_domain_name` has an `ormcache` decorator that defines just what you describe

hbrunn: I find this a little bit indirect, but if you like your cached function better, keep it

company_from_url/models/res_users.py
thomaspaulb: I'll debug more, the caching issue was located [here](https://github.com/OCA/OCB/blob/10.0/addons/web/models/ir_http.py#L38)

The `stored_company_id` field is unnecessary, it was intended more like a feature, so you can still see which company the user would get if he arrives through a non-company-forcing domain name. I can remove it for now.

thomaspaulb: Finally got this; `read_from_database` was already filling the cache, so I had to override that instead of `read`.

company_from_url/models/res_company.py
thomaspaulb: Or how about `multicompany_from_domain_name` ?

company_from_url/models/res_users.py
hbrunn: huh, sorry, didn't think of this. Thanks for the pointer

company_from_url/models/res_company.py
hbrunn: whatever you prefer as long as the name doesn't lie. Some repo maintainers indeed prefer the name of the repo as prefix (that would count against `web`), others including me choose the name based on the addon the current one is an extension/modifiction of, that would then make it `base_company_from_domain_name`. If in doubt, ask Pedro ;-)

web_company_from_domain_name/models/res_company.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo import api, exceptions, fields, models
+
+import logging
+_logger = logging.getLogger(__name__)
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    access_url = fields.Char(
```
hbrunn: shouldn't you rename the field too?

hbrunn:
if there's no other PR for this already, you can fix travis by updating it from the latest MQT template. For your tests, you can either do something like https://github.com/OCA/server-auth/pull/131/files#diff-c2041f2f2bc5a25a8ce303fb25923980R22 which allows you to do convenient assertions, or you override https://github.com/OCA/OCB/blob/10.0/odoo/tests/common.py#L248 and do some request that tells you the current company

company_from_url/models/res_company.py
thomaspaulb: @pedrobaeza What do you think? `multicompany_from_domain_name` sounds most descriptive to me

PR #161, opened 2019-10-11
[12.0][MIG] account_multicompany_easy_creation

ernestotejeda:
Cc @Tecnativa
TT19361


account_multicompany_easy_creation/wizards/multicompany_easy_creation.xml
```diff
@@ -66,6 +65,7 @@
                 <page string="Banks">
                     <field name="bank_ids">
                         <tree editable="bottom">
+                            <field name="partner_id"/>
```
pedrobaeza: You shouldn't present the `partner_id`. You need to put the new company's partner one.

pedrobaeza (changes requested):

carlosdauden (approved):
Attend this:
https://github.com/OCA/multi-company/pull/161#pullrequestreview-301712237

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -0,0 +1,341 @@
...
+        self.install_chart_account()
+        self.create_bank_journals()
+        self.create_sequences()
+
+    # TODO: Cache don't work
+    @ormcache('self.id', 'company_id', 'match_tax_ids')
+    def taxes_by_company(self, company_id, match_tax_ids):
+        AccountTax = self.env['account.tax'].sudo()
+        taxes_ids = []
+        for tax in AccountTax.browse(match_tax_ids):
```
Tardo: Perhaps can do it in a single "search" call using OR operator...

account_multicompany_easy_creation/wizards/multicompany_easy_creation.xml
ernestotejeda: I set self.env.user.company_id.partner_id as the default value for the partner_id field. I did it because it doesn't matter what value you set as default, as I could understand from the class account.multicompany.bank.wiz only the acc_number field is used

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
ernestotejeda: Thanks @Tardo , I did it in a single "search" call , check it out please

account_multicompany_easy_creation/wizards/multicompany_easy_creation.xml
pedrobaeza: That's incorrect. You have to put the new company's partner, not the current company one.

ernestotejeda: I forgot to state that I hid the partner_id field.
At the moment when the wizard appears there is no new company, when you click on the Accept button of the Wizard is when the new company is created based on the data established in the Wizard, so I have no way to access to the new company.
On the other hand, the partner_id field we are talking about belongs to another wizard with which the main Wizard has a One2many relationship, that is, it is a Transient model, which is not intended to be stored for a long time. 
Captura de pantalla de 2019-10-22 23-22-29
As i can see in de code, that other wizard is only used for the user to enter in the 'Account number' field the name of a Bank-type Journal that the new company will have when it is created. 
Captura de pantalla de 2019-10-22 23-29-35
Maybe @carlosdauden can give his opinion here.

pedrobaeza: You have been the one that has added `partner_id` field there, not @carlosdauden. Why don't you simply remove that field?

ernestotejeda: That's what I meant in the previous comment, hehe

> I forgot to state that I hid the partner_id field.

i'm sorry, I didn't express myself well, I already removed that field, but as it is now required in v12, we must set a default value so that it does not appear in the view and there is no error when clicking on accept but I have set that value as I have previously stated.

ernestotejeda:
Changes done

account_multicompany_easy_creation/wizards/multicompany_easy_creation.xml
carlosdauden: If you need partner_id here:
https://github.com/OCA/multi-company/pull/161/files?file-filters%5B%5D=.py&file-filters%5B%5D=.xml#diff-612c33ffbeca5455f8bb88183566b383R137
You can use self.new_company_id.partner_id.id

carlosdauden: I understand:
https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/res/res_bank.py#L65
https://github.com/odoo/odoo/commit/7b322d08de2b5343cdf6935a7ae4a12f1fd61958
https://github.com/odoo/odoo/blob/e1aee2aac2fd2c19dfd55734b1f51aedbc0393a6/odoo/addons/base/models/res_bank.py#L84

https://github.com/OCA/multi-company/pull/161/files#diff-612c33ffbeca5455f8bb88183566b383R333

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -0,0 +1,342 @@
...
+class AccountMulticompanyBankWiz(models.TransientModel):
+    _inherit = 'res.partner.bank'
+    _name = 'account.multicompany.bank.wiz'
+    _order = 'id'
+
+    wizard_id = fields.Many2one(
+        comodel_name='account.multicompany.easy.creation.wiz',
+    )
+    partner_id = fields.Many2one(
+        default=lambda self: self.env.user.company_id.partner_id
```
carlosdauden:
```suggestion
        required=False
```

carlosdauden: This is less confusing :wink:

pedrobaeza: Yeah, that's what I was about to propose, but you need to add the company's partner in the dictionary preparation before calling the creation.

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -0,0 +1,342 @@
...
+            ('company_id', '=', self.new_company_id.id),
+        ])
+        vals = {
+            'type': 'bank',
+            'company_id': self.new_company_id.id,
+        }
+        for i, bank in enumerate(self.bank_ids):
+            vals.update({
+                'name': bank.acc_number,
+                'bank_acc_number': bank.acc_number,
```
carlosdauden, line 139:
```suggestion
                'bank_acc_number': bank.acc_number,
                'partner_id': self.new_company_id.partner_id.id,
```

ernestotejeda:
Changes done

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
pedrobaeza: This is creating the account.journal, not the res.partner.bank record, so not needed. It's handled automatically by Odoo code.

pedrobaeza (approved):

Issue #84, opened 2018-02-10
display_name becomes non searchable once this module is installed


farazbdm:
this module makes the partner display_name field to be non-stored and thus the search function of partner on the display_name becomes invalid resulting in following error:

ERROR database odoo.osv.expression: Non-stored field res.partner.display_name cannot be searched.

pedrobaeza:
Which module?

pedroortega-punt:
Buenos días,

@pedrobaeza 

Cuando se instala el módulo "partner_multi_company" la búsqueda por nombre en clientes deja de funcionar, y en la vista tree no deja ordenar por nombre.

pedroortega-punt:
V10

pedrobaeza:
Only in English, please. I'm looking at the module and dependencies and it doesn't change anything about display_name.

pedroortega-punt:
Good Morning,

@pedrobaeza

When the "partner_multi_company" module is installed, the search by name in clients stops working, and in the tree view it does not allow to sort by name.

pedrobaeza:
As already said, there's seems to be no relation on that, as neither name nor display_name fields are touched in this module.

pedroortega-punt:
@pedrobaeza 

I'm not a programmer, they're a consultant. But I can assure you that at the moment you install the partner_multi_company module, the search by name and sort by name will stop working.

Also if you uninstall the module after associating several companies in a partner and want to install again, you get a message that can not be installed because a partner has two companies, so I infer that the relationship table does not eliminate the ids when it is uninstalled, then it does not let you install any more.

pedrobaeza:
You can try in http://3326111-10-0-f60542.runbot1.odoo-community.org/web#min=1&limit=80&view_type=list&model=res.partner&menu_id=112&action=56, that is the current 10.0 OCA runbot, where the module is installed, and everything is working fine, so please be sure before posting issues about what is the real problem.

pedroortega-punt:
@pedrobaeza 

Sorry,

I had checked it perfectly and it does not work, what should have been detailed in which menu, in the accounting / clients if it works, in the sales / clients does not work.

pedroortega-punt:
You can try in http://3326111-10-0-f60542.runbot1.odoo-community.org/web#min=1&limit=80&view_type=list&model=res.partner&menu_id=189&action=55

pedrobaeza:
OK, reopening the issue, but I haven't checked the root cause.

[...]

PR #165, opened 2019-10-22
[11.0][FIX] partner_multi_company: make display_name stored

LoisRForgeFlow:
Fixes https://github.com/OCA/multi-company/issues/84.

This is needed because after installation this field becomes unsearchable and unsortable. Which is not explicitly changed in this module and as such can be considered an undesired yield.

it is probably not a fix for the root cause but it is at at least a fix for a long-standing problem...


pedrobaeza (approved):
Let's go with this solution, which is the same as previous @igallart's one, but that PR mixed 2 unrelated things. Though we don't still know the root cause (I guess it's a bad interaction between inheritance by delegation and the abstract model).

Fast-tracking it.

PR #178, opened 2019-11-07
[FIX] Travis errors

max3903:


base_multi_company/__manifest__.py
```diff
@@ -15,4 +15,6 @@
     'data': [
         'security/ir.model.access.csv',
     ],
+    'development_status': 'Beta',
```
pedrobaeza: This one can be considered `Production/Stable` after 4 versions and with tests

pedrobaeza (approved):
Let's fast-track this

PR #187, opened 2020-02-04
12.0 FIX purch_sale_inter_cpny: split data according to companies

bealdav:
when creating sale.order

cc @chafique-delli @mourad-ehm @carlosdauden 

There is the same bug in v11
https://github.com/OCA/multi-company/blob/11.0/purchase_sale_inter_company/models/purchase_order.py#L81-L82

### DIFF
--- a/purchase_sale_inter_company/models/purchase_order.py
+++ b/purchase_sale_inter_company/models/purchase_order.py
@@ -78,8 +78,12 @@ class PurchaseOrder(models.Model):
         # create the SO and generate its lines from the PO lines
         sale_order_data = self._prepare_sale_order_data(
             self.name, company_partner, dest_company, self.dest_address_id)
-        sale_order = self.env['sale.order'].sudo(
-            intercompany_user.id).create(sale_order_data)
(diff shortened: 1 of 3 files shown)


purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -78,8 +78,13 @@ def _inter_company_create_sale_order(self, dest_company):
         # create the SO and generate its lines from the PO lines
         sale_order_data = self._prepare_sale_order_data(
             self.name, company_partner, dest_company, self.dest_address_id)
+        po_id = sale_order_data.pop("auto_purchase_order_id")
         sale_order = self.env['sale.order'].sudo(
             intercompany_user.id).create(sale_order_data)
+        # You can't set data from different companies with only the rights
+        # of only one of theses companies
+        # if you don't use parent_id in company
+        sale_order.sudo().write({"auto_purchase_order_id": po_id})
```
bealdav: I think it could be better to use base_suspend_security because sudo(intercompany_user.id) doesn't apply sudo rights in v12 at least

PR #190, opened 2020-02-27
[11.0][FIX]  partner_multi_company: If user switch company, don't delete user's partner's other companies

MiquelRForgeFlow:
Fixes https://github.com/OCA/multi-company/issues/182

PR #191, opened 2020-02-27
[12.0][FIX] partner_multi_company: If user switch company, don't delete user's partner's other companies

MiquelRForgeFlow:
Forward port of https://github.com/OCA/multi-company/pull/190.


LoisRForgeFlow (approved):
LGTM

PR #193, opened 2020-03-06
[11.0][FIX] account_multicompany_easy_creation: account code digits and currency ignored

eantones:
Account code digits and currency are overwrited by onchange_chart_template_id method

### DIFF
--- a/account_multicompany_easy_creation/README.rst
+++ b/account_multicompany_easy_creation/README.rst
@@ -42,6 +42,7 @@ Contributors
 
 * `Tecnativa <https://www.tecnativa.com>`_:
   * Carlos Dauden
+* Eric Antones - NuoBiT Solutions, S.L. <eantones@nuobit.com>
 
 Do not contact contributors directly about support or help with technical issues.
 
--- a/account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
+++ b/account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
@@ -135,6 +135,9 @@ class AccountMulticompanyEasyCreationWiz(models.TransientModel):
                 self.chart_template_id.cash_account_code_prefix,
         })
         wizard.onchange_chart_template_id()
+        if self.accounts_code_digits:
+            wizard.code_digits = self.accounts_code_digits
+        wizard.currency_id = self.currency_id
         wizard.sudo().execute()
 
     def create_bank_journals(self):


rafaelbn:
@chienandalu @pedrobaeza take a look here please

pedrobaeza:
For me both values must be removed as they are not possible to be selected when installing it normally (and in v12 code digits disappear from res.company).

eantones:
@rafaelbn @pedrobaeza I needed to do this fix because my customer was not able to create companies with more than 6 digits, now he can and he already did it perfectly using 9 digits.
Independently that it makes no sense in v12, it does in v11 and since it's clearly a bug, it should be fixed. Thanks

pedrobaeza:
That's the mistake: to create CoA with more than 6 digits. It's totally unneeded and will be conflicting on v11 or v12. This is the new myth to be doomed as the accounts per customer/supplier. Both aren't needed. The only remnant is for assets, and with latest changes to account_asset_management you mustn't create one account per asset as well. The rest is to swim against current...

pedrobaeza:
For me, the patch is to remove the digits from the wizard, but my colleague @carlosdauden that created the module probably differs, as he doesn't think the same about the usefulness of the digits, so I let him to make the final call.

eantones:
@pedrobaeza I agree, no need more than 6 digits, but, as often, it's not black and white, due to the complexity of my customer's accounting it's too risky to try to redefine everything in a switch off-switch on context, it takes time and in the meantime they should be able to keep their activities going on with more than 6 digits. Anyway and aside of that, there's a bug in the code and it must be fixed. Thanks!

pedrobaeza:
Yeah, removing the field is another bugfix :wink:

eantones:
Yes, it certainly is :-) but I have to solve my customer issues one way or another.

pedrobaeza:
Yes, if you use a flexible deployment like Doodba, you can deploy with this PR included in it without the need of merging, so this discussion and the final answer from my colleague shouldn't block you.

eantones:
Nothing blocks me, I already solved my customer problem using this PR. I just wanted to share the fix to others. Thanks

pedrobaeza:
OK, let's wait what my colleague thinks.

eantones:
Perfect :)

carlosdauden (approved):
Being in agreement with Pedro's arguments, I weigh pros and cons of this PR:
PROS | CONS
---- | ----
Help some users who use a different value in v11 | None (This change does not affect those who use 6 digits)

so for my part I approve the PR

pedrobaeza:
OK, let's go then

/ocabot merge patch

PR #197, opened 2020-03-30
[13.0][MIG] base_multi_company

JoanSForgeFlow:
**base_multi_company** migration to v13

model 'res.company.assigment' has been removed, migration scripts will be made in product_multi_company and partner_multi_company.

cc ~ @LoisRForgeFlow @JordiBForgeFlow

PR #198, opened 2020-03-30
[13.0][MIG] product_multi_company

JoanSForgeFlow:
**product_multi_company** migration to v13.
Depends on:
- #197 

cc ~ @LoisRForgeFlow @JordiBForgeFlow


LoisRForgeFlow:
@JoanSForgeFlow Can you check travis?

product_multi_company/models/product_category.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def get_user_companies(self):
+        return self.env.user.company_ids.ids
+
+    total_route_ids = fields.Many2many(
+        domain=[
+            "|",
+            ("company_id", "=", False),
+            ("company_id", "in", "get_user_companies"),
+        ]
```
LoisRForgeFlow: I think the proper way would be to create a new field computed field `allowed_route_ids` and then update this domain to be: `domain="[('id', 'in', allowed_route_ids)]

AaronHForgeFlow: Tried that but got recursion error as long as allowed_route_ids needed route_ids in order to calculate those :( Now I think it is working, but still checking

LoisRForgeFlow:
@AaronHForgeFlow Great! looks much better now! Could you attend travis and squash your commits?

AaronHForgeFlow:
Basically, the default method adds a company to https://github.com/OCA/multi-company/blob/13.0/account_invoice_inter_company/tests/inter_company_invoice.xml#L383 so, putting that `ref=""` is useless when product_multi_company is installed.

Besides that, in case company_ids is not set. The company_id should be False.

I am not sure if creating separate PRs for that...

AaronHForgeFlow:
@JoanSForgeFlow could you remove the "WIP" from the subject? :)

AaronHForgeFlow:
Finally, the extra check in base_multi_company module was unnecessary, and I removed it. In case company_ids is not set in the model it goes to the "else" in the compute method and no company_id is set.

LoisRForgeFlow (approved):
Great, Thanks for the work!

AaronHForgeFlow (approved):
Functional test + code review :innocent:

PR #200, opened 2020-03-31
[13.0][MIG] partner_multi_company

JoanSForgeFlow:
**partner_multi_company** migration to v13.
Depends on:
- #197

cc ~ @LoisRForgeFlow @JordiBForgeFlow


JordiBForgeFlow (approved):

JordiBForgeFlow:
@JoanSForgeFlow please add the oca_dependencies.txt file and include the reference to the base_multi_company, like it's done here 068c50972ca74dba0eeb148e68a4e40f5baefdc2

This will make it easier for reviewers to test using runbot. Now everything fails due to the missing dependency.

AaronHForgeFlow:
Do not merge yet. Last commit will break tests for account_invoice_multi_company when partner_multi_company is not installed. The only solution I see is separate in the CI the installation of this module from account_invoice_inter_company and, sale_purchase_inter_company

pedrobaeza:
If there's an incompatibility between 2 modules, better to use `config['test_enable']` and a context for detecting the testing of the module for enabling/disabling the incompatible behavior. Example:

https://github.com/OCA/bank-payment/blob/3c96fedbb95060c9cd3a2aa87d383324d486bee1/account_payment_transfer_reconcile_batch/models/payment_order.py#L28

AaronHForgeFlow:
I could not do that, because the error will raise in a constraint in standard Odoo when loading a file specified in the test. I tried to overwrite the records created in that file in the partner_multi_company module. But that was too late, because the original file will always be leaded as it is. The only way I found if to include a separate file and load one file or the other depending if the field company_ids exists or not in the partner.

pedrobaeza:
Well, unless you show me through Travis log, I can't tell you more.

pedrobaeza:
What I can say is that the XML import done in `account_invoice_inter_company` is very ugly and should be changed by conventional `setUp`.

AaronHForgeFlow:
Basically, the file creates the partner without a company as company partner. But this https://github.com/OCA/multi-company/blob/13.0/base_multi_company/models/multi_company_abstract.py#L24 assign a a value in company_ids, and consequently to the company_id field. Then an error will raise because the company of the partner does not match the company of the warehouse of the company.

I agree that loading that file that way is ugly, but it was done like that for a reason: https://github.com/OCA/multi-company/commit/e4ffc42837a7d82e8d8ab50bf883f82dc89e0ef1 and that reason seems still valid.

LoisRForgeFlow:
Another option is to add a travis task excluding that module.

pedrobaeza:
I'm not talking about returning to demo data, but create them in the `setUp`/`setUpClass` method. Anyway, I don't see this red. Or do you need to rebase? Please show me a red Travis with the error for being able to tell you how to fix it without isolating modules, which is a pain.

AaronHForgeFlow:
It is not red because I am loading different files depending if the module partner_multi_company is installed or not. Modules are not isolated and work together perfectly, the problem is the data loaded. I agree better solution is to create those records in the SetUpClass.

pedrobaeza:
Are you doing it or do you want me to do it?

AaronHForgeFlow:
And that is not the only module that loads data this way. In the end all those modules need a change in the way they load data for tests.

AaronHForgeFlow:
> Are you doing it or do you want me to do it?

I am doing that.

pedrobaeza:
OK, thanks. Contact me by chat if you want that both see anything in live.

account_invoice_inter_company/tests/test_inter_company_invoice.py
```diff
@@ -14,44 +14,591 @@ class TestAccountInvoiceInterCompanyBase(SavepointCase):
...
+                "name": "Creditors - (company A)",
+                "internal_type": "payable",
+                "reconcile": "True",
+                "user_type_id": cls.env.ref(
+                    "account.data_account_type_payable"
+                ).id,
+                "company_id": cls.company_a.id,
+            }
+        )
+        cls.invoice_company_a = cls.env["account.move"].create(
```
AaronHForgeFlow, line 351: Seems this line is failing:

`
 Remaining fields to compute before setting up registry: defaultdict(<class 'set'>, {account.move.amount_tax_signed: {<NewId 0x7f33b94a96d8>}, account.move.amount_total_signed: {<NewId 0x7f33b94a96d8>}, account.move.amount_residual: {<NewId 0x7f33b94a96d8>}, account.move.invoice_payment_state: {<NewId 0x7f33b94a96d8>}, account.move.amount_total: {<NewId 0x7f33b94a96d8>}, account.move.amount_residual_signed: {<NewId 0x7f33b94a96d8>}, account.move.amount_untaxed_signed: {<NewId 0x7f33b94a96d8>}, account.move.amount_untaxed: {<NewId 0x7f33b94a96d8>}, account.move.amount_tax: {<NewId 0x7f33b94a96d8>}, account.move.line.account_root_id: {<NewId 0x7f33baf29978>}, account.move.line.account_internal_type: {<NewId 0x7f33baf29978>}, account.move.line.amount_residual_currency: {<NewId 0x7f33baf29978>}, account.move.line.reconciled: {<NewId 0x7f33baf29978>}, account.move.line.amount_residual: {<NewId 0x7f33baf29978>}, account.move.line.tax_audit: {<NewId 0x7f33baf29978>}})
`

AaronHForgeFlow:
Do not merge yet. Still need to convert tests for purchase_sale_inter_company.

AaronHForgeFlow (approved):
Test refactored completed. This is now ready for review :)

LoisRForgeFlow:
Just fixed conflicts.

LoisRForgeFlow:
This module needs https://github.com/odoo/odoo/pull/59237 in order to not raise an error when trying to reconcile an statement.

juliap98 (approved):
functional review, LGTM

pedrobaeza:
Rebased on #266 for seeing if Travis is OK.

PR #209, opened 2020-04-23
[13.0][FIX] account_invoice_inter_company: Solved problems for created invoices

CarlosRoca13:
cc @Tecnativa TT22727


pedrobaeza (approved):
My only doubt is about writing in `name` field instead of `move_name`, but I remember that both are synchronized on validation.

CarlosRoca13:
@pedrobaeza Can you pass me the code where this synchronization is done?

ps-tubtim (approved):
Code LGTM.

PR #212, opened 2020-05-05
[MIG][13.0] product_tax_multicompany_default

acsonefho:
Migration of module `product_tax_multicompany_default`.

- [x] Pre-commit
- [x] Tested

### **Remarks:**
I wasn't able to make it works in v13.0 without changes.

**Previous behavior:**
In v10.0 and v11.0, Odoo add/update `ir.values`/`ir.default` with taxes set into Accounting configuration.
So this module use these values to set taxes from others company.
But since v12.0, fields into Accounting configuration are related fields (on the company).
So there is no more filled `ir.values`/`ir.default`.
So maybe it still working in v12.0 due to migration (who doesn't clear existing default values data). But if you update the configuration, I think it won't work properly (I didn't test).

**New behavior:**
Use taxes field on the company to have default taxes.
I did this change into a separated commit.


acsonefho:
@pedrobaeza @carlosdauden Can you look at this please? :)

PR #213, opened 2020-05-11
[13.0] [MIG] account_multicompany_easy_creation: Migration to v13.0

CarlosRoca13:
cc @Tecnativa TT22723


account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -0,0 +1,373 @@
...
+        self.create_company()
+        self.update_taxes()
+        self.update_properties()
+        return self.action_res_company_form()
+
+
+class AccountMulticompanyBankWiz(models.TransientModel):
+    _inherit = "res.partner.bank"
+    _name = "account.multicompany.bank.wiz"
+    _order = "id"
```
chienandalu, line 378: Add a `_description` to the model

CarlosRoca13: Done!

ernestotejeda (commented):
Code review :+1:

ernestotejeda:
I got this error during functional review in runbot. check it out.

Peek 2020-07-29 16-53

chienandalu (changes requested):
@Yajo Can you remeber the problem @ernestotejeda is having? Wasn't it something about demo data or defaults?

yajo:
It seems due to this field that is being exposed in the view:

imagen

This makes that, after getting default values, the wizard appears with a `NewId` `ir.sequence` record, which then [tries to compute its next value and fails because it needs an actual ID to do so](https://github.com/odoo/odoo/blob/5225b9ce5178302af05b63029fb184e27b781815/odoo/addons/base/models/ir_sequence.py#L95).

Possible fixes:

1. Patch the compute method upstream to support the `NewId` case.
1. Do not display that field in the wizard view.
1. Add `context={"default_number_next_actual": 1}` to the child view, so Odoo uses 1 by default and doesn't attempt to recompute it.

CarlosRoca13:
Finally I solved the problem removing the field `number_next_actual` from the wizard view. With the other options that @Yajo said I couldn't achieve solve the problem. Pease test it again :smile:

account_multicompany_easy_creation/i18n/es.po
```diff
@@ -0,0 +1,435 @@
...
+msgid "account.multicompany.bank.wiz"
+msgstr "'account.multicompany.bank.wiz'"
+
+#. module: account_multicompany_easy_creation
+#: model_terms:ir.ui.view,arch_db:account_multicompany_easy_creation.wizard_multicompany_easy_creation_view_form
+msgid "or"
+msgstr "o"
+
+#~ msgid "Number of digits in an account code"
+#~ msgstr "Número de dígitos en un código de cuenta"
```
victoralmau:
```suggestion
```

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -0,0 +1,374 @@
+# Copyright 2018 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
```
victoralmau:
```suggestion
# Copyright 2018 Tecnativa - Carlos Dauden
```

account_multicompany_easy_creation/wizards/multicompany_easy_creation.xml
```diff
@@ -0,0 +1,135 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- Copyright 2018 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
```
victoralmau:
```suggestion
<!-- Copyright 2018 Tecnativa - Carlos Dauden
```

victoralmau (approved):
Good Job

CarlosRoca13:
@victoralmau Changes done

pedrobaeza:
@ernestotejeda can you re-review?

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -0,0 +1,374 @@
...
+class AccountMulticompanyEasyCreationWiz(models.TransientModel):
+    _name = "account.multicompany.easy.creation.wiz"
+    _description = "Wizard Account Multi-company Easy Creation"
+
+    def _default_sequence_ids(self):
+        exclude_seq_list = self.env["ir.config_parameter"].get_param(
+            "account_multicompany_easy_creation.exclude_sequence_list",
+            [
+                False,
+                "aeat.sequence.type",
```
yajo, line 20: Is this appropriate for a module outside of l10n-spain?

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -0,0 +1,374 @@
...
+    def create_company(self):
+        self.new_company_id = self.env["res.company"].create(
+            {"name": self.name, "user_ids": [(6, 0, self.user_ids.ids)]}
+        )
+        self.install_chart_account()
+        self.create_bank_journals()
+        self.create_sequences()
+
+    # TODO: Cache don't work
+    @ormcache("self.id", "company_id", "match_tax_ids")
```
yajo: In such case, shouldn't you remove the cache? Or the comment? :laughing:

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
pedrobaeza: As both are OCA modules, this is allowed as "trick" for avoiding glue modules.

yajo: OK, but I think there should be some comment here explaining that.

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -0,0 +1,374 @@
...
+
+    def create_sequences(self):
+        for sequence in self.sudo().sequence_ids:
+            sequence.copy({"company_id": self.new_company_id.id})
+
+    def create_company(self):
+        self.new_company_id = self.env["res.company"].create(
+            {"name": self.name, "user_ids": [(6, 0, self.user_ids.ids)]}
+        )
+        self.install_chart_account()
```
victoralmau:
```suggestion
        self.new_company_id = self.env["res.company"].create(
            {"name": self.name, "user_ids": [(6, 0, self.user_ids.ids)]}
        )
        self.new_company_id.chart_template_id = self.chart_template_id
        self.new_company_id.chart_template_id.try_loading_for_current_company()
```
It is necessary to apply this change or it does NOT apply the accounting plan correctly (if l10n_es is tested, this can be confirmed).

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -0,0 +1,374 @@
...
+        comodel_name="account.account.template",
+        string="Default Category Expense Account",
+    )
+
+    def install_chart_account(self):
+        """ install a chart of accounts for the given company """
+        user_company = self.env.user.company_id
+        self.env.user.company_id = self.new_company_id
+        self.sudo().chart_template_id.try_loading_for_current_company()
+        self.env.user.company_id = user_company
```
victoralmau:
```suggestion
```

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -0,0 +1,374 @@
...
+
+    def create_bank_journals(self):
+        AccountJournal = self.env["account.journal"].sudo()
```
victoralmau, line 118:
```suggestion
        AccountJournal = self.env["account.journal"].sudo()
        AccountAccount = self.env["account.account"].sudo()
```

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -0,0 +1,374 @@
...
+                bank_journals[i].update(vals)
+            else:
+                vals.update(
+                    {
+                        "code": False,
+                        "sequence_id": False,
+                        "default_debit_account_id": False,
+                        "default_credit_account_id": False,
+                    }
+                )
```
victoralmau:
```suggestion
                account_account = AccountAccount.create({
                    "code": "57200X",
                    "name": vals["name"],
                    "user_type_id": cls.env.ref(
                        "account.data_account_type_liquidity"
                    ).id,
                    "company_id": vals["company_id"]
                })
                vals.update(
                    {
                        "code": False,
                        "sequence_id": False,
                        "default_debit_account_id": account_account.id,
                        "default_credit_account_id": account_account.id,
                    }
                )
```
It is necessary to create an `account.account` and define it to the journal.

victoralmau (changes requested):
After testing it, there are some bugs that I think are important to solve.

CarlosRoca13:
I applied the changes suggested @Yajo @victoralmau

victoralmau (approved):

yajo (approved):

chienandalu:
Looks like this PR didn't get correctly merged. @pedrobaeza Could you relaunch?

PR #218, opened 2020-05-21
[12.0][IMP] account_invoice_inter_company: onchange_helper dep

chienandalu:
Removes `onchange_helper` dependency and uses fancy `Form` class for the equivalent one (and even more complete and bullet-proof).

cc @Tecnativa TT21089

### DIFF
--- a/account_invoice_inter_company/models/account_invoice.py
+++ b/account_invoice_inter_company/models/account_invoice.py
@@ -3,6 +3,7 @@
 from odoo import _, api, fields, models
 from odoo.exceptions import AccessError, UserError
 from odoo.tools import float_compare
+from odoo.tests.common import Form
 
 
 class AccountInvoice(models.Model):
@@ -161,36 +162,29 @@ class AccountInvoice(models.Model):
             raise UserError(_(
                 'Please define %s journal for this company: "%s" (id:%d).')
                 % (dest_journal_type, dest_company.name, dest_company.id))
-        # find account, payment term, fiscal position, bank.
-        dest_partner_data = {
-            'type': dest_inv_type,
-            'partner_id': self.company_id.partner_id.id,
-            'company_id': dest_company.id,
-        }
-        dest_partner_data = self.env['account.invoice'].play_onchanges(
-            dest_partner_data, ['partner_id'])
-        return {
-            'name': self.name,
-            'origin': _('%s - Invoice: %s') % (self.company_id.name,
-                                               self.number),
-            'type': dest_inv_type,
-            'date_invoice': self.date_invoice,
-            'reference': self.number,
-            'account_id': dest_partner_data.get('account_id', False),
-            'partner_id': self.company_id.partner_id.id,
-            'journal_id': dest_journal.id,
-            'currency_id': self.currency_id.id,
-            'fiscal_position_id': dest_partner_data.get(
-                'fiscal_position_id', False),
-            'payment_term_id': dest_partner_data.get(
-                'payment_term_id', False),
-            'company_id': dest_company.id,
-            'partner_bank_id': dest_partner_data.get(
-                'partner_bank_id', False),
-            'auto_generated': True,
+        # Use test.Form() class to trigger propper onchanges on the invoice
+        dest_invoice_data = Form(
+            self.env['account.invoice'].with_context(
+                default_type=dest_inv_type,
+                force_company=dest_company.id
+            ))
+        dest_invoice_data.company_id = dest_company
+        dest_invoice_data.partner_id = self.company_id.partner_id
+        dest_invoice_data.name = self.name
+        dest_invoice_data.date_invoice = self.date_invoice
+        dest_invoice_data.reference = self.number
+        dest_invoice_data.comment = self.comment
+        dest_invoice_data.journal_id = dest_journal
+        dest_invoice_data.currency_id = self.currency_id
+        vals = dest_invoice_data._values_to_save(all_fields=True)
+        vals.update({
+            'origin': _('%s - Invoice: %s') % (
+                self.company_id.name, self.number),
             'auto_invoice_id': self.id,
             'comment': self.comment,
-        }
+            'auto_generated': True,
+        })
+        return vals
 
     @api.multi
     def action_invoice_cancel(self):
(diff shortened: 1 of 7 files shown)


account_invoice_inter_company/models/account_invoice.py
```diff
@@ -161,32 +162,26 @@ def _prepare_invoice_data(self, dest_company):
...
-            'company_id': dest_company.id,
-        }
-        dest_partner_data = self.env['account.invoice'].play_onchanges(
-            dest_partner_data, ['partner_id'])
+        # Use test.Form() class to trigger propper onchanges on the invoice
+        dest_invoice_data = Form(self.env['account.invoice'].with_context(
+            default_type=dest_inv_type))
+        dest_invoice_data.company_id = dest_company
+        dest_invoice_data.partner_id = self.company_id.partner_id
         return {
```
pedrobaeza: Do:
```python
vals = dest_invoice_data._values_to_save(all_fields=True)
vals.update({
    'name': self.name,
    'origin': ...
```
The idea of this is to recover all the possible extra values that other modules can inject in the invoice. You can remove also a lot of hardcoded values (or event better, assign them to the form instance instead of later, for also have the corresponding onchanges played).

Do the same on the invoice lines.

pedrobaeza (changes requested):

account_invoice_inter_company/models/account_invoice.py
chienandalu: Done in invoice, but in invoice lines I'm having this issue following that method:

current `vals`:
```python
{ 
    'account_analytic_id':	False,
    'account_id':	False,
    'analytic_tag_ids':	[(
        6, 
        False, 
        []
        )],
    'company_id':	False,
    'currency_id':	False,
    'discount':	0,
    'display_type':	False,
    'invoice_line_tax_ids':	[(
        6, 
        False, 
        [2]
        )],
    'invoice_type':	False,
    'name':	False,
    'partner_id':	False,
    'price_unit':	1.0,
    'product_id':	66,
    'quantity':	1.0,
    'sequence':	10,
    'uom_id':	6,
}
```

`vals` generated with `vals = invoice_line_dest._values_to_save(all_fields=True)`:
```python
{
    'account_analytic_id':	account.analytic.account(),
    'account_id':	237,
    'auto_invoice_line_id':	54,
    'company_id':	44,
    'discount':	0.0,
    'invoice_id':	63,
    'invoice_line_tax_ids':	[2],
    'name':	'Service Multi Company',
    'partner_id':	167,
    'price_unit':	450.0,
    'product_id':	66,
    'quantity':	1.0,
    'sequence':	10,
}
``

File: account_invoice_inter_company/models/account_invoice.py [hunk 2]
```diff
@@ -161,36 +162,28 @@ def _prepare_invoice_data(self, dest_company):
...
+        vals.update({
             'name': self.name,
             'origin': _('%s - Invoice: %s') % (self.company_id.name,
                                                self.number),
-            'type': dest_inv_type,
             'date_invoice': self.date_invoice,
             'reference': self.number,
-            'account_id': dest_partner_data.get('account_id', False),
-            'partner_id': self.company_id.partner_id.id,
             'journal_id': dest_journal.id,
```
pedrobaeza: Do these assignations on the form, not later, as they can trigger things.

pedrobaeza (approved):

account_invoice_inter_company/views/account_invoice_views.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="invoice_supplier_form" model="ir.ui.view">
+        <field name="name">account.invoice.supplier.form - HACK: Extra fields for odoo.tests.Form() to work</field>
+        <field name="model">account.invoice</field>
+        <field name="inherit_id" ref="account.invoice_supplier_form"/>
+        <field name="mode">primary</field>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='invoice_line_ids']/tree" position="inside">
+                <field name="company_id" invisible="1" readonly="0"/>
```
chienandalu: Already in the main view. Could it be causing the error in travis?
https://github.com/odoo/odoo/blob/12.0/addons/account/views/account_invoice_view.xml#L492

account_invoice_inter_company/views/account_invoice_views.xml
```diff
@@ -0,0 +1,18 @@
...
+                <field name="partner_id" invisible="1" readonly="0"/>
+                <field name="invoice_type" invisible="1" readonly="0"/>
+                <field name="currency_id" invisible="1" readonly="0"/>
```
chienandalu: In main view as well:
https://github.com/odoo/odoo/blob/12.0/addons/account/views/account_invoice_view.xml#L511

account_invoice_inter_company/views/account_invoice_views.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="invoice_supplier_form" model="ir.ui.view">
+        <field name="name">account.invoice.supplier.form - HACK: Extra fields for odoo.tests.Form() to work</field>
+        <field name="model">account.invoice</field>
+        <field name="inherit_id" ref="account.invoice_supplier_form"/>
+        <field name="mode">primary</field>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='invoice_line_ids']/tree/company_id" position="attributes">
```
chienandalu: I guess, isn't it?
```suggestion
            <xpath expr="//field[@name='invoice_line_ids']/tree/field[@name='company_id']" position="attributes">
```

account_invoice_inter_company/views/account_invoice_views.xml
```diff
@@ -0,0 +1,22 @@
...
+                <attribute name="readonly">0</attribute>
+            </xpath>
+            <xpath expr="//field[@name='invoice_line_ids']/tree/currency_id" position="attributes">
```
chienandalu: And the same here
```suggestion
            <xpath expr="//field[@name='invoice_line_ids']/tree/field[@name='currency_id']" position="attributes">
```

sergio-teruel (approved):

pedrobaeza:
Finally green, so let's merge and forward-port it:

/ocabot merge major

PR #221, opened 2020-06-03
[13.0][IMP] account_invoice_inter_company: Remove onchange_helper dependency

pedrobaeza:
Removes `onchange_helper` dependency and uses fancy `Form` class for the equivalent one (and even more complete and bullet-proof).

Forward-port to v13 of #218

cc @Tecnativa TT21089

### DIFF
--- a/account_invoice_inter_company/views/account_move_views.xml
+++ b/account_invoice_inter_company/views/account_move_views.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_move_form" model="ir.ui.view">
+        <field
+            name="name"
+        >account.move.form - HACK: Extra fields for odoo.tests.Form() to work</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="mode">primary</field>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='invoice_line_ids']/tree" position="inside">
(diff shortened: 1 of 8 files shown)


chienandalu (approved):
Tested in runbot :+1:

account_invoice_inter_company/views/account_move_views.xml
```diff
@@ -0,0 +1,16 @@
...
+    <record id="view_move_form" model="ir.ui.view">
+        <field
+            name="name"
+        >account.move.form - HACK: Extra fields for odoo.tests.Form() to work</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="mode">primary</field>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='invoice_line_ids']/tree" position="inside">
+                <field name="company_id" invisible="0" readonly="0" />
```
sergio-teruel, line 12: Always visible???

sergio-teruel (approved):
Only a little comment

account_invoice_inter_company/views/account_move_views.xml
chienandalu: It's a primary for this specific purpose, so it shouldn't be a problem, isn't it?

pedrobaeza: Indeed, although I think I wanted to put invisible in that moment and it was a typo. It doesn't matter anyway.

PR #223, opened 2020-06-05
[FIX] account_move_multi_company: Check ID in IF statement

osi-scampbell:
This if statement is incorrect. "dedicated_companies_vals" is a dictionary where "company_id" is an integer. "line.transfer_to_company_id" returns res.company(X) so the if statement always evaluates to False. Adding ".id" fixes this issue and the if statement works as intended

### DIFF
--- a/account_move_multi_company/models/account_move.py
+++ b/account_move_multi_company/models/account_move.py
@@ -53,7 +53,7 @@ class AccountMove(models.Model):
                         'debit': line.debit,
                         'credit': line.credit}))
 
-                    if line.transfer_to_company_id not in\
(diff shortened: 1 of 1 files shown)


max3903 (approved):

bodedra (approved):
Looks good.

JFI, changes can be replace with variable "company_id" as we set value "line.transfer_to_company_id.id"

patrickrwilson (approved):
LGTM

pedrobaeza:
ocabot stuck. Please relaunch

max3903:
@pedrobaeza Thank for notifying...

PR #225, opened 2020-06-25
[IMP] purchase_sale_inter_company: disallow auto validate pickings

ps-tubtim:
This PR fixes can't validate delivery order if product tracked by lot/serial number.

Selection_055

I edit this error by when validating delivery order if product tracked by lot/serial, don't auto validate receipt.

**Before**
Peek 2020-06-25 11-54

**After**
Peek 2020-06-25 12-14


ps-tubtim:
CC: @kittiu @Saran440 @newtratip @kranokporn @watcharaporn-git

purchase_sale_inter_company/models/stock_picking.py
```diff
@@ -18,7 +18,8 @@ def action_done(self):
             lambda x: x.location_dest_id.usage == "customer"
         ).sudo():
             purchase = pick.sale_id.auto_purchase_order_id
-            if not purchase:
+            # Not create from inter-company or has products tracked by lots
+            if not purchase or pick.has_tracking:
```
kittiu:
```suggestion
            if not purchase:
              continue
            if pick.has_tracking:
              continue
```

kittiu: Just IMHO, I would think it is better to separated case, and keep existing as is.
As the 2 case are quite different.

kittiu: For me, I think no need for the comment, as the `pick.has_tracking` is quite clear that it has lot/serial.

kittiu:
image

The readme is not in good shape, please take this opportunity to fix it. I.e., number list.

kittiu:
Note to contributors: this addon seem not supporting log/serial from the beginning. By this fix, it just ensure that when OUT from the seller has lot/serial, there will be no auto IN transfer. Not sure this is good enough?

@ps-tubtim or is it possible to still have auto IN transfer, while not worry about IN lot/serial counterpart.

Saran440:
@ps-tubtim Could you add test script?

kittiu:
@ps-tubtim after more thought I think this will make system inconsistent between case No LOT and case with LOT.
I think for case LOT, system should auto assign LOT (split to multi LOT if necessary) as configurable option.
Note: and from what I understand from EE, it never auto validate IN too.

I propose these to be configurable as part of multi company section.

- [ ] Auto validate buyer's incoming shipment when seller's outgoing shipment is validated.
- [ ] Propagate Lot / Serial number from seller's outgoing shipment to buyer's incoming shipment.

WDYT?

pedrobaeza:
For me all the stock things should go out of this module, and be included in a new one `purchase_sale_stock_inter_company`, handling properly all the stack:

- Constraints over lots with company associated.
- Validation of the counter-part picking when the picking is validated.
- Lot and quantities propagation.
- Backorders.
- ...

ps-tubtim:
I deleted code about pickings auto validation. I will create purchase_sale_stock_inter_company module for handling property all the stock.

ps-tubtim:
@Saran440 @kittiu Could you review this?

ps-tubtim:
Closed PR because this PR is not necessary.

PR #230, opened 2020-07-05
[10.0][FIX] account_invoice_inter_company: fix use of play_onchanges

alexis-via:
Without this fix, the invoice created in the other company didn't have any line (invoice_id was False in the call to create on
account.invoice.line)


MOSAMAD:
OK, fine !

MOSAMAD:
:+1:

PR #233, opened 2020-07-13
[FIX] acccount_payment_other_company: Search Due From not Due To

osi-scampbell:
Because we are searching for Account Move Lines in the current company, we expect to see the Due From account not the Due To

### DIFF
--- a/account_payment_other_company/__manifest__.py
+++ b/account_payment_other_company/__manifest__.py
@@ -15,6 +15,7 @@
     ],
     'data': [
         'views/account_payment.xml',
+        'views/account_register_payments.xml'
     ],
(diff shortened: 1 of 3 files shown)


account_payment_other_company/__manifest__.py
```diff
@@ -15,6 +15,7 @@
     ],
     'data': [
         'views/account_payment.xml',
+        'views/account_register_payments.xml'
```
bodedra:
```suggestion
        'wizard/account_register_payments.xml'
```

bodedra (changes requested):

bodedra (approved):

PR #234, opened 2020-07-17
[13.0][FIX] base_multi_company: consider force_company context key.

LoisRForgeFlow:
Yet another fix more for OCA multi company :laughing: - Keep in mind that multi-company functionality has changed a lot in v13 and it is hard to find all corner cases.

This is needed for transactions in which the company is forced, e.g
intercompany PO/SO synchronization.

@ForgeFlow


pedrobaeza (approved):

AaronHForgeFlow (approved):
:+1: Makes sense to me.

PR #238, opened 2020-08-10
12 IMP acc_inv_inter_cpny: detect intercompany with commercial_partner_id

bealdav:
## before
If you use any contact of the multicompany partner in your invoice, workflow is not triggered, whereas you may create an additionnal contact to specify some particular data like (specific email or any)

## after
All contact of the customer company used in invoice trigger intercompany workflow. It's more intuitive.

### DIFF
--- a/account_invoice_inter_company/models/account_invoice.py
+++ b/account_invoice_inter_company/models/account_invoice.py
@@ -26,7 +26,7 @@ class AccountInvoice(models.Model):
     def _find_company_from_invoice_partner(self):
         self.ensure_one()
         company = self.env['res.company'].sudo().search([
-            ('partner_id', '=', self.partner_id.id)
+            ('partner_id', '=', self.commercial_partner_id.id)
         ], limit=1)
         return company or False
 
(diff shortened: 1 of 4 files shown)


account_invoice_inter_company/models/account_invoice.py
```diff
@@ -26,7 +26,7 @@ class AccountInvoice(models.Model):
     def _find_company_from_invoice_partner(self):
         self.ensure_one()
         company = self.env['res.company'].sudo().search([
-            ('partner_id', '=', self.partner_id.id)
+            ('partner_id', '=', self.commercial_partner_id.id)
```
bodedra, line 29:
```suggestion
            ('partner_id', 'child_of', self.commercial_partner_id.id)
```

bodedra (commented):
Added minor comment.

bealdav:
Hi @bodedra, thanks for your review.
I'm not sure `child of` brings more added value or security for this case
commercial_partner_id always refers to parent, then no need of children.
Maybe you have a case that I missed ?

bodedra:
@bealdav Yes. I agree. That's why I delete my comment. Let's wait for green. This PR is good to me. :+1:

bealdav:
cc @chienandalu @pedrobaeza @astirpe @chafique-delli

bodedra (approved):
LGTM :+1:

astirpe (approved):

pedrobaeza (approved):

PR #241, opened 2020-09-03
[12.0] Fix to support Vendor Bill Batch Payment. Otherwise there would be a singleton error.

agyamuta:


osi-scampbell:
@agyamuta @max3903 Moved to #233. Can we closet this?

PR #255, opened 2020-11-25
[FIX] purchase_sale_inter_company: commitment_date

chienandalu:
The field was renamed from `requested_date` to `commitment_date`

cc @Tecnativa TT26981

### DIFF
--- a/purchase_sale_inter_company/models/purchase_order.py
+++ b/purchase_sale_inter_company/models/purchase_order.py
@@ -127,8 +127,7 @@ class PurchaseOrder(models.Model):
             new_order.warehouse_id = (
                 dest_company.warehouse_id.company_id == dest_company and
                 dest_company.warehouse_id or False)
-        if 'requested_date' in new_order:
-            new_order.requested_date = self.date_planned
+        new_order.commitment_date = self.date_planned
         return new_order._convert_to_write(new_order._cache)
 
     @api.model


purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -115,6 +115,7 @@ def _prepare_sale_order_data(self, name, partner, dest_company,
             'partner_id': partner.id,
             'date_order': self.date_order,
             'auto_purchase_order_id': self.id,
+            'commitment_date': self.date_planned,
```
pedrobaeza: Isn't this change in later onchanges?

chienandalu: As long as I tested it doesn't change

pedrobaeza: Well, I prefer to assign it at the end as the previous one for being sure it doesn't change

chienandalu: Ok, changed

pedrobaeza (approved):

pedrobaeza:
Let's fast-track this for fixing the bug:

/ocabot merge patch

Is this needed for v13?

chienandalu:
Yes, it seems to have the same issue

bealdav (approved):

PR #260, opened 2020-12-11
[12.0][ADD] Down payment for multi company

quentinDupont:
This module extends the functionality of sale module to handle one product per company for down payment


quentinDupont:
ping @legalsylvain as we use it in production since few months now :)

legalsylvain (approved):
thanks.

PR #262, opened 2020-12-15
Upgrade pylint for v13

joao-p-marques:
Similar to https://github.com/OCA/product-attribute/commit/846d08cd0eeb24b75a045d9ef1155943e818a213 to fix pre-commit issues with new pip resolver.

Does not introduce any linting change and keeps dependencies up to date.

ping @pedrobaeza @sbidoul


sbidoul:
@joao-p-marques To update a 13.0 branch, it's better to use `copier` to initialize from the OCA/oca-addons-repo-template, so we can update it more easily in the future. Could you try this?

.prettierrc.yml
```diff
@@ -5,4 +5,4 @@ printWidth: 88
 proseWrap: always
 semi: true
 trailingComma: "es5"
-xmlWhitespaceSensitivity: "ignore"
+xmlWhitespaceSensitivity: "strict"
```
pedrobaeza: This shouldn't be `ignore`?

joao-p-marques:
Thanks @sbidoul 
I just did that. I tried to keep everything as it is right now with the copier answers but I might have missed something. Can you check please?

BTW, is this the intended flow for all v13 repos? I thought this was only available for v14, but it seems much better this way :+1:

.prettierrc.yml
joao-p-marques: This was updated by copier from https://github.com/OCA/oca-addons-repo-template/blob/master/.prettierrc.yml#L8

Should I change it back?

pedrobaeza: This has proven to be problematic AFAIK. cc @Yajo

joao-p-marques: Ok, I will change it back, then. 
I was worried it was going to cause trouble in any `copier update` as the upstream is `strict`, but it doesn't, so no problem.

pedrobaeza (approved):

.copier-answers.yml
```diff
@@ -0,0 +1,17 @@
+# Do NOT update manually; changes here will be overwritten by Copier
+_commit: v1.1.1
+_src_path: https://github.com/OCA/oca-addons-repo-template.git
+dependency_installation_mode: OCA
+generate_requirements_txt: true
+include_wkhtmltopdf: true
```
sbidoul: Do we need wkhtmltopdf for tests in this repo?

sbidoul:
@joao-p-marques 

> I just did that. I tried to keep everything as it is right now with the copier answers but I might have missed something. Can you check please?

Thanks! Looks good to me, except the addition of wkhtmltopdf which was not there before and we may not need here. In which case it's better to not install it, for better test job performance.

> BTW, is this the intended flow for all v13 repos? I thought this was only available for v14, but it seems much better this way +1

Yes, the template is available for 13 and 14. Currently I push the updates to 13 and 14 only if a `.copier-answers.yml` is present.

.copier-answers.yml
```diff
@@ -0,0 +1,17 @@
...
+odoo_version: 13.0
+rebel_module_groups: []
+repo_description:
+  Addons for the management of multi company instances as well as inter company
+  communication.
+repo_name: Multi Company
+repo_slug: multi-company
+travis_apt_packages:
+  - expect-dev
+  - python-lxml
```
sbidoul: Ah and we probably don't need these two. At least `lxml` is not needed because it was an optimization, but nowadays there is a pre-built wheel on PyPI and if not the pip cache will be sufficient for performance.

joao-p-marques: Ok, I kept them because they were in the old travis.yaml. Should I leave `expected-dev` only?

sbidoul: I just checked and [expect-dev is installed always](https://github.com/OCA/oca-addons-repo-template/blob/416ab75ccf656fbdde261fa9846903266bf90b3a/.travis.yml.jinja#L22), so you can leave it out.

joao-p-marques:
Thanks @sbidoul 
Changes made.

.prettierrc.yml
yajo: The problematic one was "ignore", so this update was OK. I'll fix it.

PR #263, opened 2020-12-16
[FIX] Restore prettier XML configuration

yajo:
This was correctly updated in https://github.com/OCA/multi-company/pull/262 but reverted by mistake.

See https://github.com/OCA/multi-company/pull/262#discussion_r544084207.


pedrobaeza:
@Yajo the conflicting configuration then is to put it as `ignore`?

pedrobaeza:
I have seen the other reply. Sorry for the confusion.

Issue #252, opened 2020-11-09
[13.0] product_multi_company - access rule error.


amkarthik:
**Steps to Reproduce**
* Create 3 companies - Company A, Company B, Company C
* Create a Product A and assign it Company A and Company B
* Create an user A and his allowed companies be Company B and Company C
* Login with User A and Go to Sales -> Products -> Products Variants and open the Product A

It will raise the below error.
![image](https://user-images.githubusercontent.com/12694917/98553920-2421dc80-22c6-11eb-873d-2e17e747e914.png)

It is working fine with the *Product Template* (only shows Company B in Companies fields).

amkarthik:
cc ~ @JoanSForgeFlow

amkarthik:
https://github.com/OCA/multi-company/pull/264 Fixed it

PR #264, opened 2020-12-28
[13.0][FIX] The variants could not be read by users because the compa…

pankk:
…ny_ids related field was always empty. Made the field non-related to allow variants to be used in multicompany environments separately from their templates.

Proposing a fix to #252. This assumes that the company_ids fields of product.template and product.product do not have to be the same. This means that different companies can manage their own product variants. The product.template can also be managed by a separate company. On creation the company_ids fields for the other product record (product.product when creating the product.template and vice versa) will be an exact match.

If there is demand for it I can work on a configuration option to keep the company_ids fields for both models identical.


pedrobaeza:
The direct problem I see with this approach is that granting one company in the variant, you need as well permission in the template, or you will have an error, so at least that consistency should be kept.

pankk:
Thanks for the review @pedrobaeza 
I'll propose the following update later today:
- [x] Always ensure product_template.company_ids is a superset of product_product.company_ids

product_multi_company/migrations/13.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,24 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+from openerp import SUPERUSER_ID, api
```
pedrobaeza:
```suggestion
from odoo import SUPERUSER_ID, api
```

product_multi_company/__manifest__.py
```diff
@@ -7,10 +7,13 @@
     "author": "Tecnativa," "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/multi-company",
     "category": "Product Management",
-    "version": "13.0.1.0.0",
+    "version": "13.0.1.1.0",
```
pedrobaeza: As this is a great change involving migration scripts, you should bump major version

product_multi_company/migrations/13.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def migrate(cr, installed_version):
+    _logger.debug(
+        "Populate new multicompany field on product.product which we added "
+        "because it was not correctly inherited from product.template."
+    )
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    for product_product in env["product.product"].search([]):
+        if len(product_product.company_ids) == 0:
```
pedrobaeza: This is not going to work. You need to read template model.

product_multi_company/models/product_product.py
```diff
@@ -0,0 +1,47 @@
...
+
+class ProductProduct(models.Model):
+    _inherit = ["multi.company.abstract", "product.product"]
+    _name = "product.product"
+    _description = "Product Variant (Multi-Company)"
+
+    company_ids = fields.Many2many(
+        string="Companies",
+        comodel_name="res.company",
+        default=lambda self: self._default_company_ids(),
```
pedrobaeza, line 15: This default is going to act on module update, so you should add a pre-migration for adding manually the SQL column.

pedrobaeza (changes requested):

product_multi_company/migrations/13.0.1.1.0/post-migration.py
pankk: In the 13.0.1.0.0 version `product_product.company_ids` is a related field and would not always == 0.
However, the upgrade runs first. This makes `product_product.company_ids` a real field and adds the `product_product_res_company_rel` table. The default is not set even if the system user has companies (observed by testing). This may be a side effect of what you mention below (the table does not exist), but in this case I would expect an error, and I don't see that during testing (not even a warning).
After the `product_product_res_company_rel` table is added the post-migration script runs and since no writes were performed the table is still empty. It is subsequently populated by the write in the post-migration script.

I tested this by creating a new database in a clean docker environment with the 13.0.1.0.0 version of this module. I added companies, users, products and product variants and added companies on them. I then reproduced the original bug (#252).
I then checked out the 13.0.2.0.0 version and commented out all other ways `product_product.company_ids` could be written to in that module. I also added a set_trace and observed that `product_product.company_ids` was indeed written in the migration script.

That said (long breath), I'm eager to learn if I overlooked something. These _inherits and abstract models can set me on the wrong trail some times. :sweat_smile: 

I did find a mistake in product_template.py. I'll push a new commit shortly.

product_multi_company/models/product_product.py
pankk: As far as I've observed in the past, if a table does not yet exist and the field definition has a default or a compute method, then the table is created and populated accordingly on an upgrade. While testing this is also what happens, the `product_product_res_company_rel` table is created, but as self.env.company.ids is not relevant, it stays empty. This is what we want since we want the post-migration script to populate it.
I've tested it with products with and without company_ids set, and the results are as expected.

product_multi_company/migrations/13.0.1.1.0/post-migration.py
pankk: Isn't this normally handled via pre-commit? Or maybe it was a MQT script that checked it.

product_multi_company/__manifest__.py
pankk: I thought ocabot would handle this, but it doesn't rename the migration directory of course.

pankk:
I'll check the tests

pankk:
Warning from runbot:
```
2020-12-29 13:48:14,211 159 WARNING openerp_test odoo.models: method product.template._check_sale_product_company: @constrains parameter 'company_id' is not writeable
```

It looks like this is caused by `base_multi_company` (re)defining company_id on `product.template`. If so then this was in the previous version as well. If I run an `-i product_multi_company` on the 13.0 version I get the same warning.

account_invoice_inter_company/tests/test_inter_company_invoice.py
```diff
@@ -340,6 +340,7 @@ def setUpClass(cls):
         if "company_ids" in cls.env["product.template"]._fields:
             # We have to do that because the default method added a company
             cls.product_consultant_multi_company.company_ids = False
+            cls.product_consultant_multi_company.product_tmpl_id.company_ids = False
```
pedrobaeza, line 343: This is using a field that doesn't exist without the other module installed, so this highlights a problem in this PR that should be solved other way.

pankk: The same goes for the line above it. The existing comment already reads:
```
# if product_multi_company is installed
```
The dependency is checked with the existing conditional. I don't see the issue here, the test is just very thorough.

pedrobaeza: OK, I see... yes, it's not ideal, but let's continue for not blocking this.

pedrobaeza (approved):

account_invoice_inter_company/tests/test_inter_company_invoice.py
pankk: Any system that cannot by its very nature be orthogonal is not ideal, hence ERP's are not ideal. Sure beats spreadsheets though :stuck_out_tongue_winking_eye: 
As always, thanks for going into the fine details @pedrobaeza!

pedrobaeza: Thanks to you for solving this problem.

amkarthik:
@pankk 
It works well
Thanks for the fix.

amkarthik:
+1

atchuthan (approved):

atchuthan:
@pankk could you resolve the conflict with product_multi_company/__manifest__.py

atchuthan (approved):

PR #270, opened 2021-03-09
[14.0][MIG] base_multi_company

rousseldenis:
- [x] Use odoo-test-helper to test Abstract class
- [x] Added an inverse on company_id to avoid warnings in logs
- [x] Added a depends_context on company 
- [x] Cherry-picked #313 
- [x] Simplified inverse_company_id to update once


base_multi_company/models/multi_company_abstract.py
```diff
@@ -0,0 +1,78 @@
...
+        # e.g: ```
+        #     WHERE "res_partner"."id" not in (SELECT "res_partner_id"
+        #             FROM "res_company_res_partner_rel"
+        #             where "res_partner_id" is not null)
+        #         OR  ("res_partner"."id" in (SELECT "res_partner_id"
+        #             FROM "res_company_res_partner_rel" WHERE "res_company_id" IN 1)
+        # ```
+        new_args = []
+        for arg in args:
+            if type(arg) == list and arg[:2] == ["company_id", "in"]:
```
rousseldenis, line 104: @LoisRForgeFlow Could you give me the example of args and where it is generated in this case ?

I want to add further tests for this

base_multi_company/models/multi_company_abstract.py
```diff
@@ -59,16 +59,17 @@ def _name_search(
...
-            if type(arg) == list and arg[:2] == ["company_id", "in"]:
-                fix = []
-                for _i in range(len(arg[2]) - 1):
-                    fix.append("|")
-                for val in arg[2]:
-                    fix.append(["company_id", "=", val])
-                new_args.extend(fix)
-            else:
-                new_args.append(arg)
+        if args:
```
rousseldenis: args can be None.

base_multi_company/tests/test_multi_company_abstract.py
```diff
@@ -100,6 +100,12 @@ def test_search_company_id(self):
         )
         self.assertEqual(record, self.record_1)
 
+        name_result = self.test_model.name_search(
+            "test", [["company_id", "in", [self.company_2.id]]]
```
rousseldenis, line 104: @LoisRForgeFlow args domain can be a list of a list and not a list of tuples ?

rousseldenis: Ok, from js. But I think the problem was solved in 14 no ?

https://github.com/odoo/odoo/commit/69869ab681e2271ae802ca3ab7285ee83580208d#diff-fa4d9268d6e65e19aebec81c46038f0e496b91142588ed4d1c1bce7ff2338f2c

base_multi_company/models/multi_company_abstract.py
LoisRForgeFlow: I don't exactly recall now, but looking at the comment I added, I would say that something like `('company_id', 'in', [False, 1])` should be enough to reproduce the issue.

Let me know if that works, if not I would try to find the discussion when the problem was reported to me.

base_multi_company/models/multi_company_abstract.py
LoisRForgeFlow: that is more robust indeed

base_multi_company/tests/test_multi_company_abstract.py
LoisRForgeFlow: have you been able to reproduce in v13 with your test case, try to include a `False` in the "in" list?

rousseldenis:
WIP as I found an issue. Want to see if Odoo side or ours.

rousseldenis:
This one is ready and working with product_multi_company.

As there is a problem when evaluating domain like `['|', ('company_ids', '=', False), ('company_ids', 'in', company_ids)]`.  The first part is added as subquery and for an unknown reason does not return correct records.

Something has changed for v14 there : https://github.com/odoo/odoo/commit/69869ab681e2271ae802ca3ab7285ee83580208d#diff-fa4d9268d6e65e19aebec81c46038f0e496b91142588ed4d1c1bce7ff2338f2c

The idea here is to get a field that means 'no company is selected on record' as a boolean and evaluate it like that.

pedrobaeza:
Can you please take into account #313 ?

pedrobaeza:
I see that your last commit attempts to fix the same thing, but your patch seems incomplete for preserving compatibility.

rousseldenis:
@OCA/core-maintainers

rousseldenis:
> Can you please take into account #313 ?

Why company_id should be writeable, value should not to be taken into account.

rousseldenis:
> I see that your last commit attempts to fix the same thing, but your patch seems incomplete for preserving compatibility.

Do you have written a test case that shows company_ids could be void ? As your create and write part seem not useful https://github.com/OCA/multi-company/pull/313/files#diff-768c66d1106f5fdcb827fc104212e9b76a2739fae84f770ffe2687ec3bea1c4bR52

pedrobaeza:
Yes, it should be for compatibility with other modules that are not aware of this new field `company_ids`. Starting with, if you don't do it writeable, you will have a warning when installed with `sale`, as the field is used for a constraint. We traced other problems to this fact, and that's why we added the PR, and since then, everything goes smooth. If any module writes in `company_id`, it's reflected in company_ids as if you have put only one company (or none) there.

rousseldenis:
> Yes, it should be for compatibility with other modules that are not aware of this new field `company_ids`. Starting with, if you don't do it writeable, you will have a warning when installed with `sale`, as the field is used for a constraint. We traced other problems to this fact, and that's why we added the PR, and since then, everything goes smooth. If any module writes in `company_id`, it's reflected in company_ids as if you have put only one company (or none) there.

Could you give the real case as we didn't have problems ?

pedrobaeza:
I don't have the trace right now, but arose on integration tests when combined with several modules. I have already put you an example with `sale` module (it's just a warning, but it also means that you are disabling that constraint not doing it writeable). The create/write overriding is for giving always priority to `company_ids` definition instead of `company_id` if both are specified, which will mean any module aware of the multi company thing (or the module itself) can operate transparently to this `company_id` writable thing.

rousseldenis:
> I don't have the trace right now, but arose on integration tests when combined with several modules. I have already put you an example with `sale` module (it's just a warning, but it also means that you are disabling that constraint not doing it writeable). The create/write overriding is for giving always priority to `company_ids` definition instead of `company_id` if both are specified, which will mean any module aware of the multi company thing (or the module itself) can operate transparently to this `company_id` writable thing.

Ok, yes, I begin to see what  you want to say.

I will add your code and add tests too in another commit if you wnat to backport it

pedrobaeza:
The existing tests of the module lead me to add the create/write overriding analyzing the fail, so they are useful also for this, but it's OK to add some more if you see uncovered things.

GSLabIt (approved):
LGTM

pedrobaeza:
I'm missing the create/write part for discarding the double write of both `company_id` and `company_ids`.

pedrobaeza:
Sorry, wrong button

elvise:
> I'm missing the create/write part for discarding the double write of both `company_id` and `company_ids`.

@pedrobaeza how can we proceed towards the merge?

pedrobaeza:
@rousseldenis can you include such changes?

francesco-ooops:
hi @rousseldenis , any plans to finish this one?

rousseldenis:
> hi @rousseldenis , any plans to finish this one?

I'll try but I need to dig some time

LoisRForgeFlow:
@rousseldenis A friendly reminder here, have you got the time to finish this?

GSLabIt:
@rousseldenis did u have the chance to dive into this?

rousseldenis:
@GSLabIt @LoisRForgeFlow  @pedrobaeza This is done

pedrobaeza (approved):

base_multi_company/models/multi_company_abstract.py
```diff
@@ -0,0 +1,126 @@
...
+        # To allow modifying allowed companies by non-aware base_multi_company
+        # through company_id field we:
+        # - Add company to company_ids if not in existing ones (not removing
+        # existing ones as people that write company_id field has maybe
+        # no access to other ones.)
+        # - Remove all companies if company_id is False (give access to all)
+
+        for record in self:
+            company = record.company_id
+            if company and company not in record.company_ids:
```
pedrobaeza: Wait, this is not like my PR. We can't do make conditional that way, or we are modifying what is intended to do assigning a company, that is to change all the allowed companies to this one.

pedrobaeza (changes requested):

base_multi_company/models/multi_company_abstract.py
rousseldenis: That was done with that purpose. See comment before.

If people has no access to other companies, he cannot remove them.

pedrobaeza: But this is not OK. The intention to add the writeable company_id is to do exactly what is expected to do: limit the record to such company. This is working perfectly in v13. If not, please specify steps to reproduce a problem with this behavior.

rousseldenis: The behaviour changed in 13 from which it was at origin:

https://github.com/OCA/multi-company/commit/f5f3270ee8ab8139bd7344d1087f8872537332c7

You reintroduced the inverse BUT changing the behaviour.

See the original commit at creation in 10:

https://github.com/OCA/multi-company/commit/7051f69ec672aea10bbe3518916927fdbc15aba9#diff-768c66d1106f5fdcb827fc104212e9b76a2739fae84f770ffe2687ec3bea1c4bR45

rousseldenis: My change was, at purpose, restablishing the original behaviour.

pedrobaeza: Such original behavior was wrong, as it's modifying whole standard behavior. If any standard module not being aware of this multi-company think wants to set `company_id`, the expected behavior should be that the only allowed company must be that one, not that one + others not expected.

You are not adding any step to reproduce any problem with the behavior I'm mentioning. I will be glad to change my mind if you find steps that provokes a bug using my approach.

On contrary, this change will make lines like this one:

https://github.com/odoo/odoo/blob/4ad64719316bcd59c3961ed3b19d560a96215ff7/odoo/addons/base/models/res_users.py#L478

to not behave as expected.

rousseldenis: > You are not adding any step to reproduce any problem with the behavior I'm mentioning. I will be glad to change my mind if you find steps that provokes a bug using my approach.

On the contrary, I want a test from you that show me the wrong behaviour you 'corrected' as you didn't provided a test...

rousseldenis: @JoanSForgeFlow @LoisRForgeFlow @JordiBForgeFlow @JasminSForgeFlow

As you use/migrated this one in several versions, could you help here ? What is the expected behaviour with the write on company_id field ?

rousseldenis: I'll fix this next week. Need to sleep on it :-)

pedrobaeza: Thanks for considering it.

rousseldenis: @pedrobaeza I've taken your changes entirely.

I need now to improve README

pedrobaeza: Great, thanks. Let me know when this can be merged.

rafaelbn (approved):
👍 Thank you!

rousseldenis:
@pedrobaeza @Cedric-Pigeon @LoisRForgeFlow This one is ready

base_multi_company/README.rst
```diff
@@ -0,0 +1,108 @@
...
+    :target: https://translation.odoo-community.org/projects/multi-company-14-0/multi-company-14-0-base_multi_company
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/133/14.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module will provide a way to change the way Odoo manages a 'multi-company'
+implentation.
```
pedrobaeza:
```suggestion
implementation.
```

base_multi_company/README.rst
```diff
@@ -0,0 +1,108 @@
...
+--------
+
+Odoo traditional implementation of multi-company:
+
+- Some models contain a field named Company (company_id) that allows to set one company or None
+  in order to:
+  - Limit access to that company if set.
+  - not limiting access to any company if not set.
+
+This module breaks that in order to introduce a fine company access.
```
pedrobaeza: break sounds very hard

```suggestion
This module changes that in order to introduce a finer company access.
```

base_multi_company/README.rst
```diff
@@ -0,0 +1,108 @@
...
+e.g.: If you want to give record access to company A and B but not for C.
+
+This module is not doing anything by its own but provide a transversal implementation
+for further ones.
+e.g.: If you want to implement OCA multi-company behaviour for products, install
+also the 'product_multi_company' module.
```
pedrobaeza: You can also mention `partner_multi_company` as the 2 well known modules implementing this.

pedrobaeza (approved):

pedrobaeza:
Thanks, and I'm going to launch the bot with `patch` option for satisfying you although I don't like that glitch :)

/ocabot merge patch

rousseldenis:
> Thanks, and I'm going to launch the bot with `patch` option for satisfying you although I don't like that glitch :)
> 
> /ocabot merge patch

contribution is a location of compromises :smile: :rocket:

PR #277, opened 2021-05-07
[12.0] FIX purchase_sale_inter_company: method inter_company_create_invoice was not correctly ported from previous versions

eLBati:
So automatic vendor bills are not correctly linked to purchase orders

Later versions should be affected too


mrcast (approved):
Functional test ok!

eLBati:
@mourad-ehm what do you think?

PR #278, opened 2021-05-11
[12.0] base_multi_company: Fix compute visible for all companies

Sirfanas:
Just fix the compute, without that once it's set to True it's not changed anymore otherwise


pedrobaeza:
In v12, the default value on computes is applied, so the previous code was correct (although less explicit). What is the problem you find?

Sirfanas:
@pedrobaeza Yeah I test it, thought this method was the problem but in fact it was because of  #280 and / or #281 
I close it as it was a mistake, sorry !

PR #279, opened 2021-05-11
[12.0] purchase_sale_inter_company: Fix bug changing user's active company

Sirfanas:
When using it, user's company automticly change to the forced company, resulting in an access error at interface.

### DIFF
--- a/purchase_sale_inter_company/models/purchase_order.py
+++ b/purchase_sale_inter_company/models/purchase_order.py
@@ -17,11 +17,13 @@ class PurchaseOrder(models.Model):
         for purchase_order in self.sudo():
             # get the company from partner then trigger action of
             # intercompany relation
+            initial_company = self.env.user.company_id
             dest_company = purchase_order.partner_id.ref_company_ids
             if dest_company and dest_company.so_from_po:
                 purchase_order.with_context(
                     force_company=dest_company.id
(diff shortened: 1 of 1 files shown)


pedrobaeza:
That's not the way to fix things. First start indicating clear steps to reproduce the problem just using this module without the influence of any other one.

Sirfanas:
@pedrobaeza Sorry forgot to discribe how to reproduce ^^

Installed a fresh 12.0 database (with odoo official docker image builded at [20210506](https://github.com/odoo/docker/commit/c963853d004e7cb517e8229b60fd3e29197fdac9))
Install purchase_sale_inter_company

Let's say I have company A and company B (on same instance with multi-company)

My current user's active company is A
Try to make a purchase order to company B
Then when I confirm it, I got an access error for my current purchase.order record
Refresh my page and now my active company was B !

After the code I add I had the exptected behavior, and my current company didn't change.

I suppose it's not the best way to change current company, I'll look for a cleaner way to do it and push that !

pedrobaeza:
Module is already prepared for that scenario, and it's working through 2 mechanisms:

- `sudo`
- and `context(force_company=)`

so you are probably not using the right version, or there's something misconfigured (like a product not accessible to both companies).

PR #280, opened 2021-05-11
[12.0] partner_multi_company: Force recompute visible_for_all_companies after installation

Sirfanas:
Force recompute field visible_for_all_companies at installation, wasn't well target by the depends, so even if company_ids is auto filled, field wasn't compute back


pedrobaeza:
Duplicated of #281

PR #281, opened 2021-05-11
[12.0] product_multi_company: Force recompute visible_for_all_companies after installation

Sirfanas:
Force recompute field visible_for_all_companies at installation, wasn't well target by the depends, so even if company_ids is auto filled, field wasn't compute back


pedrobaeza:
Please don't shoot blindly to things that are not a fix, at least not for the expected problem. Again put steps to reproduce a problem in first place, and then the fix will explain itself. Meanwhile, this fix is assumed as not valid.

PR #283, opened 2021-05-21
[IMP][14.0][account_invoice_inter_company] add pdf invoices attachments in both companies

Kev-Roche:
cc @sebastienbeau 

This PR adds the invoice (pdf format) in attachments for both companies

### DIFF
--- a/account_invoice_inter_company/models/account_move.py
+++ b/account_invoice_inter_company/models/account_move.py
@@ -1,5 +1,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import base64
+
 from odoo import _, api, fields, models
 from odoo.exceptions import AccessError, UserError
 from odoo.tests.common import Form
@@ -45,16 +47,34 @@ class AccountMove(models.Model):
             # do not consider invoices that have already been auto-generated,
             # nor the invoices that were already validated in the past
             dest_company = src_invoice._find_company_from_invoice_partner()
-            if not dest_company or src_invoice.auto_generated:
-                continue
-            intercompany_user = dest_company.intercompany_invoice_user_id
-            if intercompany_user:
-                src_invoice = src_invoice.with_user(intercompany_user).sudo()
-            else:
-                src_invoice = src_invoice.sudo()
-            src_invoice.with_company(dest_company.id).with_context(
-                skip_check_amount_difference=True
-            )._inter_company_create_invoice(dest_company)
+            if dest_company:
+                if not src_invoice.auto_generated:
+                    intercompany_user = dest_company.intercompany_invoice_user_id
+                    if intercompany_user:
+                        src_invoice = src_invoice.with_user(intercompany_user).sudo()
+                    else:
+                        src_invoice = src_invoice.sudo()
+                    src_invoice.with_company(dest_company.id).with_context(
+                        skip_check_amount_difference=True
+                    )._inter_company_create_invoice(dest_company)
+                if src_invoice.move_type in ["out_invoice", "out_refund"]:
+                    src_invoice._attach_original_pdf_report()
+
+    def _attach_original_pdf_report(self):
+        supplier_invoice = self.auto_invoice_id
+        if not supplier_invoice:
+            supplier_invoice = self.search([("auto_invoice_id", "=", self.id)], limit=1)
+        pdf = self.env.ref("account.account_invoices")._render_qweb_pdf([self.id])[0]
+        self.env["ir.attachment"].create(
+            {
+                "name": self.name + ".pdf",
+                "type": "binary",
+                "datas": base64.b64encode(pdf),
+                "res_model": "account.move",
+                "res_id": supplier_invoice.id,
+                "mimetype": "application/pdf",
+            }
+        )
 
     def _check_intercompany_product(self, dest_company):
         self.ensure_one()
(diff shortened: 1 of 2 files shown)


account_invoice_inter_company/models/account_move.py
```diff
@@ -115,7 +117,19 @@ def _inter_company_create_invoice(self, dest_company):
...
+                    "name": self.name + ".pdf",
+                    "type": "binary",
+                    "datas": base64.b64encode(pdf),
+                    "res_model": "account.move",
+                    "res_id": dest_invoice.id,
+                    "mimetype": "application/pdf",
+                }
+            )
             dest_invoice.action_post()
+            return attachment_invoice_pdf
```
sebastienbeau: return here should be removed

sebastienbeau (changes requested):

account_invoice_inter_company/models/account_move.py
```diff
@@ -115,7 +117,19 @@ def _inter_company_create_invoice(self, dest_company):
         if dest_company.invoice_auto_validation and not float_compare(
             self.amount_total, dest_invoice.amount_total, precision_digits=precision
         ):
+            pdf = self.env.ref("account.account_invoices")._render_qweb_pdf(self.id)[0]
```
sebastienbeau: In fact it's not correct.
Indeed in case of manual validation the dest company will not have any pdf.
It's will better to inherit the method "action_post" and in if the field "auto_invoice_id" is fill, in that case call a method "_attach_original_pdf_report" that add the pdf in teh attachment.

Also having the logic in a specific method give us the possibility to use it in specific migration script

account_invoice_inter_company/models/account_move.py
Kev-Roche: done

account_invoice_inter_company/models/account_move.py
Kev-Roche: I m not sure it is that simple...

account_invoice_inter_company/models/account_move.py
```diff
@@ -52,6 +56,19 @@ def action_post(self):
             )._inter_company_create_invoice(dest_company)
         return res
 
+    def _attach_original_pdf_report(self):
+        pdf = self.env.ref("account.account_invoices")._render_qweb_pdf([self.id])[0]
```
sebastienbeau, line 67: I thinks there is something wrong here.
We can have several case of invoice 
Out and in invoice.

- Example 1: company A create an out invoice company b. We have to attach the pdf of company A into company B

- Example 2: (more rare case) company B generate an in invoice and this generate the out invoice in company A. We have to attach the pdf of company A into company B

In both case we attach the pdf of company A (out invoice) into company B (in invoice)
But the value of auto_invoice_id is not the same!

sebastienbeau (changes requested):

account_invoice_inter_company/models/account_move.py
sebastienbeau: Also take in account refund case

rafaelbn:
Thank you very much for this improvements 👍 . I kindly suggest to write the name of the module `account_invoice_inter_company` in the tittle of the PR to increase visibility and reviewers ❤️ . Thank you again

account_invoice_inter_company/models/account_move.py
Kev-Roche: As you suggested, the module take into account, both cases.

account_invoice_inter_company/models/account_move.py
Kev-Roche: As discussed together, I bring another solution.

account_invoice_inter_company/models/account_move.py
```diff
@@ -45,16 +47,36 @@ def create_counterpart_invoices(self):
...
+                        src_invoice = src_invoice.with_user(intercompany_user).sudo()
+                    else:
+                        src_invoice = src_invoice.sudo()
+                    src_invoice.with_company(dest_company.id).with_context(
+                        skip_check_amount_difference=True
+                    )._inter_company_create_invoice(dest_company)
+                if src_invoice.move_type in ["out_invoice", "out_refund"]:
+                    src_invoice._attach_original_pdf_report()
+
+        return res
```
sebastienbeau: return res is useless

Kev-Roche: Indeed, I remove it.

sebastienbeau (approved):
LGTM code review and test

bguillot (approved):
Thanks !

account_invoice_inter_company/models/account_move.py
```diff
@@ -45,16 +47,34 @@ def create_counterpart_invoices(self):
...
+                    intercompany_user = dest_company.intercompany_invoice_user_id
+                    if intercompany_user:
+                        src_invoice = src_invoice.with_user(intercompany_user).sudo()
+                    else:
+                        src_invoice = src_invoice.sudo()
+                    src_invoice.with_company(dest_company.id).with_context(
+                        skip_check_amount_difference=True
+                    )._inter_company_create_invoice(dest_company)
+                if src_invoice.move_type in ["out_invoice", "out_refund"]:
+                    src_invoice._attach_original_pdf_report()
```
omar7r, line 61: For doing that, you need sudo permissions always. If you don't enter by first "if" you don't have it, and it fails. If you don't post automatically, when you post manually the customer's invoice renters here but only in second "if"
In the other hand, if you don't post automatically, you mustn't attach anything, only attach if posted, the second time. Check if posted too.

PR #285, opened 2021-06-03
[FIX] account_invoice_inter_company: missing migration script

chienandalu:
- Without the migration script we lose the link between invoices, so if
we modified the origin we could be duplicating the destination invoice!

- Now when we've posted a move, we can set it back to draft to edit some
things and the repost it. In that case, we don't want to recreate the
destination invoice if it's already posted and maybe even payed.

- As invoices and the classic move entries are the same model now, we
should filter the later ones to avoid triggering multi-company entries, which
aren't supported.

cc @Tecnativa TT30195

ping @pedrobaeza @carlosdauden


chienandalu:
The runbot warning seems to be on `product_multi_company`

carlosdauden (approved):

account_invoice_inter_company/migrations/13.0.3.0.0/post-migration.py
```diff
@@ -0,0 +1,23 @@
...
+    openupgrade.logged_query(
+        """
+            UPDATE account_move am
+            SET
+                auto_generated = ai.auto_generated,
+                auto_invoice_id = am2.id
+            FROM account_invoice ai
+                JOIN account_invoice ai2 ON ai.auto_invoice_id = ai2.id
+                JOIN account_move am2 ON am2.old_invoice_id = ai2.id
+            WHERE am.old_invoice_id = ai.id AND ai.auto_generated IS NOT NULL
```
pedrobaeza:
```suggestion
            WHERE am.old_invoice_id = ai.id AND ai.auto_generated IS True
```
or 
```suggestion
            WHERE am.old_invoice_id = ai.id AND ai.auto_invoice_id IS NOT NULL
```

chienandalu:
Added a commit with this fix:

Now when we've posted a move, we can set it back to draft to edit some
things and the repost it. In that case, we don't want to recreate the
destination invoice if it's already posted and maybe even payed.

account_invoice_inter_company/migrations/13.0.3.0.0/post-migration.py
chienandalu: Ok, changes done

pedrobaeza:
But I think we should forbid such edition, as there's the possibility of having different amounts...

chienandalu:
> But I think we should forbid such edition, as there's the possibility of having different amounts...

Yes, but maybe the change is a trivial one. Maybe we should warn the user about the destination invoice an let them decide?

pedrobaeza:
No, for me it's clear: if the destination invoice is posted, then you can unpost the source one. If not posted, then you can do it, and then regenerate the other side one.

account_invoice_inter_company/models/account_move.py
```diff
@@ -197,6 +197,24 @@ def _prepare_invoice_data(self, dest_company):
         )
         return vals
 
+    def button_draft(self):
+        for move in self:
+            company = move._find_company_from_invoice_partner()
```
pedrobaeza: Why the need of searching for the company? Isn't `move.auto_generated` enough to know that a company is linked?

chienandalu:
> No, for me it's clear: if the destination invoice is posted, then you can unpost the source one. If not posted, then you can do it, and then regenerate the other side one.

Review the new changes. Anyway, I wonder what chance could we give to the user to make legitimate data change that requieres to set draft the origin invoice.

Also, there's never a check in the opposite direction, so we could be unbalancing the changes from the other side.

account_invoice_inter_company/models/account_move.py
```diff
@@ -197,6 +197,24 @@ def _prepare_invoice_data(self, dest_company):
...
+                    [("auto_invoice_id", "=", move.id), ("state", "=", "posted")]
+                )
+                if inter_invoice_posted:
+                    raise UserError(
+                        _(
+                            "You can't modify this invoice as it has an inter company "
+                            "invoice that's in posted state.\n"
+                            "Invoice %s to %s"
+                        )
+                        % (company.name, inter_invoice_posted.name)
```
pedrobaeza: The order of the parameters is switched and you don't need `company`, just `inter_invoice_posted.partner_id.display_name`.

pedrobaeza (changes requested):

pedrobaeza:
Yes, there's no good option here indeed. I think you can prevent any manual modification on an auto invoice.

account_invoice_inter_company/models/account_move.py
chienandalu: Well, indeed that `auto_generated` field is not very useful, as we can deduce that from the `auto_invoice_id` already.

account_invoice_inter_company/models/account_move.py
chienandalu: sure :+1:

chienandalu:
> Yes, there's no good option here indeed. I think you can prevent any manual modification on an auto invoice.

Should we have a field whitelist at least?

pedrobaeza:
I think we can do the contrary: only forbid if a different amount is detected.

chienandalu:
Ok, let's see now (I should add some tests)

account_invoice_inter_company/models/account_move.py
```diff
@@ -214,6 +233,29 @@ def button_cancel(self):
                     inter_invoice.button_cancel()
         return super().button_cancel()
 
+    def write(self, vals):
+        res = super().write(vals)
+        for move in self.filtered(
+            lambda x: x.sudo().auto_invoice_id and x.state == "posted"
```
pedrobaeza: Remove the posted check. In fact, you can only modify an invoice in draft state.

chienandalu:
Please review again

pedrobaeza (approved):

account_invoice_inter_company/models/account_move.py
```diff
@@ -33,7 +33,8 @@ def _find_company_from_invoice_partner(self):
     def action_post(self):
         """ Validated invoice generate cross invoice base on company rules """
         res = super().action_post()
-        for src_invoice in self:
+        # Intercompany account entries aren't supported
+        for src_invoice in self.filtered(lambda x: x.type != "entry"):
```
pedrobaeza: This should be explicit to the 4 types, as there can be also other types:

https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L112

chienandalu: Changes done

pedrobaeza (approved):

pedrobaeza:
@carlosdauden please review

carlosdauden (approved):

pedrobaeza:
Manually cherry-picked last 2 commits for v14 in 3bc08cbb6ba93124c2e16eb37f35ecc297954058 and 3b7264711f036f90d7a0c6d4b01f7472b8d6cfb2, and also added the missing test patch: 5686a4f9fee10ea340f250222af031df17ff36b3

PR #288, opened 2021-06-08
[FIX] partner_multi_company: reconciliation compatibility

chienandalu:
This line from the reconciliation widget
https://github.com/odoo/odoo/blob/12.0/addons/account/models/reconciliation_widget.py#L103
substitutes any occurrence of res_partner that could come in the
where_clause coming from ir_rules with the `p3` join alias of the
subsequent query.

In the case we've got a subselect, that alias won't make sense. So we're
preventing that situation renaming the table and the affected column
where such substitution was being made.

cc @Tecnativa TT30346

please review @pedrobaeza @carlosdauden

### DIFF
--- a/partner_multi_company/__manifest__.py
+++ b/partner_multi_company/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Partner multi-company",
     "summary": "Select individually the partner visibility on each company",
-    "version": "12.0.1.0.1",
+    "version": "12.0.2.0.0",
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

LoisRForgeFlow (approved):
LGTM, in v13 that has been fixed in core here:

https://github.com/odoo/odoo/commit/fe3a35006b8397425a702333cf4115a80171f94e

PR #290, opened 2021-06-23
 [ADD] [12.0] pos_restaurant_multi_company

legalsylvain:
trivial module

* add ``company_id`` field on ``pos_restaurant`` models (``restaurant.floor`` and ``restaurant.table`` (related) and ``restaurant.printer``)

* add classical rule

CC : @OCA/pos-maintainers 

Ref GRAP : apps/deck/#/board/144/card/1634 CC @quentinDupont


pos_restaurant_multi_company/models/restaurant_printer.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (C) 2021 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class RestaurantPrinter(models.Model):
+    _inherit = "restaurant.floor"
```
ivantodorovich:
```suggestion
    _inherit = "restaurant.printer"
```

pos_restaurant_multi_company/models/restaurant_floor.py
```diff
@@ -0,0 +1,17 @@
...
+
+class RestaurantFloor(models.Model):
+    _inherit = "restaurant.floor"
+
+    company_id = fields.Many2one(
+        "res.company",
+        string="Company",
+        index=True,
+        default=lambda self:
+            self.env["res.company"]._company_default_get("restaurant.floor"),
```
ivantodorovich: I would go for `default=lambda self: self.env.user.company_id`, here and in the other cases

To be honest I didn't know about `_company_default_get` but I see it's already deprecated in `14.0`: https://github.com/odoo/odoo/commit/a5b6f31cf28e5381e1c85f66730bcdb55998e643

I would also add a constraint to check, if `pos_config_id` is set, that they belong to the same company

Maybe even an onchange to automatically set the company from the pos.config, when set

pos_restaurant_multi_company/models/restaurant_printer.py
```diff
@@ -0,0 +1,17 @@
...
+
+    company_id = fields.Many2one(
+        "res.company",
+        string="Company",
+        index=True,
+        default=lambda self:
+            self.env["res.company"]._company_default_get("restaurant.printer"),
+    )
```
ivantodorovich, line 16: I think the `ir.rule` and views for the `restaurant.printer` are missing

ivantodorovich (changes requested):
:+1:

pos_restaurant_multi_company/models/restaurant_printer.py
legalsylvain: you're right ! added.

pos_restaurant_multi_company/models/restaurant_floor.py
legalsylvain: To be honest, I don't know. I copied the code from other module of the same repo. the standards seemed to be ``_company_default_get``, but I think it's the same. So I simplified the code, and make it compatible with V14.

thanks !

legalsylvain: constraint and onchange added.

pos_restaurant_multi_company/models/restaurant_printer.py
legalsylvain: done. thanks !

ivantodorovich (approved):
LGTM :+1: :tada:  Thanks !!

quentinDupont (approved):
Use in prod since more than one year, works like a charm :heavy_check_mark:

PR #292, opened 2021-06-30
[12.0][FIX] partner_multi_company CI tests fails

chienandalu:
If we enable the multicompany rules, some subsequent tests will fail
unexpectedly if they are designed for a shared contacts environment. So
we relay on the rewritten rule current value and don't enable it until
the user does it via res.config.settings (Common Contact Book).

cc @Tecnativa TT23741

cc @pedrobaeza @carlosdauden

### DIFF
--- a/base_multi_company/hooks.py
+++ b/base_multi_company/hooks.py
@@ -29,8 +29,10 @@ def set_security_rule(env, rule_ref):
     rule = env.ref(rule_ref)
     if not rule:  # safeguard if it's deleted
         return
+    # We don't want to enable the rule if it isn't already active, as this leads
+    # to subsequent test fails in a CI context. This should be done via the core
+    # res.config.settings methods
     rule.write({
-        'active': True,
         'domain_force': (
             "['|', ('company_ids', 'in', user.company_id.ids),"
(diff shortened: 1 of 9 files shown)


pedrobaeza:
Please check Travis

StefanRijnhart:
Restarted Travis after the problem with the f-string was fixed because you signaled it to the Odoo developer. Thanks for that!

base_multi_company/hooks.py
```diff
@@ -29,8 +29,10 @@ def set_security_rule(env, rule_ref):
     rule = env.ref(rule_ref)
     if not rule:  # safeguard if it's deleted
         return
+    # We don't want to enable the rule if it isn't already active, as this leads
+    # to subsequent test fails in a CI context. This should be done via the core
+    # res.config.settings methods
     rule.write({
```
pedrobaeza, line 35: I think you should activate it if you are not in test environment:
```suggestion
    from odoo.tools import config
    if not config["test_enable"]:
```

carlosdauden (approved):
Attend this:
https://github.com/OCA/multi-company/pull/292/files#r663538848

PR #271, opened 2021-03-09
[14.0][MIG] product_multi_company

rousseldenis:
Includes:

- [x] #270

PR #294, opened 2021-07-06
[14.0][ADD] mrp_multi_company

rousseldenis:
Includes:

- [ ] #271


pedrobaeza:
Shouldn't it be `mrp_bom_multi_company`?

rousseldenis:
> Shouldn't it be `mrp_bom_multi_company`?

I thought about it but my first approach was to see if including only boms is sufficent.

Maybe this one can grew to include missing features if needed

pedrobaeza:
Even if adding for example route multi-company capability, I think that should be other module for not getting a very big one difficult to maintain. About BOMs, you have a hard task assuring the integrity of all components through the companies...

rousseldenis:
> About BOMs, you have a hard task assuring the integrity of all components through the companies...

Of course, but that's product manager job (they currently doing it). I had idea to put the company_ids on boms directly (not related) but that should have been the bottleneck. Through related, the job is done on product and that's all.

PR #295, opened 2021-07-06
[14.0] intercompany_shared_contact: add

kevinkhao:
This module automatically shares all the contacts originating from companies (res.company object).

For example, suppose we have a multi-company environment with two companies (res.company) A and B in Odoo.
We have users uA, uB (belonging to their respective companies).

Normally, uA cannot access the contact B.
With this module, uA can see B (and inversely uB can see A).
uA still can't see all the other (supplier/reseller) contacts of company B.

### DIFF
--- a/intercompany_shared_contact/__init__.py
+++ b/intercompany_shared_contact/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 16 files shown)


sebastienbeau:
@kevinkhao I close this one and open a new one with the history https://github.com/OCA/multi-company/pull/315.
Easier to review and to keep a trace of the logic done for creating the module

PR #269, opened 2021-03-03
[MIG][14.0] product_tax_multicompany_default

enriquemartin:
standard migration

PR #300, opened 2021-07-12
[14.0][MIG] product_tax_multicompany_default: migration to 14.0

enriquemartin:
standard migration
old PR https://github.com/OCA/multi-company/pull/269


enriquemartin:
@StefanRijnhart I think that this would be the correct way

pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

StefanRijnhart:
Indeed, the previous attempt (https://github.com/OCA/multi-company/pull/269, before the rebase) was fine *except* for the commit message of the migration commit.

PR #258, opened 2020-12-04
[MIG] 14.0 purchase_sale_inter_company

PierrickBrun:
Depends on https://github.com/OCA/multi-company/pull/257

PR #307, opened 2021-07-26
[12.0][IMP] puchase_sale_inter_company: option not to overwrite PO prices

MPishpecki:


pedrobaeza:
I don't get the use case for adding such option. If you have that need, do it on your module overwriting the behavior. For me this adds an unneeded complexity to the base module.

MPishpecki:
Hi Pedro, we needed to remove the overwriting of the PO prices in our setup also checking if the PO and So prices are in line. As we did it by overwriting the  purchase_sale_inter_company module, we thought it may be useful for someone else in similar situation. We try to preserve the original functionality by default.

pedrobaeza:
Thanks for sharing this, but I think that in this case, "bloating" the module with such option is not very useful, but that's my opinion. Maybe others consider it useful and decide to go ahead with the review.

StefanRijnhart:
@pedrobaeza it was this comment by @enriquemartin that made me think that there would be interest from the community for this feature: https://github.com/OCA/multi-company/pull/258#discussion_r586560764, which is why I asked @MPishpecki to make this PR.

StefanRijnhart:
@MPishpecki please remove the internal ticket number

pedrobaeza:
But I think it's normal that if the sales price changes, then it's synced back to the PO, as that's going to be the final price, and it's also the real flow: you do the PO not knowing the exact price until your vendor tells you which is.

StefanRijnhart:
@pedrobaeza yes, we agree that the prices should agree eventually but some of us think that they should not be modified silently from the other company's sale order. It's just another option and we're keeping the original behaviour as a default.

pedrobaeza:
As stated, for me this is bloating the original module. If you want such behavior, then just override the module. You can put the proper hooks though here if needed. I'm not blocking anyway if others think that this is useful.

purchase_sale_inter_company/models/sale_order.py
```diff
@@ -3,7 +3,8 @@
 # Copyright 2018-2019 Tecnativa - Carlos Dauden
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import _,api, fields, models
```
StefanRijnhart: purchase_sale_inter_company/models/sale_order.py:6:19: E231 missing whitespace after ','

purchase_sale_inter_company/models/sale_order.py
```diff
@@ -16,12 +17,42 @@ class SaleOrder(models.Model):
...
     )
 
+    def assert_intercompany_prices_equal(self):
+        """Check if the prices of both orders are the same"""
+        unequal_line = []
+        for so in self:
+            for line in so.sudo().order_line:
+                if not so.currency_id.compare_amounts(
+                    line.price_unit, line.auto_purchase_line_id.price_unit):
+                    continue
```
StefanRijnhart, line 29: purchase_sale_inter_company/models/sale_order.py:26:5: E125 continuation line with same indent as next logical line

MPishpecki: moved

StefanRijnhart:
@MPishpecki suggest as a commit message '[IMP] puchase_sale_inter_company: option not to overwrite PO prices'

purchase_sale_inter_company/models/sale_order.py
StefanRijnhart: The error is now 
`purchase_sale_inter_company/models/sale_order.py:28:24: E126 continuation line over-indented for hanging indent`
Looks like you have three spaces there instead of four.

StefanRijnhart (approved):

astirpe (approved):

StefanRijnhart:
@MPishpecki can you resolve the conflicts?

PR #309, opened 2021-08-25
[14.0] add stock_intercompany module

ajaniszewska-dev:


stock_intercompany/models/stock_picking.py
```diff
@@ -0,0 +1,26 @@
...
+                company = partners[picking.partner_id]
+                warehouse = self.env["stock.warehouse"].search([('company_id','=',company.id)], limit=1)
+                vals = {
+                    'company_id': company.id,
+                    'picking_type_id': company.intercompany_in_type_id or warehouse.in_type_id.id,
+                    'state':'draft',
+                }
+
+                new_picking_vals = picking.copy_data(default=vals)
+                stock_move = self.env['stock.picking'].create(new_picking_vals)
```
ajaniszewska-dev: @yvaucher  now i guess stock.picking needs to have created stock.move and stock.package.quant for partner's company and assigned to move_lines/move_ids_without package?  i think location will be taken automatically for partner on picking since it is readonly param.

i dont understand this part:

" You may have to duplicate moves 1 by 1 changing the company and it's src and destination locations." 

then we duplicate the moves or we create the moves for picking?

gurneyalex: you create new moves for the new picking. I assume what Yannick means is that the easiest way is to copy the moves from the picking being processed, changing some values, as this will copy some fields that can be added by other addons that you are not aware of.

gurneyalex:
```suggestion
                reception_picking = self.env['stock.picking'].create(new_picking_vals)
```

stock_intercompany/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+
+        for picking in self:
+            if picking.partner_id in partners.keys():
+                company = partners[picking.partner_id]
+                warehouse = self.env["stock.warehouse"].sudo().search([('company_id','=',company.id)], limit=1)
+                vals = {
+                    'company_id': company.id,
+                    'picking_type_id': company.intercompany_in_type_id.id  or warehouse.in_type_id.id,
+                    'state':'draft',
+                    'location_id': self.env['stock.location'].sudo().search([('name','=','Vendors')]).id,
```
ajaniszewska-dev: @yvaucher  not sure about this location, there is Vendor location but no company_id on it.

stock_intercompany/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+                    'location_dest_id': self.env['stock.location'].sudo().search([('name','=','Stock'),('company_id','=',company.id)]).id,
+                }
+
+                new_picking_vals = picking.sudo().copy_data(default=vals)
```
ajaniszewska-dev: @yvaucher  as far as i see we have always 2 counter- picking on partner company created. probably copy_data creates one? Will check it soon.

stock_intercompany/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2021 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Stock Intercompany Delivery-Reception",
+    "Summary": "Module that adds possibility for intercompany Delivery-Reception",
+    "version": "14.0.1.1.1",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
```
yvaucher:
```suggestion
    "website": "https://github.com/OCA/multi-company",
```

stock_intercompany/models/stock_picking.py
yvaucher: It's fine partner location are generic.
Would be better to do it as done in `stock` module. You can rely on the XML ids.

As done in:
https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_picking.py#L177-L182

stock_intercompany/models/stock_picking.py
yvaucher: ~Maybe related to some recursive call?~

EDIT: related to the wizard when no qty are selected. -> better to override `_action_done`

stock_intercompany/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+                reception_picking = self.env['stock.picking'].sudo().create(new_picking_vals)
+                reception_picking.action_confirm()
```
yvaucher: Probably better to keep the reception picking in draft so the destination company user can handle in with more options.

yvaucher (changes requested):
Some remarks on the code.

Please also install `pre-commit`

stock_intercompany/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+        return super(StockPicking,self).button_validate()
```
yvaucher: Would make more sense to move the `super` prior to the creation of the reception picking.

stock_intercompany/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
+from odoo import api, fields, models
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+
+    #override of method from stock module
+    def button_validate(self):
```
yvaucher: Better to move this part in a dedicated method + override `_action_done` insead of `button_validate`

stock_intercompany/tests/test_intercompany_picking.py
```diff
@@ -0,0 +1,99 @@
...
+                "company_ids": [(4, self.company1.id), (4, self.company2.id)],
+            }
+        )
+        # create storable product
+        self.product1 = self.env["product.product"].create(
+            {
+                "name": "Product A",
+                "type": "product",
+                "categ_id": self.env.ref("product.product_category_all").id,
+                "qty_available": 100,
```
ajaniszewska-dev, line 50: don't understand why checking product with default_company_id inside with_context(...) shows qty_available -1.0 for both company1 and company2. with this not possible to make a  assert and check product quantities on both companies when pickings are created.

stock_intercompany/views/res_config_settings.xml
```diff
@@ -0,0 +1,25 @@
...
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="stock.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@name='operations_setting_container']" position="inside">
+                <div
+                    class="col-12 col-lg-6 o_setting_box"
+                    id="intercompany_delivery_sbox"
+                >
+                    <div class="o_setting_right_pane">
+                        <label for="intercompany_in_type_id" />
```
gurneyalex, line 15:
```suggestion
                        <label for="intercompany_in_type_id" />
                        <span class="fa fa-lg fa-building-o" title="Values set here are company-specific." groups="base.group_multi_company"/>
```

this is important for UI consistency: this is set on all res.config.settings parameters which are related fields on the current company.

stock_intercompany/models/stock_picking.py
```diff
@@ -0,0 +1,33 @@
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def create_counterpart_picking(self):
+        companies = self.env["res.company"].sudo().search([])
+        partners = {cp.partner_id: cp for cp in companies}
+        if self.partner_id in partners.keys():
```
gurneyalex:
```suggestion
        if self.partner_id in partners:
```

the "in" operator searches on keys in a dictionary

stock_intercompany/models/stock_picking.py
```diff
@@ -0,0 +1,33 @@
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def create_counterpart_picking(self):
```
gurneyalex:
```suggestion
    def _create_counterpart_picking(self):
```

this is an internal method, which should not be exposed to RPC

stock_intercompany/models/stock_picking.py
```diff
@@ -0,0 +1,33 @@
...
+                "location_dest_id": warehouse.lot_stock_id.id,
+            }
+            new_picking_vals = self.sudo().copy_data(default=vals)
+            picking = self.env["stock.picking"].sudo().create(new_picking_vals)
+            picking.action_confirm()
+
+    # override of method from stock module
+    def _action_done(self):
+        for picking in self:
+            picking.create_counterpart_picking()
```
gurneyalex:
```suggestion
            picking._create_counterpart_picking()
```

stock_intercompany/models/stock_picking.py
```diff
@@ -0,0 +1,33 @@
...
+    def create_counterpart_picking(self):
+        companies = self.env["res.company"].sudo().search([])
+        partners = {cp.partner_id: cp for cp in companies}
+        if self.partner_id in partners.keys():
+            company = partners[self.partner_id]
+            warehouse = (
+                self.env["stock.warehouse"]
+                .sudo()
+                .search([("company_id", "=", company.id)], limit=1)
+            )
```
gurneyalex: I think it is better to use the warehouse of the picking type we configured in the res.config.settings:
```suggestion
            if `company.intercompany_in_type_id.warehouse_id:`
                warehouse = company.intercompany_in_type_id.warehouse_id
            else:
                warehouse = (
                    self.env["stock.warehouse"]
                    .sudo()
                    .search([("company_id", "=", company.id)], limit=1)
                )
```

stock_intercompany/models/stock_picking.py
gurneyalex: (!) Actually there is another issue here: I think that this code we will create a counterpart also if we purchase from another company, but the counterpart will be wrong, as it will be a reception picking on the other company (which is a supplier in this case).

For now, I suggest checking that the current picking is a shipping picking (for instance by checking that the location_dest_id of the picking is a Customers virtual location). If not, then the method should do nothing. 

THIS NEEDS FIXING.

stock_intercompany/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,4 @@
+* Go to Inventory > Settings
+* Under 'Operations' set
+
+  * Intercompany operation type
```
gurneyalex: add information about the creation of the picking type (which type of picking type has to be used, how to create it), on which company you need to set it.

gurneyalex (changes requested):

gurneyalex (approved):

leemannd:
Hello @ajaniszewska-dev  I have a small fix to be pushed to your PR -> https://github.com/ajaniszewska-dev/multi-company/pull/1

stock_intercompany/models/stock_picking.py
```diff
@@ -0,0 +1,49 @@
...
+                counterparts.append((picking, counterpart))
+        res = super(StockPicking, self)._action_done()
+        for picking, counterpart in counterparts:
+            picking._finalize_counterpart_picking(counterpart)
+        return res
+
+    def _finalize_counterpart_picking(self, counterpart_picking):
+        """hook to finalize required steps on the counterpart picking after the initial
+        outgoing picking is done"""
+        pass
```
leemannd: Pre-commit failing on this point

leemannd (approved):
LGTM unless the pre-commit hook failing

PR #305, opened 2021-07-20
[12.0][purchase_sale_inter_company] fix purchase line link with vendor bill

JoanSForgeFlow:
The method had not been correctly migrated, and as a consequence the vendor bills were not being linked with the POs
Back-port of #303

PR #311, opened 2021-09-03
[12.0][FIX] account_invoice_inter_company: avoid issue writing the company_id

AdriaGForgeFlow:
This PR intends to fix an issue in Travis (https://github.com/OCA/multi-company/pull/305) making fail the tests when the module account_bill_line_distribution is installed.

image

cc ~ @ForgeFlow

### DIFF
--- a/account_invoice_inter_company/views/account_invoice_views.xml
+++ b/account_invoice_inter_company/views/account_invoice_views.xml
@@ -6,6 +6,10 @@
         <field name="inherit_id" ref="account.invoice_supplier_form"/>
         <field name="mode">primary</field>
         <field name="arch" type="xml">
+            <!-- Enforce this to avoid issues writing the company_id-->
+            <xpath expr="//field[@name='invoice_line_ids']/tree" position="attributes">
+                <attribute name="editable">bottom</attribute>
+            </xpath>
             <xpath expr="//field[@name='invoice_line_ids']/tree/field[@name='company_id']" position="attributes">
                 <attribute name="readonly">0</attribute>
             </xpath>


pedrobaeza (changes requested):
The field needs to be written because there are other fields that depends on it that are not correctly computed. The problem you mention should be by a view with higher priority that provokes to not select it correctly for the `Form`. Please check that instead.

AdriaGForgeFlow:
> The field needs to be written because there are other fields that depends on it that are not correctly computed. The problem you mention should be by a view with higher priority that provokes to not select it correctly for the `Form`. Please check that instead.

You are right. 
What do you think about the new solution, it's the only way I found but maybe you can think of a better approach. I tried working with priorities on the views but did not work.

pedrobaeza:
I don't think that's a good solution, but unless you put the steps to reproduce the problem, we can't provide alternatives.

AdriaGForgeFlow:
> I don't think that's a good solution, but unless you put the steps to reproduce the problem, we can't provide alternatives.

To reproduce the issue:

Using the branch from this PR: https://github.com/OCA/multi-company/pull/305 which fixes an issue with a wrongly migrated function.

In a fresh Odoo instance, install the modules purchase_sale_inter_company and the module account_bill_line_distribution. You can then run the tests for purchase_sale_inter_company and you will have one FAILED test. Otherwise, you can configure the Odoo and create an inter-company invoice. When you try to post that invoice you will also see the issue in the logs complaining about writing in a field that is readonly. 

image

pedrobaeza:
This fix #313 should be the solution for this (it needs to be back-ported though).

AdriaGForgeFlow:
> This fix #313 should be the solution for this (it needs to be back-ported though).

I have tried but it does not fix the issue, the code never enters that create / write and the inverse is already defined in v12.

PR #313, opened 2021-09-23
[13.0][FIX] base_multi_company: Make company_id writeable again

pedrobaeza:
Not being writeable, we have warnings on constraints over this field (it's used for example in `sale` module for `product.template` model).

Giving an inverse to the field, we avoid this problem and give compatibility to any other module that writes this field.

A little patch on `_name_search` is also provided.

@Tecnativa

### DIFF
--- a/base_multi_company/models/multi_company_abstract.py
+++ b/base_multi_company/models/multi_company_abstract.py
@@ -13,6 +13,7 @@ class MultiCompanyAbstract(models.AbstractModel):
         string="Company",
         comodel_name="res.company",
         compute="_compute_company_id",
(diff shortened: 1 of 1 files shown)


chienandalu (approved):

AaronHForgeFlow (approved):
:+1:

LoisRForgeFlow (approved):
LGTM :+1:

LoisRForgeFlow:
@pedrobaeza will you forwardport this to v14?

pedrobaeza:
The module is not yet migrated.

PR #317, opened 2021-09-29
[13.0][FIX] account_multicompany_easy_creation: Apply the accounting template correctly + Tests

victoralmau:
Currently, although the wizard does not give an error, it does not really do anything and does not apply the accounting plan (creation of accounting accounts, taxes, etc.)

Please @pedrobaeza and @CarlosRoca13, can you review it?

@Tecnativa TT32209

### DIFF
--- a/account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
+++ b/account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
@@ -157,8 +157,15 @@ class AccountMulticompanyEasyCreationWiz(models.TransientModel):
         self.new_company_id = self.env["res.company"].create(
             {"name": self.name, "user_ids": [(6, 0, self.user_ids.ids)]}
         )
-        self.new_company_id.chart_template_id = self.chart_template_id
-        self.new_company_id.chart_template_id.try_loading_for_current_company()
+        allowed_company_ids = (
+            self.env.context.get("allowed_company_ids", []) + self.new_company_id.ids
+        )
+        new_company = self.new_company_id.with_context(
+            allowed_company_ids=allowed_company_ids
+        )
+        self.with_context(
+            allowed_company_ids=allowed_company_ids
+        ).chart_template_id.try_loading_for_current_company(company=new_company)
         self.create_bank_journals()
         self.create_sequences()
 
--- a/account_multicompany_easy_creation/tests/__init__.py
+++ b/account_multicompany_easy_creation/tests/__init__.py
@@ -1,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from . import test_easy_creation
--- a/account_multicompany_easy_creation/tests/test_easy_creation.py
+++ b/account_multicompany_easy_creation/tests/test_easy_creation.py
@@ -1,0 +1,36 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo.tests import Form, common
+
+
+class TestEasyCreation(common.SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.user = cls.env.ref("base.user_admin")
+        cls.chart_template_id = cls.env.ref(
+            "l10n_generic_coa.configurable_chart_template"
+        )
+
+    def _test_model_items(self, model, company_id):
+        self.assertGreaterEqual(
+            self.env[model].search_count([("company_id", "=", company_id.id)]), 1
+        )
+
+    def test_wizard_easy_creation(self):
+        wizard_form = Form(
+            self.env["account.multicompany.easy.creation.wiz"].with_user(self.user)
+        )
+        wizard_form.name = "test_company"
+        wizard_form.chart_template_id = self.chart_template_id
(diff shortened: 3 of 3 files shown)


account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -157,8 +157,19 @@ def create_company(self):
...
+        self.env.user.write(
+            {
+                "company_ids": [(4, self.new_company_id.id)],
+                "company_id": self.new_company_id.id,
+            }
+        )
+        ctx = dict(self.env.context)
+        allowed_company_ids = ctx.get("allowed_company_ids") or []
+        allowed_company_ids.append(self.new_company_id.id)
+        ctx.update(allowed_company_ids=allowed_company_ids)
```
pedrobaeza: Is this really needed?

victoralmau: After many tests in local I have discovered that this is necessary to avoid reading errors in all the models in which you try to create records by the access rules.

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -157,8 +157,19 @@ def create_company(self):
...
+        self.with_context(
+            context=ctx
```
pedrobaeza: OK, then let's favor the readability:
```suggestion
        self.with_context(allowed_company_ids=(
            self.env.context.get("allowed_company_ids", []) + 
            self.new_company_id.ids
        )
```

victoralmau: Thanks, but although this change works in the tests it does not work in UI, perhaps because the context already has `allowed_company_ids` defined and that is why the entire context needs to be redefined.

pedrobaeza: There should be another thing then, as `with_context` replace the corresponding key, no matter if it exists previously or not.

victoralmau: https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L5192

pedrobaeza: The code you mention is not wasting my proposal. It's designed for always including `allowed_company_ids` in the context, although you replace the whole dictionary and remove it, but it allows to replace it by another. Look this shell extract:

Selección_001

victoralmau: Ok, I have changed it according to your suggestion even though it doesn't work in UI.

pedrobaeza: This is not what I want. What I want is that you find the real root of the problem, which is not what you were saying it is.

victoralmau: Sorry for the confusion, it is already fixed. It is also necessary to specify `allowed_company_ids` on the company that we have previously created.

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -157,8 +157,21 @@ def create_company(self):
...
+        allowed_company_ids = (
+            self.env.context.get("allowed_company_ids", []) + self.new_company_id.ids
+        )
+        new_company_id = self.new_company_id.with_context(
+            allowed_company_ids=allowed_company_ids
+        )
+        self.env.user.write(
+            {
+                "company_ids": [(4, self.new_company_id.id)],
+                "company_id": self.new_company_id.id,
```
pedrobaeza: This change is too invasive. I don't get why this is needed and in other versions it's not. Maybe you can check the current 14.0 migration PR to see if there's the same problem.

victoralmau: True, this is not really necessary.
In the PR of 14.0 this problem does not exist because as this in v13 it does not work (it creates the company but not the data of the accounting template).

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -157,8 +157,15 @@ def create_company(self):
         self.new_company_id = self.env["res.company"].create(
             {"name": self.name, "user_ids": [(6, 0, self.user_ids.ids)]}
         )
-        self.new_company_id.chart_template_id = self.chart_template_id
-        self.new_company_id.chart_template_id.try_loading_for_current_company()
+        allowed_company_ids = (
+            self.env.context.get("allowed_company_ids", []) + self.new_company_id.ids
+        )
+        new_company_id = self.new_company_id.with_context(
```
pedrobaeza: This is a record, so the `_id` at the end of the variable name should go out.

pedrobaeza:
/ocabot merge patch

I don't have it clear if needed on v14 or not according your comments in https://github.com/OCA/multi-company/pull/317#discussion_r720284499, so you tell me.

victoralmau:
> I don't have it clear if needed on v14 or not according your comments in [#317 (comment)](https://github.com/OCA/multi-company/pull/317#discussion_r720284499), so you tell me.

This change is necessary also in v14.

pedrobaeza:
OK, then comment it in the PR.

PR #319, opened 2021-10-01
14.0 mig crm lead tag multi company

cvinh:
The model name changed from crm.lead.tag to crm.tag so there's a lot of renaming in that migration


cvinh:
Can you please add no stale label ?

cvinh:
Can you please add no stale label ?

cvinh:
Can you please apply 'no stale' label ?

PR #321, opened 2021-10-06
[13.0][FIX] account_multicompany_easy_creation: Fix tests if l10n es previously installed.

victoralmau:
Related to: https://github.com/OCA/multi-company/pull/317

Error in some cases: `ValueError: External ID not found in the system: l10n_generic_coa.configurable_chart_template`

**Steps to reproduce the error previously**:
- Install `l10n_es`
- Install `account_multicompany_easy_creation`
- Run the tests.

**Reasons**:
When `account_multicompany_easy_creation` is installed in the first place, `l10n_generic_coa` is self-installed and everything works correctly, but if we have previously installed `l10n_es`, `l10n_generic_coa` will no longer be installed which caused the error of not being able to load that accounting template.

Additionally, a cache problem that appeared in some cases is solved (install l10n_es + account_multicompany_easy_creation + run tests).

Please @pedrobaeza can you review it?

@Tecnativa TT32209

### DIFF
--- a/account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
+++ b/account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
@@ -169,13 +169,12 @@ class AccountMulticompanyEasyCreationWiz(models.TransientModel):
         self.create_bank_journals()
         self.create_sequences()
 
-    @ormcache("self.id", "company_id", "match_tax_ids")
-    def taxes_by_company(self, company_id, match_tax_ids):
+    @ormcache("self.id", "company_id", "match_taxes")
+    def taxes_by_company(self, company_id, match_taxes):
         AccountTax = self.env["account.tax"].sudo()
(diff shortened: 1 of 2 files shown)


account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -169,7 +168,6 @@ def create_company(self):
         self.create_bank_journals()
         self.create_sequences()
 
-    @ormcache("self.id", "company_id", "match_tax_ids")
```
pedrobaeza, line 172: Why are you removing this? It's vital for the performance.

victoralmau: I have finally left the cache but changed something to avoid errors.

pedrobaeza (approved):

pedrobaeza (approved):

carlosdauden (approved):

PR #322, opened 2021-10-07
[FIX] account_invoice_intercompany: Force no default journal

etobella:
This change is necessary because an error is raised due to journal when creating a new invoice from the dashboard. As you can see, the journal is forced:

https://github.com/odoo/odoo/blob/12.0/addons/account/views/account_journal_dashboard_view.xml#L148
https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_journal_dashboard.py#L383

### DIFF
--- a/account_invoice_inter_company/models/account_move.py
+++ b/account_invoice_inter_company/models/account_move.py
@@ -3,7 +3,7 @@
 from odoo import _, api, fields, models
 from odoo.exceptions import AccessError, UserError
 from odoo.tests.common import Form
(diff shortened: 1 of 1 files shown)


pedrobaeza:
A more general approach is to use `clean_context` method from utils to avoid this and other possible defaults.

albariera (approved):
LGTM :+1:

rafaelbn (approved):
LGTM 👍 +1

olgamarcocb (approved):
LGTM :+1:

PR #325, opened 2021-10-13
[12.0][FIX] account_multicompany_easy_creation: match taxes by xml_id

chienandalu:
Matching taxes by description is no longer reliable

cc @Tecnativa TT32316

ping @carlosdauden @pedrobaeza  @victoralmau

### DIFF
--- a/account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
+++ b/account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
@@ -120,7 +120,7 @@ class AccountMulticompanyEasyCreationWiz(models.TransientModel):
         """ install a chart of accounts for the given company """
         user_company = self.env.user.company_id
         self.env.user.company_id = self.new_company_id
-        self.sudo().chart_template_id.try_loading_for_current_company()
+        self.chart_template_id.try_loading_for_current_company()
         self.env.user.company_id = user_company
 
     def create_bank_journals(self):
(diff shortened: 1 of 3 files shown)


account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -169,10 +169,16 @@ def create_company(self):
     def taxes_by_company(self, company_id, match_tax_ids):
```
pedrobaeza: The TODO is because `match_tax_ids` is a list, which is a non hashable type. We have solved it in later versions passing the recordset, which is hashable (and there's no problem to have a recordset as argument. The problem is to return a recordset).

```suggestion
    def taxes_by_company(self, company_id, match_taxes):
```

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -169,10 +169,16 @@ def create_company(self):
...
-            ('company_id', '=', company_id),
-        ]).ids
+        xml_ids = account_taxes.get_external_id().values()
+        # If any tax doesn't have xml, we won't be able to match it
+        if "" in xml_ids:
+            return []
+        records = []
+        for xml_id in xml_ids:
+            module, ref = xml_id.split(".", 1)
+            company, name = ref.split("_", 1)
```
pedrobaeza: As the variable is not going to be used, prepend `_` for avoiding non used variable warnings:
```suggestion
            _company, name = ref.split("_", 1)
```

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -169,10 +169,16 @@ def create_company(self):
...
+            records.append(self.env.ref("%s.%s_%s" % (module, company_id, name)).id)
```
pedrobaeza: The XML-ID may not exists:
```suggestion
            record = self.env.ref("%s.%s_%s" % (module, company_id, name), False)
            if record:
                records.append(record.id)
```

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -169,10 +169,16 @@ def create_company(self):
...
-        return AccountTax.search([
-            ('description', 'in', account_taxes.mapped('description')),
-            ('company_id', '=', company_id),
-        ]).ids
+        xml_ids = account_taxes.get_external_id().values()
+        # If any tax doesn't have xml, we won't be able to match it
+        if "" in xml_ids:
+            return []
+        records = []
+        for xml_id in xml_ids:
```
pedrobaeza: Process only existing ones:
```suggestion
        records = []
        for xml_id in xml_ids:
            if not xml_id:
                continue
```

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -186,15 +192,19 @@ def update_product_taxes(self, product, taxes_field, company_from):
         return False
 
     def match_tax(self, tax_template):
-        if not tax_template.description:
+        """We can only match the new company tax if the chart was used"""
```
pedrobaeza, line 196: When is this method used? It seems non covered by tests.

pedrobaeza (changes requested):

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
chienandalu: It's used for the default company taxes (purchase and sales)

pedrobaeza: And can't be used the same or cover it by tests?

chienandalu: Well, indeed this to choose a tax.template and match it to the new company one. I realize that there aren't tests. I'll try and take some from v13

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
chienandalu: Done

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
chienandalu: Done

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
chienandalu: Done

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
chienandalu: Done

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
chienandalu: Added tests

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
```diff
@@ -164,37 +164,48 @@ def create_company(self):
...
+            if not xml_id:
+                continue
+            module, ref = xml_id.split(".", 1)
+            _company, name = ref.split("_", 1)
+            record = self.env.ref(
+                "%s.%s_%s" % (module, company_id, name), False
+            )
+            if record:
+                records.append(record.id)
+        return records
```
pedrobaeza: This variable should be `record_ids`.

pedrobaeza (approved):

account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
chienandalu: Done

victoralmau (approved):

pedrobaeza:
/ocabot merge patch

Please move it to v13.

PR #328, opened 2021-10-22
[14.0][ADD] partner_contact_company_propagation module

Kev-Roche:
Small module allowing to propagate company_id between parents and children contacts.

### DIFF
--- a/partner_contact_company_propagation/__manifest__.py
+++ b/partner_contact_company_propagation/__manifest__.py
@@ -1,0 +1,18 @@
+# Copyright 2021 Akretion (https://www.akretion.com).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Partner Contact Company Propagation",
+    "summary": "Propagate company info to children contacts",
+    "version": "14.0.1.0.0",
+    "category": "Tools",
+    "website": "https://github.com/OCA/multi-company",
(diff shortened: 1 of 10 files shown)


Kev-Roche:
It is a small module but very helpfull, @pedrobaeza, @bealdav, @sebastienbeau what do you think about it?

pedrobaeza:
The name is not suitable, as this has nothing to do with intercompany itself. It should be named `partner_contact_company_propagation` or similar (for being inline with https://github.com/OCA/sale-workflow/tree/13.0/partner_contact_sale_info_propagation).

partner_contact_company_propagation/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2021 Akretion (https://www.akretion.com).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Partner Contact Company Propagation",
+    "summary": "Share company name between Partner's contacts",
```
pedrobaeza:
```suggestion
    "summary": "Propagate company info to children contacts",
```

pedrobaeza (approved):

ajaniszewska-dev (approved):

partner_contact_company_propagation/__manifest__.py
Kev-Roche: Indeed, it's clearer that way. Thank you

PR #330, opened 2021-11-04
[14.0][FIX] account_invoice_inter_company: Delete default_journal_id if present

omar7r:
Hi,

If you create an inter-company invoice from a journal's card in Accounting dashboard, the key "default_journal_id" with the journal clicked reaches up creation of the invoice in the other company, showing the error "Cannot create an invoice of type %(move_type)s with a journal having %(journal_type)s as type." raised by "_get_default_journal".

Regards


pedrobaeza:
The best thing is to use `from odoo.tools.misc import clean_context` method for avoiding this or any other error related to defaults.

pedrobaeza (approved):

sebastienbeau (approved):
LGTM, thanks for the fix

PR #336, opened 2021-11-28
[FIX] fix synchronisation of commercial field

sebastienbeau:
Company dependent field can be synchronized to children, if the child is a partner related to a user
we need to do the sync with sudo to avoid user error (as current user do not have access right on res.user).


sebastienbeau:
@Kev-Roche @kevinkhao @PierrickBrun

kevinkhao (approved):
code review 
LGTM

Kev-Roche (approved):
LGTM

PR #342, opened 2022-02-08
[FIX] fix editing bank_ids fields

sebastienbeau:
bank field are company dependent the object have an company_id field, so we allow to edit it
@Kev-Roche @kevinkhao @PierrickBrun


kevinkhao (approved):

Kev-Roche (approved):
LGTM

PR #344, opened 2022-02-17
[14.0][ADD] product_category_inter_company module

Kev-Roche:
This module allows to set product categories as company dependent.

### DIFF
--- a/product_category_inter_company/models/product_category.py
+++ b/product_category_inter_company/models/product_category.py
@@ -1,0 +1,47 @@
+# Copyright 2022 Akretion (https://www.akretion.com).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class ProductCategory(models.Model):
+    _inherit = "product.category"
+    _check_company_auto = True
+
+    company_id = fields.Many2one(
+        comodel_name="res.company",
+        string="Company",
+        index=True,
+    )
+    parent_id = fields.Many2one(check_company=True)
+
+    @api.constrains("parent_id", "company_id")
+    def check_company_restriction(self):
+        for record in self:
+            if (
+                record.parent_id.company_id
+                and not record.parent_id.company_id != record.company_id
+            ):
+                raise UserError(
+                    _(
+                        "The parent category and your category %s "
+                        "must belong to the same company."
+                    )
+                    % record.name
+                )
+            if record.company_id:
+                for child in record.child_id:
+                    if record.company_id != child.company_id:
+                        if child.company_id:
+                            msg = _(
+                                "The category %s must be shared as the "
+                                "child %s belong to company %s."
+                            ) % (record.name, child.name)
+                        else:
+                            msg = _(
+                                "The category %s must be shared as the "
+                                "child %s is shared."
+                            ) % (record.name, child.name)
+                        raise UserError(msg)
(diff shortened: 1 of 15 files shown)


sebastienbeau:
@Kev-Roche you should inherit the product.template and 

- add a "check_company" on the field categ_id
- add _check_company_auto=True on the class product.template

You can grep example on odoo code

Kev-Roche:
@sebastienbeau done

flotho (approved):
code review LGTM

product_category_inter_company/models/product_category.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2022 Akretion (https://www.akretion.com).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductCategory(models.Model):
+    _inherit = "product.category"
+
```
sebastienbeau, line 12: _check_company_auto = True is missing here

product_category_inter_company/views/product_category.xml
```diff
@@ -0,0 +1,15 @@
...
+     @author Kévin Roche <kevin.roche@akretion.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="product_category_form_view" model="ir.ui.view">
+        <field name="model">product.category</field>
+        <field name="inherit_id" ref="product.product_category_form_view" />
+        <field name="arch" type="xml">
+            <field name="parent_id" position="after">
+                <field name="company_id" />
+            </field>
```
sebastienbeau, line 12: add a domain on parent_id ['|', ('company_id', '=', False), ('company_id', '=', company_id)]

Kev-Roche:
@sebastienbeau, I added your suggestions.

product_category_inter_company/models/product_category.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductCategory(models.Model):
+    _inherit = "product.category"
+    _check_company_auto = True
+
+    company_id = fields.Many2one(comodel_name="res.company", string="Company")
```
sebastienbeau: add index=True

product_category_inter_company/models/product_category.py
```diff
@@ -0,0 +1,13 @@
...
+    parent_id = fields.Many2one(check_company=True)
```
sebastienbeau, line 18: We can add a python constraint to force all child of a category that belong to a company to belong to this company too.

For example it should be not possible to have
categ A > categ B > categ C

with
Categ A without company
Categ B belong to company 1
Categ C without company

Categ C must belong to company 1 or Categ B must be without company


Something like

```

@api.constraint("parent_id", "company_id")
def check_company_restriction(self):
    for record in self:
        if record.parent_id.company_id and not record.parent_id.company_id != record.company_id:
            raise UserError("The parent category have a company so your category %s must belong to the same company" % record.name)
        if record.company_id:
            for child in records.child_ids:
                if record.company_id != child.company_id:
                    if child.company_id:
                        msg = _("The category %s must be shared as the child %s belong to company %s") % (record.name, child.name)
                    else:
                        msg =  _("The category %s must be shared as the child %s is shared") % (record.name, child.name)
                    raise UserError(msg)

sebastienbeau (changes requested):

sebastienbeau (approved):

sebastienbeau:
@cvinh it's possible for you to do a review, as you have the same need? thanks

PR #354, opened 2022-03-25
[14.0][IMP] base_multi_company: Don't set default on company_ids

rousseldenis:
As company_ids field is not required and as company_ids can be void (record
can be accessed through all companies), no default should be set.

### DIFF
--- a/base_multi_company/models/multi_company_abstract.py
+++ b/base_multi_company/models/multi_company_abstract.py
@@ -19,7 +19,6 @@ class MultiCompanyAbstract(models.AbstractModel):
     company_ids = fields.Many2many(
         string="Companies",
         comodel_name="res.company",
-        default=lambda self: self._default_company_ids(),
     )
     # TODO: Remove it following https://github.com/odoo/odoo/pull/81344
     no_company_ids = fields.Boolean(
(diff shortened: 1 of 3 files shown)


rousseldenis:
If record is created through a process which is not aware of OCA multi-company and intended to be shared across all companies
`record.create({
    "company_id': False,
})`

The default value will fill company_ids with the user current company. So, when evaluating 'company_id' field, we will have a record with the current one.

This diverges from wanted behaviour: we really want a void record for company_ids AND comapny_id

pedrobaeza (approved):
Agreed when `company_id` is not specified in the vals. For `company_id = False`, the inverse should act.

matteoopenf (approved):

pedrobaeza:
Can you rebase to check the possible problem?

francesco-ooops:
@rousseldenis can you rebase?

pedrobaeza:
@rousseldenis can you check the error in the test?

acsonefho:
@rousseldenis Need rebase

acsonefho (approved):
Code LGTM and already in production

rousseldenis:
@acsonefho @pedrobaeza I've fixed tests in product_multi_company too due to this change

PR #358, opened 2022-04-21
[14.0][FIX] purchase_sale_inter_company: fix new sale date_order

Kev-Roche:
The sale order_date actually is the purchase order date creation. This fix puts the sale order date as the purchase validation date (date_approve of po)


flotho (commented):
Hi, 
Thanks for your PR,
could you precise why you don't explicitly use the validation date of the purchase ?
Regards

Kev-Roche:
@flotho 
I did not use validation date of the po because, I did not think about it...
But your idea is better, cleaner than the today date, I applied it. 
Best

sebastienbeau (approved):
LGTM code review

flotho (approved):
thanks

PR #363, opened 2022-06-11
[14.0] account_invoice_inter_company: fix check product

sebastienbeau:
We need to check the access right of the product not the line
if the dest_user do not have access to the invoice line you have an error


sebastienbeau:
@PierrickBrun @Kev-Roche

sebastienbeau:
@bealdav

bealdav (approved):

Kev-Roche (approved):

PR #387, opened 2022-09-28
[FIX] [product_supplierinfo_intercompany] same company as buyer and seller

Kev-Roche:
If a buyer company is also a seller in another company for the same product, this company can be selected as seller for itself. 
This PR fixe it.


sebastienbeau (commented):
LGTM (code review), but can you add a test that fail before your commit? Thanks

sebastienbeau:
Replace by https://github.com/OCA/multi-company/pull/423

PR #391, opened 2022-10-14
[14.0][IMP] purchase_sale_inter_company: the sale order is updated when the purchase order is modified

chafique-delli:
**Features added to the module:**
If an inter-company purchase order is validated, its modification is prevented.
If you want to modify a purchase order, you have to cancel it first, then put it back in draft so that you can modify it and finally revalidate it.
When you revalidate the purchase order after modification, it canceled the sales order and recreates one.


chafique-delli:
@sebastienbeau

purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -3,13 +3,20 @@
...
 
-from odoo import _, models
+from odoo import _, fields, models
 from odoo.exceptions import UserError
 
 
 class PurchaseOrder(models.Model):
     _inherit = "purchase.order"
 
+    intercompany_sale_id = fields.Many2one(
```
sebastienbeau: There is already a field on sale order that point on the purchase order https://github.com/OCA/multi-company/blob/20eba83c19a33009fb3c9e01a99ef32bd7b20d7f/purchase_sale_inter_company/models/sale_order.py#L12

It will be better to create an inverse field

```
Even if it's a One2many, it's in reality a One2one
intercompany_sale_id = fields.One2many(
    "sale.order",
    "auto_purchase_id",
    readonly=True,
)
```

Then in "sale.order" add

```
_sql_constraints = [
(                                                
    "auto_purchase_id_uniq",                              
    "unique (auto_purchase_id)",                       
    "The auto_purchase_id must be unique !",   
)                                                
]
```

chafique-delli:
up @sebastienbeau

francesco-ooops:
@chafique-delli can you rebase in order to trigger runbot rebild?

chafique-delli:
> @chafique-delli can you rebase in order to trigger runbot rebild?

@francesco-ooops , it's done.

elvise:
@chafique-delli can you rebase ?

francesco-ooops:
@chafique-delli did you implement this successfully in the end?

sebastienbeau (approved):

sebastienbeau:
@francesco-ooops @elvise @Kev-Roche 
it's ready, I just rebase it due to change in base module

purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -81,16 +88,28 @@ def _inter_company_create_sale_order(self, dest_company):
...
             )
+        # if a sale order has already been generated
+        # delete it and force the same number
+        force_number = False
+        if self.intercompany_sale_id and self.intercompany_sale_id.state in [
+            "draft",
+            "cancel",
+        ]:
+            force_number = self.intercompany_sale_id.name
+            self.intercompany_sale_id.with_context(force_delete=True).unlink()
```
sebastienbeau: I am not sure about this.
If a sale order already have a picking and you cancel it. I pretty sure you can not delete the sale.

Maybe we should add a try except and if we can unlink it, we can keep the same number. If not we should use a new number as sale name should be uniq

purchase_sale_inter_company/models/purchase_order.py
sebastienbeau: I set this comment as resolve, because a PO can have several sale order (you can have several canceled so).
But I think it will be better to set the field "intercompany_sale_id" as computed (with compute_sudo and take the last created) to avoid storing a duplicated information.

sebastienbeau (changes requested):

purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -175,14 +194,23 @@ def _prepare_sale_order_line_data(self, purchase_line, dest_company, sale_order)
...
+                    )
+                    % (
+                        self.name,
+                        self.intercompany_sale_id.name,
+                        self.partner_id.name,
+                        self.intercompany_sale_id.state,
+                    )
+                )
+            self.intercompany_sale_id.sudo().action_cancel()
+            self.write({"partner_ref": False})
```
HekkiMelody: **question**: do you not need `sudo` to read these fields `self.intercompany_sale.id.state`, `self.intercompany_sale_id.name` ?

HekkiMelody: **question**: why delete the `partner_ref`? The PO is still the same even if it was cancelled

HekkiMelody (commented):
Some minor issues from code review.

Most importantly, from a functional POV: I agree with canceling the SO if the PO is canceled. I'm not so sure about having a new SO with the same name if the PO is then re-confirmed. It could be misleading if the PO has changed, but the user on the SO side thinks they're looking at the same one as before.

In general, also, I'm against deleting things (as opposed to canceling/archiving). Preserving the history of what happened is important. There can be attachments, comments, or edits on the SO that would suddenly be deleted as well. If there's no issue with making a new SO, I would make a new SO instead.

purchase_sale_inter_company/views/purchase_view.xml
```diff
@@ -0,0 +1,17 @@
...
+    <record id="purchase_order_form" model="ir.ui.view">
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form" />
+        <field name="arch" type="xml">
+            <field name="partner_id" position="after">
+                <field name="intercompany_sale_id" invisible="True" />
+            </field>
+            <field name="order_line" position="attributes">
+                <attribute name="attrs" operation="python_dict" key="readonly">
+                    ['|', ('state', 'in', ('done', 'cancel')), ('intercompany_sale_id', '!=', False), ('state', '=', 'purchase')]
```
HekkiMelody: **issue**: I'm not 100% sure about the order of operations, so, for readability if nothing else, I'd change this to:

```suggestion
                    ['|', 
                            ('state', 'in', ('done', 'cancel')),
                             '&nbsp;' 
                            ('intercompany_sale_id', '!=', False), 
                            ('state', '=', 'purchase')]
```

Translated: Always make it readonly during states `done` and `cancel`. If there is an `intercompany_sale_id`, make it readonly during `purchase` as well.

francesco-ooops:
@chafique-delli do you plan on working further on this one?

chafique-delli:
> do you plan on working further on this one

Yes, from March, I will have more time to think about the logic of canceling the PO as suggested by @aleuffre

purchase_sale_inter_company/models/purchase_order.py
Kev-Roche: Indeed, I think we need the sudo here.

chafique-delli: @Kev-Roche , it's fixed.

francesco-ooops:
@chafique-delli let us know when ready for review :)

francesco-ooops:
@chafique-delli do you think this will be completed eventually?

PR #393, opened 2022-10-17
[16.0][MIG] res_company_code (from 14.0 to 16.0)

legalsylvain:
Note : please review trivial dependency that makes CI red : https://github.com/OCA/multi-company/pull/390

### Description

In a big multi-company context, users often use a code to indentify
the differents companies.

* This module adds ``code`` fields on company model.

res_company_form

This field is optional, but should be unique, if defined.

* This code will prefix the name, in the Odoo Switch Company Menu widget:

switch_company_menu

* this module also change the order of the companies, ordering by code, then by company name. (instead of by sequence, then by name). It so hides the now useless field sequence.

res_company_tree


luisg123v:
Hi @legalsylvain,

We need this module on v15. Are you planning to migrate it to v15?

I could do so, but then I think it should go first so this one is based on the 15.0 branch.

What do you think?

CC @hugho-ad

legalsylvain:
Hi @luisg123v. No planned to migrate this module in 15.0, ou can do it. And you can base your work on 14.0 or on 16.0 branch. (as you wish). (branch 16 doesn't have to be based on 15.0, no.)

For the time being, could you review this trivial PR and the according #390 ? 

Thanks in advance !

res_company_code/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (C) 2019 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Company Code",
+    "summary": "Add 'code' field on company model",
+    "version": "16.0.1.0.1",
```
luisg123v, line 8:
```suggestion
    "version": "16.0.1.0.0",
```

res_company_code/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "category": "Tools",
+    "author": "GRAP, Odoo Community Association (OCA)",
+    "maintainers": ["legalsylvain"],
```
luisg123v, line 11: Could you also add myself, please?

Add a trailing comma so its formatted one user per line

res_company_code/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "website": "https://github.com/OCA/multi-company",
+    "license": "AGPL-3",
+    "depends": [
+        "web",
+        "res_company_search_view",
+    ],
+    "data": [
+        "views/view_res_company.xml",
```
luisg123v, line 19:
```suggestion
        "views/res_company_views.xml",
```

res_company_code/models/ir_http.py
```diff
@@ -0,0 +1,36 @@
...
+        result = super().session_info()
+        user = request.env.user
+        display_switch_company_menu = (
+            user.has_group("base.group_multi_company") and len(user.company_ids) > 1
+        )
+
+        # 1. Replace company name by company complete name in the session
+        #    The values are used in the switch_company_menu widget (web module)
+        # 2. Recompute sequence. (as the widget hardcod the order by sequence). See :
+        #    https://github.com/odoo/odoo/blob/16.0/addons/web/static/src/webclient/
```
luisg123v, line 23: - Don't break URLs
- Use a permalink (i.e. including commit instead of branch) otherwise it could get broken if new changes are introduced, because the line you're referring to may have changed its position

res_company_code/models/ir_http.py
```diff
@@ -0,0 +1,36 @@
...
+        #    switch_company_menu/switch_company_menu.xml#L10
+        if display_switch_company_menu:
+            sequence = 0
+            for company_id in user.company_ids.ids:
+                company = user.company_ids.filtered(lambda x: x.id == company_id)
```
luisg123v, line 28: This makes no sense, you could just iterate using:

`for company in user.company_ids:`

That way, company will have the record.

Even if you got the ID, you could just use `browse(company_id)`, filtering by ID is inefficient

res_company_code/models/ir_http.py
```diff
@@ -0,0 +1,36 @@
...
+                sequence += 1
```
luisg123v, line 29: If you iterate using `enumerate(user.company_ids)`, you wouldn't need to assign a manual sequence. Besides, companies already have sequences. Can't those be used instead?

res_company_code/models/res_company.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (C) 2019 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+    _order = "code,name"
```
luisg123v, line 10: Change not necessary AFAIK

res_company_code/tests/test_res_company_code.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2021 Akretion
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
```
luisg123v, line 4:
```suggestion
from odoo.tests import TransactionCase
```

res_company_code/views/view_res_company.xml
```diff
@@ -0,0 +1,71 @@
...
+        <field name="model">res.company</field>
+        <field name="inherit_id" ref="base.view_company_tree" />
+        <field name="arch" type="xml">
+
+            <field name="sequence" position="attributes">
+                <attribute name="invisible">1</attribute>
+                <!-- In odoo V16, if a field has a widget='handle'
+                it change the order of the display of the tree view...
+                We so remove the widget value -->
+                <attribute name="widget" eval="False" />
```
luisg123v, line 31: Are you sure this is necessary? This is not new AFAIK

res_company_code/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "summary": "Add 'code' field on company model",
+    "version": "16.0.1.0.1",
+    "category": "Tools",
+    "author": "GRAP, Odoo Community Association (OCA)",
+    "maintainers": ["legalsylvain"],
+    "website": "https://github.com/OCA/multi-company",
+    "license": "AGPL-3",
+    "depends": [
+        "web",
+        "res_company_search_view",
```
luisg123v, line 16: I think this is not expected, why would you need an extra module to add a company code?

It should be on a different module IMHO

luisg123v (changes requested):
CC @JesusValdez96

luisg123v:
Title: `[MIG] module_name: migrate to 16.0`

PR #408, opened 2022-11-17
[IMP] account_invoice_inter_company: 10 times faster counterpart invo…

ilyasProgrammer:
…ice creation.

### DIFF
--- a/account_invoice_inter_company/models/account_move.py
+++ b/account_invoice_inter_company/models/account_move.py
@@ -153,7 +153,6 @@ class AccountMove(models.Model):
             dest_invoice_data["name"] = force_number
         dest_invoice = self.create(dest_invoice_data)
         # create invoice lines
-        dest_move_line_data = []
         form = Form(
             dest_invoice.with_company(dest_company.id),
             "account_invoice_inter_company.view_move_form",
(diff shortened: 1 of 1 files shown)


francesco-ooops (approved):
Tested locally, time significantly decreased!

image

@OCA/intercompany-maintainers can you take a look?

pedrobaeza (changes requested):
Speed here is not the only parameter. We want to do it through Form to get all the needed values, including extension modules that you can't handle at this level.

elvise:
> Speed here is not the only parameter. We want to do it through Form to get all the needed values, including extension modules that you can't handle at this level.

@pedrobaeza Got it, any suggestion ? 
Right now for 15 invoices it took around 12 hours and that's not reasonable.

pedrobaeza:
It seems something incorrect indeed, as the same `Form` is used for all the invoice lines. I don't think it's something of this base.

ilyasProgrammer:
@pedrobaeza can play_onchanges() help here instead of Form ?

elvise:
> It seems something incorrect indeed, as the same Form is used for all the invoice lines. I don't think it's something of this base.

@pedrobaeza are you assuming that the problem is not with this module?

pedrobaeza:
Not easily, as you don't know all the possible ramifications. `onchange_helper` does it, but if you get that degraded performance, it's due to these onchanges, so you will get similar results or not the whole dependency chain.

PR #416, opened 2022-12-07
[FIX] creation of company-bound category with parent that is bound to …

Rad0van:
…the same company

The condition is obviously wrong (double negated) meaning: if the parent  has company and own company and parent company are equal, throw an error...

A test would be fine but I don't currently have time for that...


Rad0van:
@Kev-Roche would you care to take a look at this? Am I right here? Thanx

PR #421, opened 2022-12-14
[13.0] account_invoice_inter_company : actually postpone tests after install

metaminux:
This is a follow-up of the PR #419, it seems in 13.0, subclasses should also be tagged


metaminux:
Hi @pedrobaeza I ran the tests in Odoo.sh for my customer using 13.0 and the changes from #419 aren't enough.

I verified locally and it seems we need to tag subclasses too...

pedrobaeza (approved):
OK, I would swear that it wasn't needed, but if you have checked it, it's not a big problem to include it.

PR #422, opened 2022-12-15
[15.0][FIX] base_multi_company: move test dependency to test-requirements

LoisRForgeFlow:
remove external dependency. it is only needed for tests.


pedrobaeza (approved):
Fast-tracking this:

/ocabot merge patch

PR #423, opened 2022-12-21
product_supplierinfo_intercompany: when using sudo seller should be filtered correctly

sebastienbeau:
@Kev-Roche @PierrickBrun 

- [x] todo add test


PierrickBrun (approved):
code LGTM, I guess this is the kind of context where a lot of people will need to use sudo

PR #427, opened 2023-01-16
[ADD] login_all_company

etobella:
With this Module, once a user logins, all the companies are selected. If he already had defined it, the original value is maintained


rafaelbn (approved):
@etobella 👍 

I would recommend another module name like user_multicompany_login_all

But it's OK

legalsylvain (approved):

PR #428, opened 2023-01-18
[12.0][ADD] product_category_company_favorite module

legalsylvain:
### Description

This module is usefull in a multi-company context if
- the companies doesn't have to use all the same categories
- you define companies to your products.

Without that module, when creating products, end user can select categories that don't make sense for the current company.
With that module, categories manager can set for each company, which category will be displayed, reducing configuration error. (and so accounting error if accounts are set at the category level.)

**Note**

- When creating new root category, it is favorite for all the existing company.

- When creating new child category, the category will inherit the configuration of the parent category in each company.

- when creating a new category, all the existing categories are favorite in the new created company.

### Configure

* Select a company you want to configure

* Go to "Inventory" > "Configuration" > "Product Categories"

* Uncheck categories you don't want to use for the current company.

product_category_tree

### Usage

* Go in "Inventory" > "Master Data" > "Products"

* Create a new product

product_form_view

End user doesn't have access to the category "All / Consumable" that is not favorite in the current company.

### Know issues

There is a little UX bug. When user display product categories in the tree view and click to select which category is favorite and which category is not, if the category has children, the value of the children is updated in the database but the display is not updated.

Refresh the page resolve the problem.


product_category_company_favorite/models/product_category.py
```diff
@@ -0,0 +1,57 @@
...
+        company_dependent=True,
+        help="If this field is unchecked, the category will"
+        " be hidden when searching category in a drop-down list"
+        " like in the product form view."
+    )
+
+    @api.onchange("parent_id")
+    def onchange_parent_id_favorite(self):
+        if self.parent_id:
+            self.is_favorite = self.parent_id.is_favorite
```
carmenbianca, line 22: This seems like it violates the principle of least astonishment to me. The user may not expect the value of is_favorite to change when changing an unrelated value.

product_category_company_favorite/models/product_category.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @api.model
+    def create(self, vals):
+        category = super().create(vals)
+        # Set the new category as a favorite for all the other companies
```
carmenbianca: Comment is incorrect.

product_category_company_favorite/models/product_category.py
```diff
@@ -0,0 +1,57 @@
...
+        company_ids = self.env["res.company"].with_context(
+            active_test=False
+        ).search([("id", "!=", self.env.user.company_id.id)]).ids
+        for company_id in company_ids:
+            contextual_category = category.with_context(force_company=company_id)
+            contextual_category.is_favorite = (
+                not contextual_category.parent_id
+                and True
+                or contextual_category.is_favorite
+            )
```
carmenbianca: Hard to read.

```suggestion
            contextual_category.is_favorite = (
                True if not contextual_category.parent_id
                else contextual_category.is_favorite
            )
```

I'm not really sure what `else contextual_category.is_favorite` means here, though. I guess it defaults to False? Why not be explicit?

```suggestion
            contextual_category.is_favorite = not bool(
                contextual_category.parent_id
            )
```

product_category_company_favorite/models/product_category.py
```diff
@@ -0,0 +1,57 @@
...
+            super(ProductCategory, children).write(
+                {"is_favorite": vals["is_favorite"]}
+            )
+        return super().write(vals)
+
+    def _name_search(
+            self, name, args=None, operator='ilike', limit=100, name_get_uid=None):
+        # if self.env.user.company_id.use_favorite_product_categories:
+        args = list(args or [])
+        args += [("is_favorite", "=", True)]
```
carmenbianca: Should first check whether `is_favorite` is already in the domain. If it is, don't add it.

product_category_company_favorite/models/product_category.py
carmenbianca: On second inspection, maybe I'm wrong. All writes to this field are recursive. Maybe it makes sense.

product_category_company_favorite/tests/test_module.py
```diff
@@ -0,0 +1,63 @@
...
+        self.assertFalse(self.category_A_2_x.is_favorite)
+
+    def test_20_create_new_category(self):
+        root_categ_vals = {
+            "name": "New Root Category",
+        }
+        new_categ = self.ProductCategory.create(root_categ_vals)
+
+        self.assertFalse(new_categ.is_favorite)
+        self.assertTrue(self._change_company(new_categ).is_favorite)
```
carmenbianca: This makes no sense to me. Why is it not favourite for main_company, but it _is_ for demo_company?

carmenbianca (changes requested):
Mostly LGTM. Really interesting module. Some remarks.

product_category_company_favorite/models/product_category.py
legalsylvain: well, if the parent change, and the new one is not favorite, it makes no sense if the child is favorite (by default). that is the sense of this onchange feature. 
when you create a new category.
- step one, click on create parent is undefined.
- step two, select a parent. the onchange is raised.
-> we can guess that the "expected" default setting of is_favorite is to inherit the configuration of the parent.

makes sense for me.

product_category_company_favorite/models/product_category.py
legalsylvain: I think I mean or ``contextual_category.parent_id.is_favorite`` to make the feature similar to the onchange.

If we create a new root category, by default it is available in other companies.
If we create a new child category, by default, it inherit from the configuration of it parent in other companies.

does it make sense for you ?

product_category_company_favorite/tests/test_module.py
legalsylvain: you're right, if no "is_favorite" value is provided, it should be true by default.

product_category_company_favorite/models/product_category.py
legalsylvain: > I'm not really sure what else contextual_category.is_favorite means here, though. I guess it defaults to False? Why not be explicit?

Reading again : 
- you're right It makes no sense. 
- I changed the algorithm. 
  - if it is a child category -> inheritance : OK
  - if it is a new root category : same configuration as the current company

So it allow to create "opt-in" categories, or "opt-out" categories, depending on the user.

legalsylvain:
Hi @carmenbianca : could you update your review ? 
thanks !

product_category_company_favorite/models/product_category.py
```diff
@@ -0,0 +1,61 @@
...
+                {"is_favorite": vals["is_favorite"]}
+            )
+        return super().write(vals)
+
+    def _name_search(
+            self, name, args=None, operator='ilike', limit=100, name_get_uid=None):
+        args = list(args or [])
+        if (
+            not args
+            or not any([x[0] == "is_favorite" for x in args if isinstance(x, tuple)])
```
carmenbianca: `isinstance(x, tuple)` is a little too naïve. Sometimes Odoo uses lists instead of                                                                                                              tuples for domains.

Maybe let's do this:

```suggestion
            or not any([x[0] == "is_favorite" for x in args if getattr(x, "__getitem__", False)])
```

carmenbianca (changes requested):
Very small nitpick. Mostly LGTM to me :+1:

product_category_company_favorite/models/product_category.py
legalsylvain: Thanks ! 
Updated!

carmenbianca (approved):
LGTM, thank you

product_category_company_favorite/models/product_category.py
fcayre: Makes sense for me as well.

product_category_company_favorite/models/product_category.py
```diff
@@ -0,0 +1,62 @@
...
+        category = super().create(vals)
+        # Configure is_favorite for all the other companies
+        company_ids = self.env["res.company"].with_context(
+            active_test=False
+        ).search([("id", "!=", self.env.user.company_id.id)]).ids
+        for company_id in company_ids:
+            ctx_category = category.with_context(force_company=company_id)
+            if ctx_category.parent_id:
+                # We inherit the contextual configuration of the parent category, if any
+                ctx_category.is_favorite = ctx_category.parent_id.is_favorite
```
fcayre, line 35: As stated by codecov, I think a quite important test is missing here: creation of a category that is a child of favorite=True category which should set the created child category favorite attribute to true.

product_category_company_favorite/models/product_category.py
```diff
@@ -0,0 +1,62 @@
...
+        company_dependent=True,
+        help="If this field is unchecked, the category will"
+        " be hidden when searching category in a drop-down list"
+        " like in the product form view."
+    )
+
+    @api.onchange("parent_id")
+    def onchange_parent_id_favorite(self):
+        if self.parent_id:
+            self.is_favorite = self.parent_id.is_favorite
```
fcayre, line 22: Less important than the other for me, but as seen by codecov, a small test is missing here.

fcayre (approved):
Interesting module, thanks very much for contributing them.
Maybe not blocking but 2 unit tests seems to be missing here.

product_category_company_favorite/models/product_category.py
legalsylvain: Done in ``test_40_test_onchange``

product_category_company_favorite/models/product_category.py
legalsylvain: improved ``test_20_create_new_category_not_favorite`` and ``test_21_create_new_category_favorite``

legalsylvain:
Hi @fcayre. thanks for your review. i added test.

legalsylvain:
Hi PSC team. Could you merge this approved PR ? CC  : @jgrandguillaume, @JordiBForgeFlow, @gurneyalex 

Note : there are 8 PRs approved that are ready to merge : 
https://github.com/OCA/multi-company/pulls?q=is%3Apr+is%3Aopen+label%3Aapproved

Thanks in advance !

legalsylvain:
No news since monthes from the PSC team.

I'll try the @pedrobaeza wildcard ;-)

pedrobaeza:
/ocabot merge nobump

2 remaining wildcards ;p

legalsylvain:
Thanks Pedro ! 

> 2 remaining wildcards ;p

Got damn'it ! My stock is almost gone.
How do I get more? If it's a matter of paying for a few beers, let me know ! OCA days are coming ;-)

PR #432, opened 2023-01-25
[16.0] [MIG] product_tax_multicompany_default

Shide:
Straightforward migration

Things changed:
- Groups id of view on field
- Performing an SQL to remove all taxes from other companies (cannot do it with sudo or SUPERUSER_ID) (done in a separate Commit to be able to cherry-pick)

MT-1564 @moduon @rafaelbn @carlosdauden @cubells


yajo (approved):
Code OK!

loida-vm (approved):
If in most of the invoices the tax is X, this is the one that will be suggested in the product, when we are going to create a new one.
If in a company in the database, it does not have any invoice yet, when we are going to create the first product, it does not suggest any tax, because it does not recognize which one is the most used. 

The button in the product view to set taxes in other companies, only appears if there is more than one company in the DB.

LGTM! 👌

rafaelbn:
Hello @carlosdauden , do you agree fixing this BUG?

https://www.loom.com/share/0f50c6ee2dd04c0f80390a2bd94362e9

@Shide we should FIX this BUG in 15 ASAP please 🙏 

Thank you!

MT-2121 @moduon

carlosdauden:
:+1:

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,120 @@
...
+        excl_supplier_tax_ids: List[int] = None,
+    ):
+        """Delete taxes from product excluding chosen taxes
+
+        :param excl_customer_tax_ids: Excluded customer tax ids
+        :param excl_supplier_tax_ids: Excluded supplier tax ids
+        """
+        tax_where = " AND tax_id NOT IN %s"
+        # Delete customer taxes
+        customer_sql = "DELETE FROM product_taxes_rel WHERE prod_id IN %s"
```
carlosdauden, line 42: Why use SQL?
Set customer_tax_ids  and supplier_tax_ids as you like before than this code:
```
        self.write(
            {
                "taxes_id": [(6, 0, customer_tax_ids)],
                "supplier_taxes_id": [(6, 0, supplier_tax_ids)],
            }
        )

```

carlosdauden (changes requested):

product_tax_multicompany_default/models/product.py
Shide: Cannot do it with sudo or SUPERUSER_ID, don't know why, but I've tried a lot of possibilities with sudo, with_user(SUPERUSER_ID), with_context(allowed_company_ids=...), with_company(), and nothing...

Systematically is excluing all the [(5, None)] on other companies. 
This is the reason why I've chosen to empty vía sql

loida-vm (approved):
Functional review :+1:

fcvalgar (approved):
LGTM 
Thank you @Shide

carlosdauden (approved):
I'm not very convinced, but I don't want to block :wink:

PR #454, opened 2023-03-07
[14.0][FIX] `partner_multi_company` removes partner_share from res.partner ir.rule

DaniilDigtyar:
## Module `base_multi_company` and `partner_multi_company`
The bug is caused by the hook on `base_multi_company` and called by installing  `partner_multi_company`.
 
Odoo uses `partner_share` to share all the partner of internal users, so it can use them when there are relations to records. The hook of `base_multi_company` remove it from `base.res_partner_rule`

### `base.res_partner_rule` domain before and after installing `partner_multi_company`:
After: `['|', '|', ('partner_share', '=', False), ('company_id', 'in', company_ids), ('company_id', '=', False)] `
Before: `['|', ('no_company_ids', '=', True), ('company_ids', 'in', company_ids)]`

## Describe the bug
When a user A interacts with a record that is has interaction from another user B using `mail.activity` or another relation to the user, it causes an access error to the information of the partner related user if User A doesn't have access to User B company.

## To Reproduce
Tested on **14.0**

1. Create user A with only access to company A
2. Create user B with only access to company B
3. Create a record (res.partner for example) and give access to both companies (A and B)
4. User A creates a message or mail.activity on shared record.
5. User B when accessing shared record has an access error. (Can go inside the record but will not load the User A relations)

**Expected behavior**
When User B enters, the shared record should load without access error.

**Proposed solution**
Modify the `post_init_hook` and `unisstall_hook` of `partner_multi_company` to contemplate the `partner_share` field in the domain.

Add `partner_share` in domain of `post_init` and `unistall` hook of `partner_multi_company`

### DIFF
--- a/partner_multi_company/hooks.py
+++ b/partner_multi_company/hooks.py
@@ -36,9 +36,8 @@ def uninstall_hook(cr, registry):
             {
                 "active": False,
                 "domain_force": (
-                    "['|','|',('company_id.child_ids','child_of',"
-                    "[user.company_id.id]),('company_id','child_of',"
-                    "[user.company_id.id]),('company_id','=',False)]"
+                    "['|', '|', ('partner_share', '=', False), ('company_id', 'in', company_ids), "
+                    "('company_id', '=', False)]"
                 ),
             }
         )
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I have copied the issue information here. It's not needed to open 2 things. I'm closing the issue now, and following here.

PR #457, opened 2023-03-13
Migración de version de la v15 a la v16

DMopen2s:
Migracion de la v15 a la v16 de los modulos : purchase_sale_inter_company, purchase_sale_stock_inter_company


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0.

Please check also red CIs for fixing them.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #458, opened 2023-03-14
[16.0][MIG] base_multi_company

atchuthan:
cc: @SodexisTeam

PR #459, opened 2023-03-14
[16.0][MIG] product_multi_company

atchuthan:
cc: @SodexisTeam


amkarthik (approved):

product_multi_company/models/product_product.py
```diff
@@ -0,0 +1,49 @@
...
+            return super(ProductProduct, self).write(values)
+        if values.get("company_ids") and values["company_ids"][0][0] == 6:
+            # Add companies to the template, but don't remove them.
+            companies_to_add = list(
+                set(values.get("company_ids")[0][2]).difference(self.company_ids.ids)
+            )
+            companies_to_add.extend(self.product_tmpl_id.company_ids.ids)
+            if companies_to_add:
+                self.product_tmpl_id.write({"company_ids": [(6, 0, companies_to_add)]})
+        return super(ProductProduct, self).write(values)
```
yajo: All this code doesn't exist in the previous branch. Actually, there was no `product_product.py` file. Only a small `product_template.py`:

https://github.com/OCA/multi-company/tree/ae4570d4f327ce5e0ad3779bbcde3369b82c3acb/product_multi_company/models

I observe also that in the commit 9633be1b29f0a04eba9fa5fa3fc0efd7853f006e, labeled `[MIG] product_multi_company: black, isort, prettier`, you're not actually just passing the automated pre-commit reformats. You're refactoring the module completely.

Most of the logic you add there should be unnecessary at first sight, as it should exist already in the inherited `multi.company.abstract` model.

I'm confused. Why did you have to refactor the module? Didn't the abstract model work? Why did you hide these big changes in the reformat commit?

yajo (changes requested):

amkarthik:
@yajo 
We used 13.0 branch during our migration as it contain more commits than previous.
And we tested by migrating the previous branch and it looks those changes in 13.0 is not needed any more.
We will contribute it ASAP

Thanks!

yajo:
At this point, it will probably be easier to just restart migrating from 15.0.

amkarthik:
@yajo 

We are finished with the work. Please check again.

Thanks!

product_multi_company/models/product_product.py
amkarthik: @yajo 

In v13, they added those extra changes to avoid the below error.
https://github.com/OCA/multi-company/issues/252
And it is still happening in the latest v15 code.

Detailed info regarding the issues is in the below link
https://github.com/OCA/multi-company/pull/264 

We are having the same issue in migrated code too.

I think we need to add those back. 

Thanks

amkarthik: @yajo 
Creating a related stored company_ids field on the product.product solves this issue.
Shall we go with this instead of having those all changes from v13?
I can't think of any disadvantages with this method.

product_multi_company/tests/test_product_multi_company.py
```diff
@@ -11,8 +11,7 @@
 class TestProductMultiCompany(ProductMultiCompanyCommon, common.TransactionCase):
     def test_create_product(self):
         product = self.env["product.product"].create({"name": "Test"})
-        company = self.env.company
-        self.assertTrue(company.id in product.company_ids.ids)
+        self.assertFalse(product.company_ids)
```
yajo: issue: wrong test fix.

This test was checking that the current company was added to `product.company_ids`. That's exactly the purpose of this module.

If you commit this change and the tests go ✔️, it means that something's not working properly in the module. Instead, probably you should undo this change and actually fix the module.

yajo (changes requested):

product_multi_company/tests/test_product_multi_company.py
amkarthik: @yajo 
We will check on it.
Have you had a chance to look into this comments?
https://github.com/OCA/multi-company/pull/459#discussion_r1365327722
https://github.com/OCA/multi-company/pull/459#discussion_r1365388411

Please review and let us know your feedback, that will be very helpful.

Thanks

amkarthik: > issue: wrong test fix.
> 
> This test was checking that the current company was added to `product.company_ids`. That's exactly the purpose of this module.
> 
> If you commit this change and the tests go ✔️, it means that something's not working properly in the module. Instead, probably you should undo this change and actually fix the module.

We have removed the default value from the company_ids fields.
Please review this link
https://github.com/OCA/multi-company/pull/458/files#diff-768c66d1106f5fdcb827fc104212e9b76a2739fae84f770ffe2687ec3bea1c4bR19

And this is to support the statement that if a product doesn't have any companies set then it is shared by all the company.

I think we just need to rename the test method name. Correct me If I misunderstood something.

Thanks

Shide:
Tests still failing

amkarthik:
@Shide 
The reason for the error is https://github.com/OCA/multi-company/pull/459#discussion_r1365327722
The suggested fix in v16 is https://github.com/OCA/multi-company/pull/459#discussion_r1365388411
I would like to have your thoughts on the fix.
cc: @yajo

micheledic (approved):

pilarvargas-tecnativa:
Superseed by #628

PR #464, opened 2023-03-21
[IMP] product_tax_multicompany_default: beautier and mass propagation

yajo:
- Propagate button was ugly. It's prettier now.

  | before | after |
  |--------|--------|
  | flameshot_2023-03-21_11-24 | flameshot_2023-03-21_11-42 |

- Add ability to filter products that have divergent cross-company taxes.
- Allow propagating massively.

@moduon MT-2586

### DIFF
--- a/product_tax_multicompany_default/models/product.py
+++ b/product_tax_multicompany_default/models/product.py
@@ -5,11 +5,57 @@
 
 from typing import List
 
-from odoo import api, models
+from odoo import api, fields, models
 
 
 class ProductTemplate(models.Model):
     _inherit = "product.template"
+
+    divergent_company_taxes = fields.Boolean(
+        string="Has divergent cross-company taxes",
+        compute="_compute_divergent_company_taxes",
+        compute_sudo=True,
+        store=True,
+        help=(
+            "Does this product have divergent cross-company taxes? "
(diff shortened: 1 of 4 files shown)


rafaelbn:
Tested in runboat, doesn't work in my test

imagen

yajo:
El mar, 21 de mar de 2023 a las 10:57:19 AM, Rafael Blasco 
***@***.***> escribió:
> Tested in runboat, doesn't work in my test
> 

The button only appears now if there are divergent taxes accross 
companies. Otherwise, it's hidden, as it would do nothing.

yajo:
Please also make sure that you have full accounting permissions. Just like before this patch, the option will not be shown unless you have that group. Remember that it is hidden unless you use Enterprise, so you have to go to the group itself and add the user, instead of doing it the other way around:

image

I added that group to `admin` in current runboat. Please try now.

fcvalgar (approved):
LGTM, thank you @yajo

rafaelbn (approved):
tested 👍 in runbot

rafaelbn:
@etobella could you please make a quick review here please?

etobella (approved):
LGTM Technically

PR #470, opened 2023-04-06
[16.0][MIG] product_category[_inter]_company

carmenbianca:
Simple porting and some tests. A name change was also requested.

Internal task: https://gestion.coopiteasy.be/web#id=10426&action=479&model=project.task&view_type=form&menu_id=536


product_category_company/models/product_category.py
```diff
@@ -0,0 +1,41 @@
...
+        index=True,
+    )
+    parent_id = fields.Many2one(check_company=True)
+
+    @api.constrains("parent_id", "company_id")
+    def check_company_restriction(self):
+        for record in self:
+            if (
+                record.parent_id.company_id
+                and record.parent_id.company_id != record.company_id
```
bealdav: Good catch @carmenbianca

bealdav (commented):
Thanks a lot.
@Kev-Roche @sebastienbeau  could you review please ?

Thanks

cathLemb:
functional review OK!

cathLemb (approved):
functional approbation

PaulGoubert (approved):

bealdav:
Additionnal requests for this PR 

image
Add company_id in tree view with dedicated group for company

image

Remove `col="2"` to a better display with company_id on the same line than parent_id
That's an idea for a usability improvement. But if you think it's bad for maintainability, maybe we can discuss other option

Thanks a lot

Kev-Roche:
> Additionnal requests for this PR
> 
> image Add company_id in tree view with dedicated group for company
> 
> image
> 
> Remove `col="2"` to a better display with company_id on the same line than parent_id That's an idea for a usability improvement. But if you think it's bad for maintainability, maybe we can discuss other option
> 
> Thanks a lot

@bealdav really good idea ! what do you think of it @carmenbianca ?

product_category_company/security/ir_rule.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" ?>
+<odoo noupdate="1">
```
carmenbianca: Are we sure `noupdate="1"` is correct here?

product_category_company/models/product_category.py
```diff
@@ -0,0 +1,41 @@
...
+
+    company_id = fields.Many2one(
+        comodel_name="res.company",
+        string="Company",
+        index=True,
+    )
+    parent_id = fields.Many2one(check_company=True)
+
+    @api.constrains("parent_id", "company_id")
+    def check_company_restriction(self):
```
carmenbianca: Is this entire function even needed? `check_company=True` is set on `parent_id`, which should do all the work for us.

product_category_company/security/ir_rule.xml
bealdav: You're right, please remove noupdate, thanks @carmenbianca

carmenbianca:
> Remove col="2" to a better display with company_id on the same line than parent_id

Removing `col="2"` doesn't achieve that result for me. I need to move the parent_id field into a div inside of the group to achieve this effect. See screenshot.

image

Is that approximately correct, or did I misinterpret?

I'm not sure moving the field into a div is quite the cleanest thing to do though.

product_category_company/models/product_category.py
bealdav: ping @Kev-Roche @sebastienbeau could you have a look here, Thanks.

Is suite tests is OK without this method ?

bealdav:
Fine for me. Excellent

product_category_company/models/product_category.py
carmenbianca: I will test this in a moment. I noticed this while busy with something else :)

carmenbianca: Removed the function in [67fb21d](https://github.com/OCA/multi-company/pull/470/commits/67fb21d5cdc65a65c74d4e0f554dd40e1f2a6e2b). Tests still run, although as a side effect, the error messages may be different.

image

carmenbianca:
Everything implemented :) Ready for re-review.

product_category_company/models/product_category.py
bealdav: No matter if message is different. Thanks

product_category_company/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import product_category
```
bealdav, line 1: product_template file is not imported here whereas inside models dir  !? weird.

carmenbianca: Weird. Fixed and added a test for this.

carmenbianca:
Should many of the fixes in this PR be backported to 14.0?

product_category_company/views/product_template_view.xml
```diff
@@ -0,0 +1,18 @@
...
+<odoo>
+    <record id="product_template_view_form" model="ir.ui.view">
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view" />
+        <field name="arch" type="xml">
+            <field name="categ_id" position="attributes">
+                <attribute
+                    name="domain"
+                >['|', ('company_id', '=', False), ('company_id', '=', company_id)]</attribute>
+
```
bealdav, line 14: Hi @carmenbianca , I have an other requirement here, sorry. It's an improvement, not a migration topic but I think it can be useful for everybody.
Maybe a test could be added:
- I'm a user without "base.group_multi_company" group (mono company user)
- when I create a product, there is no company_id in default/creating data, then company_id = False
- domain in categ_id doesn't allow me to select category with a company_id (matching with self.env.user.company_id) => it Should allow to select a category of my company.

Do you know how could we fix that ?

Please @carmenbianca 
Do you confirm this case

Thanks a lot.

bealdav: Workaround

`    company_id = fields.Many2one(
        comodel_name="res.company", default=lambda s: s.env.user.company_id.id
    )
`

It works for my customer, but not sure it's generic

carmenbianca: > when I create a product, there is no company_id in default/creating data, then company_id = False

This seems like an Odoo bug in a multi-company database to me. But I'm like 80% sure that I'm misunderstanding this, because I doubt that there would be a bug _that_ obvious in Odoo.

Your proposed workaround would work for new product records. Old records would still need to be corrected. (Although `env.user.company_id` has been changed to `env.company` since Odoo 13.)

I'll write a test first, and see what actually happens.

bealdav: Thanks for your explanation. Maybe you're right, It's probably a bug in odoo. In this case it shouldn't be resolved in this PR !?

bealdav: @legalsylvain do you an idea here as multi company intensive user ?

carmenbianca: Wrote a test in [0d021b4](https://github.com/OCA/multi-company/pull/470/commits/0d021b4b83160d64485ff5b74388551f9c388655). Your analysis was right.

I'm still doubtful that this is actually a bug, though. Surely it's intended behaviour. Maybe products are supposed to be shared between companies? And if so, then this module is not the correct place to limit that behaviour.

bealdav: Thanks a lot @carmenbianca , nice job

carmenbianca: Oops, setting a default company_id on the product template broke an unrelated test in a different module. :warning:

bealdav: Maybe you can only set this default value only for not multicompany user to avoid to break some tests ?

Call a method instead of lambda

carmenbianca: I'll await @legalsylvain 's sage input, unless there's a rush.

legalsylvain:
> I'm still doubtful that this is actually a bug, though. Surely it's intended behaviour. Maybe products are supposed to be shared between companies? 

Yes, in V16, partners and products are "global by default". That's the new design. Not a bug IMO.


> @legalsylvain do you an idea here as multi company intensive user ?

Well, i have similar needs in GRAP. 
- Size : 70 companies, and like 200 categories (shop categories ; brewery categories ; bakeries categories ; ...)
- We need to share categories between similar activities (many breweries, in the instance)
- BUT, bakeries categories are useless for brewery and displaying 200 items with 190 useless is not ergonomic, and can generate bad configuration.

I recently solved "my issue" with the module ``product_category_company_favorite``  [PR available here](https://github.com/OCA/multi-company/pull/428).
It allows to design : 
- category A : global for all companies
- category B : available for company 1 and 3
- category C : available for company 2 and 3.

1) review welcome. (work in production since 2 monthes)
2) @carmenbianca : what do you think of this alternative design? can it cover your need, or do you need to isolate all categories?

legalsylvain (approved):
That being said, :+1: LGTM.

bealdav:
OK, thanks a lot for your replies. Interesting module Sylvain, maybe one of my favorite ones for the future ?
That being said, this PR fails because Odoo new design. We should make it works.
Should we remove default override in company and suggest in Readme to set it in final project in limitations part ?
Any other suggestions ?

carmenbianca:
I removed the `check_company` stuff from product.template, envisioning more bugs, because:

All users now need to be able to read all categories, otherwise the products that have categories belonging to a different company from the user are broken.

[13a7af1](https://github.com/OCA/multi-company/pull/470/commits/13a7af164a1545641f0dc01cde8ab5cd4f9ff512) represents my best effort, but it's not perfect. On a product linked to a company, you can now set a category belonging to a different company. I could add a restraint on this, but I'm not sure how far I want to go down the rabbit hole yet.

bealdav:
Thanks @carmenbianca please add you as contributor on readme.

bealdav (approved):

product_category_company/views/product_template_view.xml
jumeldi74: > Hi @carmenbianca , I have an other requirement here, sorry. It's an improvement, not a migration topic but I think it can be useful for everybody. Maybe a test could be added:
> 
> * I'm a user without "base.group_multi_company" group (mono company user)
> * when I create a product, there is no company_id in default/creating data, then company_id = False
> * domain in categ_id doesn't allow me to select category with a company_id (matching with self.env.user.company_id) => it Should allow to select a category of my company.
> 
> Do you know how could we fix that ?
> 
> Please @carmenbianca Do you confirm this case
> 
> Thanks a lot.

@bealdav. sorry for my late comment. it's just an additional comment.  for general purpose,  in V16 already enought as mention by @legalsylvain. I mean for your case, in real business flow in a company, once the rules of system in internal management of a company use multi company, all users must be set as multi company, no user in single company.  Just my opinion.

PR #474, opened 2023-04-14
[12.0][FIX] product_intercompany_account: sudo when searching companies

MiquelRForgeFlow:
If self.env.user is not SUPERUSER_ID and the companies are independent (no one is child of the other), then normally a user of company A cannot search for company B. Thus, some flows are wrong without this fix.


GuillemCForgeFlow (approved):
👍🏽

JoanSForgeFlow (approved):
LGTM! I think in v12 this approach is justified.

AaronHForgeFlow (approved):

HviorForgeFlow:
ping @OCA/intercompany-maintainers

PR #477, opened 2023-04-28
[ADD] product_account_multicompany_default: propagate product account to other companies

yajo:
This module extends the functionality of products and accounting to allow you to propagate multi-company product accounts from the current company to all other companies where you have access rights.

@moduon MT-2626 @ArantxaSudon @loida-vm @rafaelbn


loida-vm (commented):
Functional review done!  It needs to add to the README that companies must share the same fiscal location package, otherwise the propagation will not work.

ArantxaSudon (approved):
Functional review done! Accounts assigned to a product are propagated correctly to the other companies.

rafaelbn (changes requested):
STR:

If company A has USA account_chart and one account is 600000 Pepito
If company B has Spain account_chart and one account is 600000 Menganito

Because is a coincidence that 2 account_chart in diferent countries have same account code

Now:

The account is propagated between companies and is an errror

Expected:

Only propagate accounts of the SAME account chart :-)

Detailed video 5 min: https://www.loom.com/share/1ad66d715bf548f7bb395bddc70540b5

Thank you!

yajo:
Thanks everybody for your reviews!

> Only propagate accounts of the SAME account chart :-)

In Spain there are 4 charts of accounts, and all of them should be compatible. You already noticed this in your video! Thus, that is not a valid solution. 😕 



Therefore, my proposal is:

1. Add this problem to known issues.
2. When it happens in the Real World™️, apply the new design you can see in the roadmap. That design would solve this problem as a side effect.

loida-vm:
> Therefore, my proposal is:
> 
>     1. Add this problem to known issues.
> 
>     2. When it happens in the Real Worldtm, apply the new design you can see in the roadmap. That design would solve this problem as a side effect.
>        

The accounts should indeed match as it is a chart of the same location. However, it can lead to mistakes. I think the suggested proposal is a good solution. What do you think @rafaelbn @ArantxaSudon ?

rafaelbn (approved):
Add this problem to known issues @yajo (recommend use this module only when all companies are for the same country)

Thanks! 👍🏼

yajo:
Thanks everybody. I have changed the Roadmap / Known Issues section accordingly. This should be ready to merge now then.

rafaelbn:
Hello @etobella @bguillot ! Please 🙏🏼 could you please take a look here and merge? We are running un productions with this module 😄 

Thank you! ❤️ 

Maybe @loida-vm you could make a simple video showing the functionality or screen shoot to make more easy the review 😀

loida-vm (approved):
Functional review done!
You can see the review here: https://www.loom.com/share/28243d72f0144ac29ed9c9c318a0efe8

I noticed that in this case, which I show in the video, it does propagate because the chart of accounts have an account with the same numbering. So it isn't that they should have the same chart of accounts, but whether they share numbering or not.

Thank you!

yajo:
Yes, that's the expected behavior for now. Notice that you don't need to click that button upon creation, because on creation this happens automatically; only if you change the accounts later and want to re-propagate that change.

loida-vm:
LGTM!

rafaelbn:
This is running un production por a long time, and maintainer can merge or review please? I don't know which is the group of the maintainers for multi-company @simahawk do you know?

rafaelbn:
Thank you @etobella !

PR #481, opened 2023-05-29
[ADD] [14.0] purchase_sale_lot_intercompany module

Kev-Roche:
This module allows to propagate lot numbers from a purchase to a sale line, in a multi-company context.

cc @metaminux, @sebastienbeau, @florian-dacosta


purchase_sale_lot_intercompany/models/stock_production_lot.py
```diff
@@ -0,0 +1,29 @@
...
+        self.ensure_one()
+        lot = self.sudo().search(
+            [
+                ("name", "=", self.name),
+                ("product_id", "=", self.product_id.id),
+                ("company_id", "=", company.id),
+            ]
+        )
+        if not lot:
+            lot = self.sudo().create(
```
florian-dacosta, line 22: Not blocking comment as it is an improvment, but I think it would be nice to have a hook to get these vals.
Indeed, we can imagine cases where we add additional information on lot and we would like to propagate it.

florian-dacosta (approved):
Minor  comment

PR #488, opened 2023-06-23
[15.0][FIX] account_invoice_inter_company: post dest invoice

chienandalu:
If had a src and dest invoices that were passed to drafts and the dest invoice is published before the src one, a duplicated dest invoice will be generated for the dest company.

cc @Tecnativa TT44069

ping @carlosdauden @pedrobaeza


carlosdauden (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #489, opened 2023-06-26
[15.0][FIX] account_invoice_inter_company: Avoid crash on False dest_company

pedrobaeza:
If `dest_company` is False, the expression `dest_company.id` fails, so we avoid to enter that code early if no dest_company.

Doing this also, we avoid a useless search.

@Tecnativa TT44078


pedrobaeza:
Fast-tracking for solving the clear bug that affects everyone with this installed:

/ocabot merge patch

PR #490, opened 2023-06-30
[FIX] base_multi_company: allow reinstalling submodules

yajo:
When a submodule is uninstalled, garbage records are kept with the company relationships.

When you reinstall it, duplicate keys are attempted to be written. Thus, the module cannot be reinstalled without this patch.

@moduon MT-3240

### DIFF
--- a/base_multi_company/hooks.py
+++ b/base_multi_company/hooks.py
@@ -51,6 +51,7 @@ def post_init_hook(cr, rule_ref, model_name):
             INSERT INTO {}
             ({}, {})
             SELECT id, company_id FROM {} WHERE company_id IS NOT NULL
+            ON CONFLICT DO NOTHING
         """.format(
             table_name,
             column1,


pedrobaeza:
But shouldn't the table should be removed on uninstall?

yajo:
Well, I'm not sure. The benefit with the current behavior is that, after reinstallation, you didn't lose all the multi-company links across records. So, if some submodule produces some ir.rule conflict elsewhere, you can uninstall, do what you need to do, and reinstall, all without losing data.

This is just the MVP to allow reinstalling. You can consider changing the behavior later if needed IMHO.

pedrobaeza (approved):

yajo:
This one should be ready to merge.

rafaelbn:
@etobella please could you review here? Is ready to merge

rafaelbn (approved):
👍🏼

rafaelbn:
@pedrobaeza could we have this merge in order to forward port it?

PR #491, opened 2023-06-30
[16.0][MIG] account_invoice_consolidated

MS-OSI:
Account Invoice Consolidated
============================
This module allows organizations with multiples companies to transfer the
liability of a customer to another company and provide the customer with a
unique consolidated invoice to pay.

PR #494, opened 2023-07-10
[16.0][MIG] account_move_multi_company

MS-OSI:
Depends on #491 

Account Move Multi-Company
=======================
Odoo allows a journal entry to be posted in the same company. In a multi-company environment, it allows to have inter-company transactions at the sale order and purchase order level.

Some group/holding have setup a company structure where an employee is paid by a company of the group but is working part-time for another company of the group.

When the accounting team is loading the accounting entry for the payroll, the amount dedicated to the other companies needs to be transferred to those companies.


legalsylvain:
hi @MS-OSI. Could you take a look on red CI ? thanks !

legalsylvain (commented):
Hi. If I undertand correctly, the name of the module should be ``account_move_inter_company``. 

Don't you think ? 

CC : @max3903

PR #503, opened 2023-08-21
[15.0][FIX] base_multi_company: allow reinstalling submodules

yajo:
PR(s) ported locally: #270, #490

MT-3240


yajo:
@pedrobaeza this should be fast-trackable.

PR #445, opened 2023-02-13
[IMP] purchase_sale_inter_company: apply customer specific changes and improve with latest changes from v14

yankinmax:
This PR returns customer specific changes introduced in these commits:
- [ ] 3f0510e68d64b4cd71824dc5ee33b5971f3a1733
- [ ] b8d006c30cf94c8be73fb9caef169a77ae955019
- [ ] 628f3c20f5a9b3b017c819fb17476aa8bd4726af

PR #504, opened 2023-08-28
[IMP] purchase_sale_inter_company: apply customer specific changes and improve with latest changes from v14

yankinmax:
This PR returns customer specific changes introduced in these commits:
- [ ] 3f0510e68d64b4cd71824dc5ee33b5971f3a1733
- [ ] b8d006c30cf94c8be73fb9caef169a77ae955019
- [ ] 628f3c20f5a9b3b017c819fb17476aa8bd4726af

This PR is opened as a replacement of the closed one: https://github.com/OCA/multi-company/pull/445

### DIFF
--- a/purchase_sale_inter_company/models/purchase_order.py
+++ b/purchase_sale_inter_company/models/purchase_order.py
@@ -43,18 +43,25 @@ class PurchaseOrder(models.Model):
             dest_company = (
                 po_partner.ref_company_ids or po_partner.parent_id.ref_company_ids
             )
+            if len(dest_company) > 1:
+                raise UserError(
+                    _("You can generate sale order only for one vendor company at once")
+                )
             intercompany_user = dest_company.intercompany_sale_user_id
             if not intercompany_user:
                 intercompany_user = self.env.user
             if dest_company and dest_company.so_from_po:
                 # one sale order is expected to be created when confirm purchase order
                 # so we check for existing sale order with auto_purchase_order_id set
+                # and confirm that sale order if found (it's possible it can be draft)
                 # we don't need extra sale orders if purchase order was confirmed again
                 auto_sale_order_id = (
                     self.env["sale.order"]
                     .sudo()
                     .search([("auto_purchase_order_id", "=", purchase_order.id)])
                 )
+                # standard OCA flow raises an error if we try to confirm purchase order
+                # with already generated auto sale order, comment it for now
                 # if len(auto_sale_order_id) > 1:
                 #     raise UserError(
                 #         _(
@@ -63,10 +70,11 @@ class PurchaseOrder(models.Model):
                 #         )
                 #         % ",".join([order.display_name for order in auto_sale_order_id])
                 #     )
-                if auto_sale_order_id and (
-                    auto_sale_order_id.state not in ("sale", "done")
-                ):
-                    auto_sale_order_id.with_user(
+                if auto_sale_order_id:
+                    not_confirmed_orders = auto_sale_order_id.filtered(
+                        lambda order: order.state not in ("sale", "done")
+                    )
+                    not_confirmed_orders.with_user(
                         intercompany_user.id
                     ).sudo().action_confirm()
                 else:
--- a/purchase_sale_inter_company/models/purchase_order.py
+++ b/purchase_sale_inter_company/models/purchase_order.py
@@ -105,12 +105,10 @@ class PurchaseOrder(models.Model):
                 }
             )
         # create the SO and generate its lines from the PO lines
-        sale_order_data = self._prepare_sale_order_data(
-            self.name,
-            company_partner,
-            dest_company,
-            self.dest_address_id,
-            related_pricelist,
+        sale_order_data = self.with_context(
+            pricelist=related_pricelist
+        )._prepare_sale_order_data(
+            self.name, company_partner, dest_company, self.dest_address_id
         )
         sale_order = (
             self.env["sale.order"]
@@ -137,7 +135,7 @@ class PurchaseOrder(models.Model):
             sale_order.with_user(intercompany_user.id).sudo().action_confirm()
 
     def _prepare_sale_order_data(
-        self, name, partner, dest_company, direct_delivery_address, pricelist
+        self, name, partner, dest_company, direct_delivery_address
     ):
         """Generate the Sale Order values from the PO
         :param name : the origin client reference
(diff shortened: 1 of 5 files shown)


purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -3,26 +3,76 @@
...
+            po_partner = purchase_order.partner_id
+            # vendor can be a member of group company or it's child
+            # dest_company will be anyway member of group company
             dest_company = (
-                purchase_order.partner_id.commercial_partner_id.ref_company_ids
+                po_partner.ref_company_ids or po_partner.parent_id.ref_company_ids
             )
+            intercompany_user = dest_company.intercompany_sale_user_id
+            if not intercompany_user:
+                intercompany_user = self.env.user
```
SilvioC2C: These lines may return a `res.users` recordset as `intercompany_user`, but at line 70 we have `intercompany_user.id` which will raise an error if there's more than 1 record. Please fix this.

purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -3,26 +3,76 @@
...
+                    .search([("auto_purchase_order_id", "=", purchase_order.id)])
+                )
+                # if len(auto_sale_order_id) > 1:
+                #     raise UserError(
+                #         _(
+                #             "You can have only one auto-generated sale order "
+                #             "but following orders were found '%s' "
+                #         )
+                #         % ",".join([order.display_name for order in auto_sale_order_id])
+                #     )
```
SilvioC2C, line 72: Why is this commented out?

purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -3,26 +3,76 @@
...
-                    dest_company.id
-                )._inter_company_create_sale_order(dest_company)
+                # one sale order is expected to be created when confirm purchase order
+                # so we check for existing sale order with auto_purchase_order_id set
+                # we don't need extra sale orders if purchase order was confirmed again
+                auto_sale_order_id = (
+                    self.env["sale.order"]
+                    .sudo()
+                    .search([("auto_purchase_order_id", "=", purchase_order.id)])
+                )
```
SilvioC2C, line 62: This `search()` may return a `sale.order` recordset, but at line 67 we have `auto_sale_order_id.state` which will raise an error if there's more than 1 SO. Please fix this.

purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -3,26 +3,76 @@
...
 
+    auto_sale_order_info = fields.Html(compute="_compute_auto_sale_order_info")
+
+    def _compute_auto_sale_order_info(self):
+        for purchase_order in self.sudo():
+            auto_sale_order_id = (
+                self.env["sale.order"]
+                .sudo()
+                .search([("auto_purchase_order_id", "=", purchase_order.id)])
+            )
```
SilvioC2C, line 21: This `search()` may return a `sale.order` recordset, but at line 28 we have `auto_sale_order_id.display_name` which will raise an error if there's more than 1 SO. Please fix this.

purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -70,19 +121,40 @@ def _inter_company_create_sale_order(self, dest_company):
...
+                    ("company_id", "=", dest_company.id),
+                ]
+            )
+            if len(related_pricelist) > 1:
+                raise UserError(
+                    _(
+                        "There is more than one pricelist for this vendor "
+                        "with the same currency"
+                    )
                 )
```
SilvioC2C, line 153: Why do you raise this error instead of just fetching 1 of the many pricelists?

purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -137,14 +213,14 @@ def _prepare_sale_order_data(
         if self.notes:
             new_order.note = self.notes
         new_order.commitment_date = self.date_planned
+        if self.env.context.get("pricelist"):
+            new_order.pricelist_id = self.env.context.get("pricelist")
         return new_order._convert_to_write(new_order._cache)
 
-    def _prepare_sale_order_line_data(self, purchase_line, dest_company, sale_order):
+    def _prepare_sale_order_line_data(self, purchase_line, sale_order):
```
SilvioC2C, line 228: Changing a method signature can be disruptive. Have you checked it usage in other modules?

SilvioC2C (changes requested):

purchase_sale_inter_company/models/purchase_order.py
yankinmax: It's standard OCA flow for this module.
The client refuses it, so I commented for future self to understand.

purchase_sale_inter_company/models/purchase_order.py
yankinmax: this method (I mean with such parameters) is only used here for this customer

yankinmax: This is the exact change we do in OCA module to have customer's expected workflow.

yankinmax:
@SilvioC2C thanks for your review.
I've analysed your suggestions and will update this PR and propose these changes in next release.
AS for the PR itself: we've analysed with @gurneyalex different ways to handle customer needs.
Of course the best would be to create customer specific module and override workflow, but it isn't possible in current OCA state.

SilvioC2C:
@yankinmax OCA modules should not contain customer-specific needs, they should be generic enough to be included in public repos. If these changes are customer-specific, they should be made directly in the customer's project.

Besides, these comments are about bugs that break the execution of Odoo itself, so they should be fixed regardless:
1. https://github.com/OCA/multi-company/pull/504#discussion_r1403569549
2. https://github.com/OCA/multi-company/pull/504#discussion_r1403571288
3. https://github.com/OCA/multi-company/pull/504#discussion_r1403571727

purchase_sale_inter_company/models/purchase_order.py
yankinmax: customer request was to have only one pricelist for vendor company with the same currency
it's a signal to check and remove not needed one

yankinmax:
@SilvioC2C I agree with you. It was a difficult decision we've taken with @gurneyalex
We had two options using standard purchase sale intercompany flow or OCA one.
The customer wanted OCA variant, but with his specific needs.
What we could do, we could have customer module with overwritten code. I'll propose that to the BA
Thanks again for your help and impact on these questions

legalsylvain:
@SilvioC2C, is it ok now ?

SilvioC2C:
@legalsylvain unfortunately no, there still are bugs in this code and it still includes customer-specific behavior :sweat_smile:

PR #506, opened 2023-09-01
[MIG] calendar_event_multi_company: Migration to 15.0

amh-mw:


calendar_event_multi_company/security/calendar_event.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2019 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
```
legalsylvain, line 4:
```suggestion
<odoo noupdate="1">
```

In Odoo, Ir.rules are set as no updatable, so it can be changed by UI.

legalsylvain (approved):
Hi. Thanks for porting this module.
- Diff review : LGTM.
- One remark inline. Could you take a look ? 

Thanks  !

calendar_event_multi_company/security/calendar_event.xml
pedrobaeza: Not always. And I prefer to let them as noupdate=0, for not having to deal with migration scripts. At the end, if you are installing this module, it's because you want this behavior set by code. If not, create your own record rule.

legalsylvain: Make sense ! I thought it was an OCA convention.

legalsylvain (commented):
/ocabot merge nobump

PR #512, opened 2023-09-22
[13.0][IMP]purchase_sale_inter_company: is intercompany fields and little improvements

GuillemCForgeFlow:
- Add is intercompany fields for the SO, PO, Stock Moves and Transfers
- Fill values for these fields in a correct way
- Add corresponding filters

@ForgeFlow

### DIFF
--- a/account_invoice_inter_company/tests/test_inter_company_invoice.py
+++ b/account_invoice_inter_company/tests/test_inter_company_invoice.py
@@ -342,6 +342,19 @@ class TestAccountInvoiceInterCompanyBase(SavepointCase):
             cls.product_consultant_multi_company.company_ids = False
             cls.product_consultant_multi_company.product_tmpl_id.company_ids = False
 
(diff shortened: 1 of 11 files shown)


account_invoice_inter_company/tests/test_inter_company_invoice.py
```diff
@@ -342,6 +342,19 @@ def setUpClass(cls):
...
 
+        cls.product_product_multi_company = cls.env["product.product"].create(
+            {
+                "name": "Product Multi Company",
+                "uom_id": cls.env.ref("uom.product_uom_unit").id,
+                "uom_po_id": cls.env.ref("uom.product_uom_unit").id,
+                "categ_id": cls.env.ref("product.product_category_2").id,
+                "type": "product",
+            }
+        )
```
JordiMForgeFlow, line 353: you could move the creation of the new product just after the creation of product_consultant_multi_company and then clear the field company_ids under the same 'if' condition in line 340

PR #515, opened 2023-10-02
[16.0][ADD] account_reconcile_model_multicompany_propagate

EmilioPascual:
Allow propagate reconcile model to other companies with same chart template.

https://www.loom.com/share/1d4c0856e990467087e9e824477784c8

@yajo @Shide @loida-vm @ArantxaSudon @rafaelbn please review, if you want 😉. Thanks.

MT-3690 @moduon


account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
```diff
@@ -0,0 +1,183 @@
...
+            taxes_map[account.company_id.id][tax.name] = tax.id
+
+        # Propagate to other companies
+        for company in companies_chart_template_equal:
+            rec_model = self.search(
+                [("company_id", "=", company.id), ("name", "=", self.name)]
+            )
+            if rec_model:
+                for field in self._propagated_fields():
+                    rec_model.write({field: self[field]})
```
yajo: This is the same, but a bit more readable:

```suggestion
                    rec_model[field] = self[field]
```

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
```diff
@@ -0,0 +1,183 @@
...
+        ]
+
+    def propagate_multicompany_account_reconcile_model(self):
+        """Set the same value for all companies with same chart template."""
+        alien_companies = self.env.user.company_ids - self.env.company
+        if not alien_companies:
+            raise UserError(
+                _(
+                    "There are no other companies to propagate to. "
+                    "Make sure you have access to other companies."
```
yajo: Let's make the message clearer for the user.

```suggestion
                    "Make sure you have access to other companies "
                    "that share the same chart of accounts."
```

account_reconcile_model_multicompany_propagate/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,3 @@
+To configure this module, you need to:
+
+#. Install module account that lets you manage reconcile models (if not already installed).
```
yajo: Not really. This is a dependency, so it will be installed 100% of the times.

However, you have to get permissions to manage those, as you did in your video. Maybe you can explain that here instead.

You can also explain that you need to be granted access to other companies that share the same CoA.

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
```diff
@@ -0,0 +1,183 @@
...
+            "payment_tolerance_param",
+            "payment_tolerance_type",
+            "match_partner",
+            "match_partner_ids",
+            "match_partner_category_ids",
+            "past_months_limit",
+            "number_entries",
+        ]
+
+    def propagate_multicompany_account_reconcile_model(self):
```
yajo: Just to make it a bit shorter and clearer:

```suggestion
    def propagate_to_other_companies(self):
```

IMHO it'd be good to split this method into smaller methods. However it's just a suggestion, it's ok as it is.

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
```diff
@@ -0,0 +1,183 @@
...
+                        for line in self.partner_mapping_line_ids
+                    ]
+                )
+                rec_model.update(
+                    {
+                        "partner_mapping_line_ids": [(6, 0, rec_model_partners.ids)],
+                    }
+                )
+
+    def _get_vals_create(self, company):
```
yajo: Also a small rename to make the method's purpose more obvious:

```suggestion
    def _get_vals_create_propagated(self, company):
```

account_reconcile_model_multicompany_propagate/views/account_reconcile_model.xml
```diff
@@ -0,0 +1,47 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="//form/sheet" position="before">
+                <header>
+                    <button
+                        name="propagate_multicompany_account_reconcile_model"
+                        icon="fa-clone"
+                        string="Propagate to companies"
+                        title="Propagate to companies with same chart of accounts"
+                        type="object"
+                        groups="account.group_account_user"
```
yajo, line 25: IMHO it'd be good to confirm before proceeding:

```suggestion
                        groups="account.group_account_user"
                        confirm="This will copy this model to other companies where you have access, if they share the same chart of accounts. If they already have a model with the same name, it will be updated in place. Are you sure to proceed?"
```

yajo (changes requested):
In general terms, this looks pretty good.

There are some suggestions, most of them not very important.

I think it would be good if you could add a couple of more tests. Coverage is a bit low on this PR.

Thanks!

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
EmilioPascual: It's true. Changed.

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
EmilioPascual: Done

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
EmilioPascual: You are right. Method renamed and splitted.

account_reconcile_model_multicompany_propagate/views/account_reconcile_model.xml
EmilioPascual: Added.

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
EmilioPascual: Method renamed

account_reconcile_model_multicompany_propagate/tests/test_reconcile_model.py
```diff
@@ -0,0 +1,162 @@
...
+            self.rule_1.partner_mapping_line_ids.mapped("partner_id"),
+        )
+        self.assertEqual(
+            rule_1_company_2.partner_mapping_line_ids.mapped("payment_ref_regex"),
+            self.rule_1.partner_mapping_line_ids.mapped("payment_ref_regex"),
+        )
+        self.assertEqual(
+            rule_1_company_2.partner_mapping_line_ids.mapped("narration_regex"),
+            self.rule_1.partner_mapping_line_ids.mapped("narration_regex"),
+        )
```
yajo: IMHO it would be better to replace all these assertions for just one call, similar to this:

```suggestion
        self.assertRecordValues(
            rule_1_company_2,
            [{
                "sequence": "1000",
                "rule_type": "invoice_matching",
                ...,
                "partner_mapping_line_ids": [
                    {
                        "partner_id": cls.partner_1.id,
                        "payment_ref_regex": "payment ref regex",
                        "narration_regex": "narration regex",
                    },
                    ...,
                ],
            }]
        )
```

The benefits of having it hardcoded is that, instead of implicitly testing that all the calls to `mapped()` do what you expect them to do, you assert the final results. Also, the test becomes more readable.

However, this is just a enhancement suggestion. The test is good. 😉

yajo (approved):

loida-vm (changes requested):
Paso a detallar un poco lo visto y sugerencias: 

**VISTO:** 

- Modelos de conciliación con una cuenta genérica, el modelo se propaga con la cuenta incluida
- Si el partner está establecido en el modelo, y es un partner compartido para todas las empresas, también se aplica
- Salta un aviso de seguridad (comento más abajo sugerencia sobre el aviso)
- Si el modelo contiene una cuenta específica del plan de cuentas de la compañía en la que se creó, el modelo se duplicará en el resto pero sin la cuenta contable aplicada
- Si le doy más de una vez al botón PROPAGAR, no se duplica cuantas veces pinche en él. Solo se duplica una vez  (@rafaelbn)

**SUGERENCIAS:**

- Texto de aviso: Da a entender si se comparte el plan de cuentas, y un plan de cuentas no se puede compartir, es específico para cada compañía.
Sugiero: "This will copy this model to other companies that the user has access to. It will copy the account, in case it belongs to the generic chart of accounts.  If they already have a model with the same name, it will be updated in place. Are you sure to proceed?"
- ¿Podría contener un campo el modelo para indicar "Compañía"? Si este campo está establecido con la compñía a la que queremos que el modelo pertenezca, entonces no permitir que se propague. Evitará errores.

Gracias!!

account_reconcile_model_multicompany_propagate/tests/test_reconcile_model.py
EmilioPascual: You're right. I changed test.

EmilioPascual:
> * Texto de aviso: Da a entender si se comparte el plan de cuentas, y un plan de cuentas no se puede compartir, es específico para cada compañía.
>   Sugiero: "This will copy this model to other companies that the user has access to. It will copy the account, in case it belongs to the generic chart of accounts.  If they already have a model with the same name, it will be updated in place. Are you sure to proceed?"

@yajo has already requested a change in confirm text, but i forgot update translation to spanish. If you test module in spanish, you didn't see right confirm text. I have just update spanish translation

> * ¿Podría contener un campo el modelo para indicar "Compañía"? Si este campo está establecido con la compñía a la que queremos que el modelo pertenezca, entonces no permitir que se propague. Evitará errores.

I think it's strange behaviour, maybe a wizard where you select companies with same CoA. What do you think? @loida-vm  @rafaelbn

loida-vm:
> I think it's strange behaviour, maybe a wizard where you select companies with same CoA. What do you think?

I think that having a wizard with that behaviour doesn't make sense. Maybe, the mistake is already avoided when the confirmation window comes out....
What is your opinion @rafaelbn ?

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def propagate_to_other_companies(self):
+        """Set the same value for all companies with same chart template."""
+        alien_companies = self.env.user.company_ids - self.env.company
+        if not alien_companies:
+            raise UserError(
+                _(
+                    "There are no other companies to propagate to. "
+                    "Make sure you have access to other companies "
+                    "that share the same chart of accounts."
```
Shide, line 85:
```suggestion
                    "that has the same chart of accounts."
```

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
```diff
@@ -0,0 +1,205 @@
...
+        # Map accounts by company and code
+        accounts_map = self._get_account_map_propagate(companies_chart_template_equal)
+
+        # Map taxes by company and name
+        taxes_map = self._get_taxes_map_propagate(companies_chart_template_equal)
+
+        # Propagate to other companies
+        for company in companies_chart_template_equal:
+            rec_model = self.search(
+                [("company_id", "=", company.id), ("name", "=", self.name)]
```
Shide, line 102: If the reconciliation model is matched by the name, could be good to show a banner informing that the reconciliation model will be propagated to another ones if someone clic on the button.
In case of an user creates a new one that matches the name of another reconciliation model in another company, he/she will be warn by this banner.

Shide (approved):
Seems good, some minor change and a suggestion

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
EmilioPascual: Good suggestion. Added

image

Shide: I think the banner can be blue (it's showing information and maybe nothin needs to change) and the button can throw a `confirm="are you sure bla bla?"` popup to avoid miss-clicks

EmilioPascual: You're right abount banner. I changed it

image

About confirm, it already exists in button 

image

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
EmilioPascual: Done

loida-vm (approved):

loida-vm:
Funtional review done. LGTM!!
Thanks @EmilioPascual

EmilioPascual:
@gurneyalex, @JordiBForgeFlow or @jgrandguillaume can you merge please?

PR #520, opened 2023-10-17
[14.0] [IMP] stock_intercompany: Extend behavior to add a delivery creation on receipt creation.

paradoxxxzero:
- Add `intercompany_picking_creation_mode` configuration to specify which pickings will be created.
- Add `intercompany_out_type_id` for delivery picking creation.
- Add `intercompany_parent_id`/`intercompany_child_ids` relation between automatically created pickings.
- Refactor creation code.
- Fix and add unit tests (test was a tautology).
- Update documentation.


stock_intercompany/models/stock_move.py
```diff
@@ -0,0 +1,17 @@
...
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _action_assign(self):
+        """
+        Override to create counterpart pickings after moves are assigned.
+        """
+        res = super()._action_assign()
+
+        counterparts = self.mapped("picking_id")._create_counterpart_pickings("out")
```
sebastienbeau: mapped is not required you can directly use self.picking_id (and it will map all picking_id or moves in self)

```suggestion
        counterparts = self.picking_id._create_counterpart_pickings("out")
```

stock_intercompany/models/stock_picking.py
```diff
@@ -1,49 +1,142 @@
...
+        Hook to check if the given company is valid for the given mode.
+        """
+        return True
+
+    def _prepare_counterpart_picking_vals(
+        self, company, intercompany_type, warehouse, counterusage
+    ):
+        """Prepare the values to create the counterpart picking"""
+        vals = {
+            "partner_id": self.env.user.company_id.partner_id.id,
```
sebastienbeau: Do not take the company on the user. It's the company of the picking
```suggestion
            "partner_id": self.company_id.partner_id.id,
```

sebastienbeau (commented):
for the "in" case (When we create a "in" for company A => create a out for company B)
We should also sync the cancel.
So If I cancel the "in" in company A it cancel the "out" in company "B"

stock_intercompany/models/stock_picking.py
paradoxxxzero: This was in the original module : https://github.com/OCA/multi-company/blob/14.0/stock_intercompany/models/stock_picking.py#L22 

Should I change it anyway?

stock_intercompany/models/stock_move.py
```diff
@@ -0,0 +1,17 @@
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _action_assign(self):
```
sebastienbeau: Are you sure that we need this ?

stock_intercompany/models/stock_picking.py
sebastienbeau: yes it's a bug.
Since v13, the company of the user is not changed when you change the company in the header "company selector". A user belong to one company whatever the company where is logged.
So when you are in company A and create an "in" in company B. The out will be delivered to company A.
So you can fix it, it was a bug

A user always have one company (that do not depend on the seelect

stock_intercompany/models/stock_move.py
paradoxxxzero: Yes in case of picking creation from `procurement.group`.

stock_intercompany/models/stock_picking.py
```diff
@@ -1,49 +1,270 @@
...
+
+        vals = {
+            "partner_id": self.company_id.partner_id.id,
+            "company_id": company.id,
+            "picking_type_id": intercompany_type.id,
+            "state": "draft",
+            "location_id": location_src.id,
+            "location_dest_id": location_dest.id,
+            "intercompany_parent_id": self.id,  # Keep track of the parent picking
+        }
```
bguillot, line 158: Please set group_id: False in the counterpart picking. procurement groups are not shared between companies.

PR #522, opened 2023-10-19
[14.0][FIX] base_multi_company: allow `_check_company` to work with multiple companies

renda-dev:
In this PR we wrote a monkey patch of `BaseModel._check_company` to make it compatible with the field `company_ids` introduced by this module.

The bug behind this fix was presented in the following scenario:

- A record of a model that has been extended by `multi.company.abstract` had two companies set: C1 and C2
- The behavior of this module would have made the company used when creating the record the main `company_id`, while also having both companies in `company_ids`
- Odoo's `BaseModel._check_company` would have checked only `company_id`, raising an error if the user belonged to company C2 even if the record has both companies set


francesco-ooops (approved):
Functional ok, although module stock_intercompany still triggers error, but that module lacks documentation and tests

cc @paradoxxxzero

ilyasProgrammer (approved):
code LGTM

francesco-ooops:
@pedrobaeza do you think this is a correct solution?

pedrobaeza:
I'm afraid I'm not understanding the problem scenario.

francesco-ooops:
@pedrobaeza the functional issue is as follows: 

INSTALL

https://github.com/OCA/multi-company/tree/14.0/purchase_sale_inter_company
https://github.com/OCA/multi-company/tree/14.0/product_multi_company

SETUP

For both companies C2 and C3: 

image

STEPS TO REPRODUCE

Create product and enable both companies with product-multi-company:

image

Now let's proceed with flow for purchase_sale_inter_company:

C2 creates PO with Vendor C3 for product 1
C3 confirms SO for customer C2 > validate picking: 

CURRENT BEHAVIOR

User error: 

Incompatible companies on records:
- 'S00014/Vendors>Stock' belongs to company 'C2' and 'Stock Rule' (rule_id: 'C3: Stock → Customers') belongs to another company.

francesco-ooops:
This is not the only scenario where this error is displayed to user, eg: 

Select C2 as main active company for user and C3 enabled >
go to product 1 > update qty on hand to 100 >
switch to C3 as main active company > try to change qty on hand to 50

Incompatible companies on records:
- 'Stock>Inventory adjustment' belongs to company 'C2' and 'Product' (product_id: 'Product 1') belongs to another company.

this is why @renda-dev approached the issue with this solution instead of a bridge module

francesco-ooops:
@pedrobaeza go?

pedrobaeza:
I have to analyze it. At first sight, the monkey-patching is not a good practice.

francesco-ooops:
@pedrobaeza sorry to ask as you've already checked many PRs of ours in the last weeks, but this is breaking

francesco-ooops:
@pedrobaeza we really miss you at OCA Days! 

I would have loved to ask you a feedback on this in person ;)

francesco-ooops:
@JordiBForgeFlow opinions? this is a major incompatibility between two important modules of this repo

francesco-ooops:
@OCA/core-maintainers I really need this issue fixed. Who can help out? Please and thank you 🙏

francesco-ooops:
Hi @pedrobaeza I really need your feedback on this as it's blocking the whole flow, else please merge on existing reviews

pedrobaeza:
This monkey-patching is a very serious thing which I dislike and provokes other side effects. I need to dig deeper to find the issue and propose an alternative solution. Meanwhile, you can use this PR in your deployments.

francesco-ooops:
> Meanwhile, you can use this PR in your deployments.

@pedrobaeza This is one of the biggest challenges we have with trying to be an OCA-only company, as we keep all our production DBs regularly updated with OCA code, and aggregating in production a PR that can change heavily in the future feels way too risky. So far we haven't found a viable way to keep DBs regularly updated while also adding OCA open PRs to production environments. 

On the contrary, for OCA modules, we feel having code outside the "standard" (merged) versions is invalidating the main reasons that motivate us to develop in OCA rather than internally; for this reason we do not use gitaggregator in production environments, and this is also why we are constantly requesting code to be merged.

francesco-ooops:
Apart for this, I understand your concern for the monkey-patch. If you can take some time to share your impressions, we can dig deeper into the issue ourselves :)

pedrobaeza:
My company is also OCA only, and you don't need to wait for the PRs to be merged to deploy to the customer (at least not all). It's not all black or white. First, you have to distinguish between the kind of pull requests:

- A new module or data model changes. This one may require extra careful indeed, but there's even methods I will tell you later.
- A patch that is fixing a problem, that don't involve any data model change. You can deploy this patch immediately to your customer, as if it solves the current problem, unless the solution is worst than the bug itself, it's OK to put it. This PR is paradigmatic for this. It solves the bug, but the implementation is not the best one, because it can have side effects if other module does the same, or give problems on corner cases, so it's not good for a "massive deployment". But meanwhile, it's good enough for that specific customer having the problem. When the definitive patch comes (if there's any), you just have to remove the PR in your deployment instructions. Even more, having this approach serves for detecting such side issues without having the risk to contaminate other customers of yours using this module.

About the first case, there's also options. You do a datamodel, and then, someone here at OCA suggest something that you see reasonable, but it's already deployed on the customer. Migration scripts can be added even being in PR, so the 1.0.0 version never releases publicly, but 2.0.0 version does, and you include migration scripts for migrating from 1.0.0 to 2.0.0, and your customer smoothly transitions with the module deployed.

Anyway, as your developers are getting more and more experienced, less of these drastic datamodel changes will happen in the middle of the cycle. And it's also OK to say no to suggestions, specially on new modules, like it's happening on the module you have proposed for product default codes. The only conflict here is if you modify existing modules that may vary the experience on the installed base, or if the new module conflicts in any way with the existing ones.

Said that, I will try to check for this ASAP.

francesco-ooops:
Thanks Pedro, I appreciate the guidance!

Will discuss this internally. Cheers

francesco-ooops:
@pedrobaeza got any time to look at it?

simahawk:
> On the contrary, for OCA modules, we feel having code outside the "standard" (merged) versions is invalidating the main reasons that motivate us to develop in OCA rather than internally; for this reason we do not use gitaggregator in production environments, and this is also why we are constantly requesting code to be merged.

Ciao @francesco-ooops  :)

As per our discussion at OCA days, you don't have to wait. Go for git-aggregator or if you are using pip install just install modules from PRs straight (aggregating to your own fork).

I completely share what @pedrobaeza explained. Here are some additional thoughts:

* relying on the fact that updating a whole repo blindly to the latest commit is "safe" is quite risky. You might include changes that you've never seen. In that case I don't see too much difference compared to including a not merged PR which is kind of ready.
* if you want to make sure the changes are frozen, you can pin the base commit for the repo (eg: last tested commit instead of 14˙.0) and you can pin the PRs themselves by pushing to your own fork

From my POV non merged PRs are the real "alpha" or "beta" version of the module. If we had just one repo per module I would see this as a feature branch where the module is released on pypi as alpha version 1.0.0a1

pedrobaeza:
Hi, please check a cleaner solution for the problem in #535

PR #526, opened 2023-10-25
[14.0] [FIX] account_invoice_inter_company: Remove self company from eligible companies in `_find_company_from_invoice_partner`

renda-dev:
In some countries (for example, Italy) they need to create invoices to themselves which should not be counted as "inter company invoice"

This PR address that issue, also because, from my prospective, it does not make a lot of sense to create an invoice to yourself and also create the related inter-company invoice as well.

### DIFF
--- a/account_invoice_inter_company/models/account_move.py
+++ b/account_invoice_inter_company/models/account_move.py
@@ -57,7 +57,13 @@ class AccountMove(models.Model):
         company = (
             self.env["res.company"]
             .sudo()
-            .search([("partner_id", "=", self.commercial_partner_id.id)], limit=1)
+            .search(
+                [
+                    ("partner_id", "=", self.commercial_partner_id.id),
+                    ("id", "!=", self.company_id.id),
+                ],
+                limit=1,
+            )
         )
         return company or False
 


ilyasProgrammer (approved):
LGTM

GSLabIt (approved):

francesco-ooops:
@OCA/intercompany-maintainers is this ok for you?

MiquelRForgeFlow (approved):

francesco-ooops:
@pedrobaeza merge on existing reviews?

pedrobaeza:
Ouch, this is also affected by https://github.com/OCA/oca-addons-repo-template/pull/214

renda-dev:
Would it be possible to be merged as minor? 

I kinda need this change in another PR (https://github.com/OCA/l10n-italy/pull/3659) 👀

pedrobaeza:
No, it's not possible. If you are in a hurry, make another PR applying manually on the file the patch proposed in the template.

francesco-ooops:
@pedrobaeza for your knowledge, any idea why patched versions are not updated in other repos?

pedrobaeza:
I don't get your question.

francesco-ooops:
@renda-dev can you explain the issue you're facing as you're the technical one?

renda-dev:
> @renda-dev can you explain the issue you're facing as you're the technical one?

Sure! Please correct me if I’m mistaken tho @pedrobaeza 

Substantially in other OCA's repos, the dependencies for testing are being fetched in `Looking in indexes: https://wheelhouse.odoo-community.org/oca-simple-and-pypi` ([quick ref in this PR](https://github.com/OCA/multi-company/actions/runs/6638030583/job/18033555139?pr=526#step:4:19)) and in my recent experience it considerate only the latest minor (or major) version of the module, ignoring pre-releases (patches).

These small changes are pretty important for us because integrating `account_invoice_inter_company` in **l10n-italy** would break a couple of modules and tests will fail (for the same reason explained in this PR's description)

The small term solution to this problem would be to manually add the pre-release version to `test-requirements.txt` (like [here](https://github.com/OCA/l10n-italy/pull/3659/files#diff-fac4c6890301d4de5c3f4266837803d5240c84a3d8b6c735bbc6a64c39d2f94eR2)), but that's obviously wrong.

Please, let me know if I'm clear enough and if there are any other solutions for bringing minor module versions to other OCA repos, thanks!

Quick note: We've had experience with this issue where there was no bumping at all, maybe even patches are enough to update the module in pypi :)

pedrobaeza:
But why don't you simply fix this repository and then we can merge?

https://x.com/PedroMBaeza/status/1717834741943009487?s=20

HekkiMelody:
I think what @renda-dev is pointing out is a separate problem from the pre-commit one.

If you merge without increasing the major or minor version of the module, the pypi package for the module is not updated and the change is not available for testing on other repositories, therefore blocking any other PR that depends on this one.

pedrobaeza:
OK, but I think this is not a problem here. I tried to merge with `patch` option, which also put the same mechanism and it's not a problem. There's no need of bumping the minor version AFAIK. The only problem is if we use `nobump` option.

renda-dev:
> OK, but I think this is not a problem here. I tried to merge with `patch` option, which also put the same mechanism and it's not a problem. There's no need of bumping the minor version AFAIK. The only problem is if we use `nobump` option.

Thanks a lot for clarifying, I'm totally fine with `patch` then :)

francesco-ooops:
@renda-dev can you check if it can be fw-ported?

renda-dev:
Yes, there shouldn't be any problem at all @francesco-ooops

PR #534, opened 2023-11-23
[14.0] ir_config_parameter_multi_company implementation

deniscraciungabriel:


eLBati (approved):
Thanks

legalsylvain (commented):
Hi. Thanks for your contribution.
A question regarding implementation. Why not just add a company_id field in the ir.config_parameter model ?

eLBati:
Hi @legalsylvain thanks for your feedback.

We considered that implementation, but, as `ir.config_parameter` contains the constraint  `('key_uniq', 'unique (key)', 'Key must be unique.')`, having a `company_id` field would imply to have duplicated `key`; and modifying the SQL constraint seemed too much invasive.

legalsylvain:
> modifying the SQL constraint seemed too much invasive.

Well, I allready saw some overload of SQL constraint in OCA modules. that is not hard to do, and it makes the code very much simpler. what is the drawback you think ?

eLBati:
@legalsylvain thanks.
Then, I would close this in favour of https://github.com/OCA/multi-company/pull/574

PR #535, opened 2023-11-28
[14.0][FIX] base_multi_company: Proper _check_company when multiple companies selected

pedrobaeza:
**Steps to reproduce:**

- Create two companies C1 and C2.
- Create a product with C1 and C2 as allowed companies.
- Having C1 as main company, update quantity on hand to 100.
- Switch to C2 as main company, but check the C1 company to be seen as well.
- Now update quantity on hand to any other value for the entry of the company C1.

**Result:**

'Stock>Inventory adjustment' belongs to company 'C2' and 'Product' (product_id: 'Product 1') belongs to another company.

**Explanation:**

This is because the computed field `company_id` that is used for checking the product company vs the stock.move company (the error is not accurate, as it says the location, but it's the move), is returning the main company where the user is.

**Solution:**

Inject a context whenever we run `_check_company` for having available the company involved, and use that company as value for the record being checked if included in the allowed companies.

@Tecnativa

### DIFF
--- a/base_multi_company/models/base.py
+++ b/base_multi_company/models/base.py
@@ -1,0 +1,20 @@
+# Copyright 2023 Tecnativa - Pedro M. Baeza
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import models
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    def _check_company(self, fnames=None):
+        """Inject as context the company of the record that is going to be compared
+        for being taking into account when computing the company of many2one's
+        relations that links with our multi-company models.
+        """
+        if self._name == "res.company":
+            company_source_id = self.id
+        elif "company_id" in self._fields:
+            company_source_id = self.company_id.id
+        self = self.with_context(_check_company_source_id=company_source_id)
+        return super()._check_company(fnames=fnames)
--- a/base_multi_company/models/multi_company_abstract.py
+++ b/base_multi_company/models/multi_company_abstract.py
@@ -1,4 +1,5 @@
 # Copyright 2017 LasLabs Inc.
+# Copyright 2023 Tecnativa - Pedro M. Baeza
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
 from odoo import api, fields, models
@@ -38,12 +39,19 @@ class MultiCompanyAbstract(models.AbstractModel):
                 record.no_company_ids = True
 
     @api.depends("company_ids")
-    @api.depends_context("company")
+    @api.depends_context("company", "_check_company_source")
     def _compute_company_id(self):
         for record in self:
-            # Give the priority of the current company of the user to avoid
-            # multi company incompatibility errors.
-            company_id = self.env.context.get("force_company") or self.env.company.id
+            # Set this priority computing the company (if included in the allowed ones)
+            # for avoiding multi company incompatibility errors:
+            # - If this call is done from method _check_company, the company of the
+            #   record to be compared.
+            # - Otherwise, current company of the user.
+            company_id = (
+                self.env.context.get("_check_company_source_id")
+                or self.env.context.get("force_company")
+                or self.env.company.id
+            )
             if company_id in record.company_ids.ids:
                 record.company_id = company_id
             else:
--- a/base_multi_company/models/__init__.py
+++ b/base_multi_company/models/__init__.py
@@ -1,4 +1,5 @@
 # Copyright 2017 LasLabs Inc.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
+from . import base
 from . import multi_company_abstract


pedrobaeza:
@francesco-ooops @ilyasProgrammer @renda-dev you can check if this one fulfills all the cases. I have tried with the 2 provided and everything is OK.

base_multi_company/models/base.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def _check_company(self, fnames=None):
+        """Inject as context the company of the record that is going to be compared
+        for being taking into account when computing the company of many2one's
+        relations that links with our multi-company models.
+        """
+        if self._name == "res.company":
+            company_source_id = self.id
+        else:
+            company_source_id = self.company_id.id
```
moylop260: `base` model inherits all the models itself but not all the models have `company_id` field defined?

You can reproduce the error using `self.env["res.lang"].company_id`
 - `AttributeError: 'res.lang' object has no attribute 'company_id'`


What about?

```suggestion
        company_source_id = False
        if self._name == "res.company":
            company_source_id = self.id
        elif "company_id" in self._fields:
            company_source_id = self.company_id.id
```

Notice the following odoo-bin shell commands:
 - Screenshot 2023-11-28 at 3 54 03 p m

MiquelRForgeFlow: `elif "company_id" in self._fields:`

pedrobaeza: AFAIK, this method is only run when `_check_company_auto` is set to True in the model from `create` and `write`, which, by definition, requires a `company_id` field.

You can try on runboat to write any record in `res.lang`, and you will see that no error is raised. I have also tried changing lang on a res.partner.

Do you have any use case where the error is triggered?

moylop260: @pedrobaeza 
You have a good point!

I don't have a functional case triggering this error

I'm only seeing a future technical error for corner cases where we don't know if we should protect

Before this PR you can call: `self.env["res.lang"].search([])._check_company()` without errors

After this PR this will raise error

But you have a good point, maybe there are not code calling the method directly like this


e.g.
 - Screenshot 2023-11-28 at 5 48 40 p m

pedrobaeza: OK, no problem. Let's include the safeguard. It's barely overhead, so it's not so important.

HekkiMelody (approved):
Code review, LGTM.

Thank you for taking the time to look at it!

ilyasProgrammer (approved):
LGTM

francesco-ooops:
@moylop260 @MiquelRForgeFlow good for you?

base_multi_company/models/base.py
```diff
@@ -0,0 +1,20 @@
...
+    def _check_company(self, fnames=None):
+        """Inject as context the company of the record that is going to be compared
+        for being taking into account when computing the company of many2one's
+        relations that links with our multi-company models.
+        """
+        if self._name == "res.company":
+            company_source_id = self.id
+        elif "company_id" in self._fields:
+            company_source_id = self.company_id.id
+        self = self.with_context(_check_company_source=company_source_id)
```
simahawk:
```suggestion
        self = self.with_context(_check_company_source_id=company_source_id)
```

base_multi_company/models/base.py
```diff
@@ -0,0 +1,20 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import models
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    def _check_company(self, fnames=None):
+        """Inject as context the company of the record that is going to be compared
```
simahawk, line 11: This is not what the method does, which still is the original behavior.
Looks  more a comment than a docstring in fact :wink:

simahawk (approved):
Only code review. Minor remarks.

moylop260 (approved):
👍

base_multi_company/models/base.py
pedrobaeza: OK, changed for better comprehension of its content.

base_multi_company/models/base.py
pedrobaeza: Yes, it is exactly what it does this method override. I'm not documenting the original method, but what and why I'm overriding the method. The original behavior is documented in the original method.

base_multi_company/models/multi_company_abstract.py
```diff
@@ -38,12 +39,19 @@ def _compute_no_company_ids(self):
                 record.no_company_ids = True
 
     @api.depends("company_ids")
-    @api.depends_context("company")
+    @api.depends_context("company", "_check_company_source")
```
moylop260, line 42: I think this part was missing to use the new variable name `_check_company_source_id`

Is not?

base_multi_company/models/base.py
```diff
@@ -0,0 +1,20 @@
...
+        for being taking into account when computing the company of many2one's
+        relations that links with our multi-company models.
+        """
+        if self._name == "res.company":
+            company_source_id = self.id
```
moylop260, line 16: Too late to comment but... I think we missed to initialize the variable `company_source_id ` here

pedrobaeza: Yes, damn variable renaming added at last moment... Doing a hot patch right now.

pedrobaeza: Fixed in d0782cf5a12b63414d2167e26d1868e234ac1fd9

pedrobaeza:
Forward-port to v15 in #536

base_multi_company/models/multi_company_abstract.py
pedrobaeza: Fixed in d0782cf5a12b63414d2167e26d1868e234ac1fd9

base_multi_company/models/base.py
pedrobaeza: Fixed in 1a359ac4a479b8778a306ab5e5137c501e582330 and applied all directly in v15.

GSLabIt:
@pedrobaeza, maybe you can help. We're experiencing a weird issue since this changes has been merged. 

Here is the error:

```
2023-11-30 11:52:37,645 152 [1;33m[1;49mWARNING[0m openerp_template odoo.modules.loading: Module stock demo data failed to install, installed without demo data 
  File "/.repo_requirements/odoo/odoo/tools/convert.py", line 677, in _tag_root
    f(rec)
  File "/.repo_requirements/odoo/odoo/tools/convert.py", line 330, in _tag_function
    _eval_xml(self, rec, env)
  File "/.repo_requirements/odoo/odoo/tools/convert.py", line 201, in _eval_xml
    return odoo.api.call_kw(model, method_name, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 404, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 391, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/.repo_requirements/odoo/addons/stock/models/stock_picking.py", line 739, in action_confirm
    self._check_company()
  File "/home/odoo/dependencies/multi-company/base_multi_company/models/base.py", line 19, in _check_company
    company_source_id = self.company_id.id
  File "/.repo_requirements/odoo/odoo/fields.py", line 3825, in __get__
    raise ValueError("Expected singleton: %s" % record)
ValueError: Expected singleton: res.company(1, 2)

The above exception was the direct cause of the following exception:

  File "/.repo_requirements/odoo/odoo/modules/loading.py", line 86, in load_demo
    load_data(cr, idref, mode, kind='demo', package=package)
  File "/.repo_requirements/odoo/odoo/modules/loading.py", line 69, in load_data
    tools.convert_file(cr, package.name, filename, idref, mode, noupdate, kind)
  File "/.repo_requirements/odoo/odoo/tools/convert.py", line 733, in convert_file
    convert_xml_import(cr, module, fp, idref, mode, noupdate)
  File "/.repo_requirements/odoo/odoo/tools/convert.py", line 799, in convert_xml_import
    obj.parse(doc.getroot())
  File "/.repo_requirements/odoo/odoo/tools/convert.py", line 719, in parse
    self._tag_root(de)
  File "/.repo_requirements/odoo/odoo/tools/convert.py", line 677, in _tag_root
    f(rec)
  File "/.repo_requirements/odoo/odoo/tools/convert.py", line 685, in _tag_root
    )) from e
odoo.tools.convert.ParseError: while parsing /.repo_requirements/odoo/addons/stock/data/stock_demo2.xml:462, near
<function model="stock.picking" name="action_confirm">
            <value model="stock.picking" eval="[                 obj().env.ref('stock.outgoing_shipment_main_warehouse').id,                 obj().env.ref('stock.outgoing_shipment_main_warehouse1').id,                 obj().env.ref('stock.outgoing_shipment_main_warehouse4').id,                 obj().env.ref('stock.outgoing_shipment_main_warehouse5').id,                 obj().env.ref('stock.outgoing_shipment_main_warehouse6').id,                 obj().env.ref('stock.incomming_shipment1').id,                 obj().env.ref('stock.incomming_shipment2').id,                 obj().env.ref('stock.incomming_shipment3').id,                 obj().env.ref('stock.outgoing_chicago_warehouse1').id,                 obj().env.ref('stock.outgoing_chicago_warehouse2').id,                 obj().env.ref('stock.outgoing_chicago_warehouse3').id,                 obj().env.ref('stock.incomming_chicago_warehouse1').id,                 obj().env.ref('stock.incomming_chicago_warehouse2').id,                 obj().env.ref('stock.incomming_chicago_warehouse3').id]"/>
        </function>
```


For some reason `company_id` having multiple values and it is raising singleton error. 

Any hint on how to solve it?

pedrobaeza:
@GSLabIt very weird steps to reproduce, but technically possible, so here you have the patch: #539

GSLabIt:
> @GSLabIt very weird steps to reproduce, but technically possible, so here you have the patch: #539

big thanks

PR #537, opened 2023-12-04
[16.0][MIG] pos_restaurant_multi_company

legalsylvain:
Trivial migration.

### Description

This module adds support for multi company on Point of Sale restaurant module.
* It adds ``company_id`` on ``restaurant.floor`` model (and a related field on ``restaurant.table``) model.
  If defined, the floor will be accessible only in the given company.

* It adds ``company_id`` on ``restaurant.printer`` model.
  If defined, the printer will be accessible only in the given company.


quentinDupont (approved):
Code review LGTM :heavy_check_mark: 
And tested whitout and with this module  :heavy_check_mark: 
But one test doesn't pass :x:

pedrobaeza:
Please indicate first what is your investigation and why you are declaring as rebel the other module instead of yours.

legalsylvain:
> Please indicate first what is your investigation and why you are declaring as rebel the other module instead of yours.

Answered here :  https://github.com/OCA/multi-company/issues/632

pedrobaeza:
Well, I think you should isolate your module, not the other. Before yours coming, everything was green, and if you isolate yours, that will continue the same. Anyway, it seems that just putting at https://github.com/OCA/multi-company/blob/91b1eafbd21417610dfd3e9622b7d66947016e26/account_multicompany_easy_creation/wizards/multicompany_easy_creation.py#L196 a sudo should be enough.

legalsylvain:
> Well, I think you should isolate your module, not the other

Definitively not. 
1) Otherwise, we have to set as rebel all the futur OCA modules that depends on ``pos_restaurant`` (or maybe other Odoo modules, who knows ?).
2) A bug should occures for the time being, in production with ``account_multicompany_easy_creation`` and ``pos_restaurant`` installed.

The problem is in ``account_multicompany_easy_creation`` not in ``pos_restaurant_multi_company``

> Anyway, it seems that just putting at  ...  a sudo should be enough.

Thanks for the investigation and the patch ! 
It is now green.

/ocabot merge nobump

PR #538, opened 2023-12-04
[14.0] [FIX] base_multi_company: apply '_name_search' fix to 'search_read'

renda-dev:
The same patch applied to `_name_search` could be applied to `search_read`, to fix some different scenarios (e.g. "Search More..." button in m2x relations would open a tree view that applies `search_read` and didn't work before this fix).

### DIFF
--- a/base_multi_company/models/multi_company_abstract.py
+++ b/base_multi_company/models/multi_company_abstract.py
@@ -82,11 +82,9 @@ class MultiCompanyAbstract(models.AbstractModel):
         return super().write(vals)
 
     @api.model
-    def _name_search(
-        self, name, args=None, operator="ilike", limit=100, name_get_uid=None
-    ):
+    def _patch_company_domain(self, args):
         # In some situations the 'in' operator is used with company_id in a
-        # name_search. ORM does not convert to a proper WHERE clause when using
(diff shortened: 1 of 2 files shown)


GSLabIt (approved):

francesco-ooops (approved):
Functional ok!

before this PR, in a field like "product" in a pricelist for a specific company, clicking "search more" would not display products without company set in field "Companies"

With the same flow for a pricelist without a company set, such products would not appear as well

francesco-ooops:
@pedrobaeza good to go?

pedrobaeza:
/ocabot merge patch

You should apply these patches to upper versions to be all paired. If not, when you decide to go to another future version (you seem stuck at v14 for now, which is not very advisable IMO, but everyone follow the path they consider), you will need to analyze again the same problems (memory is short), to finally find that you already fixed this in the past.

francesco-ooops:
> /ocabot merge patch
> 
> You should apply these patches to upper versions to be all paired. If not, when you decide to go to another future version (you seem stuck at v14 for now, which is not very advisable IMO, but everyone follow the path they consider), you will need to analyze again the same problems (memory is short), to finally find that you already fixed this in the past.

agreed, time and budget allowing we'll try to fw-port as much as possible

PR #539, opened 2023-12-04
[14.0][FIX] base_multi_company: _check_company multi-record aware

pedrobaeza:
Follow-up of #535

`_check_company` is not an `ensure_one` method, so we need to make sure that if it's called with several records, and they don't belong to the same company, everything is OK.

@Tecnativa


GSLabIt (approved):

victoralmau (approved):

pedrobaeza:
Forward-port to 15 in #540

PR #508, opened 2023-09-06
[12.0][IMP] purchase_sale_inter_company: sync pickings between companies

MiquelRForgeFlow:
Better version of https://github.com/OCA/multi-company/pull/362.

~~Note: I will fix the pre-commit only if I see interest of people in wanting the merge of this PR.~~

PR #507, opened 2023-09-01
[14.0] [IMP] purchase_sale_inter_company: sync pickings between companies

renda-dev:
Migration of https://github.com/OCA/multi-company/pull/362

Added some fixes and migration to v14

### Disclaimer:
I've added some things in the CI's tests matrix because this module can't run it's test with `stock_intercompany` installed, since the purposes of these two modules are quite similar.

Also, I can't exclude `stock_intercompany` in this module's `manifest.py` because the runboat would not even start.

I'm totally open to suggestion on how we can handle this situation.

PR #541, opened 2023-12-06
[FIX]purchase_sale_inter_company: handle pickings with lots/serials

HekkiMelody:
Previously an intercompany picking with tracked products would would throw an error.

With this fix, the method searches for a lot in the destination company that matches the one in the source company (same name and same product); a new lot is created by duplicating the original, if none is found.

In this PR I'm also adding myself and my colleague @renda-dev as maintainers of the module.


francesco-ooops (approved):
Functional ok!

purchase_sale_inter_company/models/stock_picking.py
```diff
@@ -21,19 +21,42 @@ def _action_done(self):
             purchase.picking_ids.write({"intercompany_picking_id": pick.id})
             if not pick.intercompany_picking_id and purchase.picking_ids[0]:
                 pick.write({"intercompany_picking_id": purchase.picking_ids[0]})
-            for move_line in pick.move_line_ids:
-                sale_line_id = move_line.move_id.sale_line_id
-                po_move_lines = sale_line_id.auto_purchase_line_id.move_ids.mapped(
+            for move in pick.move_lines:
```
geomer198, line 24: Why is the variable name `move`? I think the variable name should be `line` because you are iterating over the lines.

geomer198 (approved):
Code review ok!

purchase_sale_inter_company/models/stock_picking.py
HekkiMelody: those are actually `stock.move`s, not `stock.move.line`. The name is misleading.

https://github.com/OCA/OCB/blob/14.0/addons/stock/models/stock_picking.py#L320C7-L320C7

francesco-ooops:
@pedrobaeza good for you?

pedrobaeza (approved):
Another candidate one to be fw-port it

/ocabot merge patch

HekkiMelody:
> Another candidate one to be fw-port it
> 
> /ocabot merge patch

Thank you once again for the merge @pedrobaeza 

Unfortunately, this whole part of the code that deals with synchronizing the picking between two companies (and not just the SO/PO) was only added to v12 and then to v14 quite recently. 

v12: https://github.com/OCA/multi-company/pull/508
v14: https://github.com/OCA/multi-company/pull/507 

In v15 and v16 there is a whole new module stock_intercompany https://github.com/OCA/multi-company/tree/15.0/stock_intercompany that's entirely independent code.

So a few of our recent PRs can't be FW easily. I'm not sure if you have any advice on how to proceed from here.

pedrobaeza:
In fact, I think this should go into `purchase_sale_stock_intercompany` or similar module name, to act as glue module, unless the other module is enough to join everything. The new split is logical, as you may have service sale/purchase inter-connection, but not forcing to have `stock` installed. You should accommodate the code of this PR to the new module structure.

pedrobaeza:
It seems the module already exists: https://github.com/OCA/multi-company/tree/15.0/purchase_sale_stock_inter_company, so you only need to plug this code into them.

PR #542, opened 2023-12-06
[FIX] purchase_sale_inter_company: sync moves with same product

HekkiMelody:
This commit fixes a bug that occurred when the intercompany PO/SO and therefore the pickings had multiple lines of the same product.

In that case, the quantities done would not be synced correctly, but rather the larger quantity would be applied
to all destination moves. The receiving company would receive more product than was ordered in the PO.


francesco-ooops (approved):
Bug reproduced, fix is ok!

geomer198 (approved):
LGTM!

PR #549, opened 2023-12-17
 [16.0][MIG] product_category_company_favorite (from 12.0)

legalsylvain:
### Description

This module is usefull in a multi-company context if
- the companies doesn't have to use all the same categories
- if you apply companies to your products.

Without that module, when creating products, end user can select categories
that don't make sense for the current company.
With that module, categories manager can set for each company, which category
will be displayed, reducing configuration error. (and so accounting error
if accounts are set at the category level.)

**Note**

- When creating **new root category**:

    - it is favorite for all the other company if it is favorite in the current context.
    - it is not favorite for all the other company if it is not favorite in the current context.

- When creating **new child category**:

    - the category will inherit the configuration of the parent category in each company.

- when creating a **new company**:

    - all the existing categories are favorite in the new created company.

### Configure

* Select a company you want to configure

* Go to "Inventory" > "Configuration" > "Product Categories"

* Uncheck categories you don't want to use for the current company.

product_category_tree

### Usage

* Go in "Inventory" > "Master Data" > "Products"

* Create a new product

product_form_view

End user doesn't have access to the category "All / Consumable" that is not favorite in the current company.


legalsylvain:
This one should be ready to review :

- CC : original reviewer of https://github.com/OCA/multi-company/pull/428 (@fcayre, @carmenbianca)

- CC : authors / reviewers of ``product_category_company`` that is a concurrent implementation. @ carmenbianca, @Kev-Roche, @cathLemb, @PaulGoubert

rvalyi (approved):
great module!

product_category_company_favorite/i18n/fr.po
```diff
@@ -0,0 +1,63 @@
...
+"If you have any queries regarding your account, Please contact us.\n"
+"\n"
+"Thank you in advance for your cooperation.\n"
+"Best Regards,"
+msgstr ""
+
+#. module: product_category_company_favorite
+#: model:ir.model.fields,field_description:product_category_company_favorite.field_product_category__is_favorite
+msgid "Favorite"
+msgstr "favorie"
```
quentinDupont:
```suggestion
msgstr "favorite"
```

product_category_company_favorite/i18n/fr.po
```diff
@@ -0,0 +1,63 @@
...
+
+#. module: product_category_company_favorite
+#: model_terms:ir.ui.view,arch_db:product_category_company_favorite.view_product_category_search
+#, fuzzy
+msgid "Favorites"
+msgstr "favorie"
```
quentinDupont:
```suggestion
msgstr "favorites"
```

quentinDupont (changes requested):
Otherwise, code lgtm and well tested :heavy_check_mark:

legalsylvain:
Thanks @quentinDupont. I fixed the translation.

/ocabot merge patch

PR #548, opened 2023-12-13
[14.0][FIX]base_multi_company fix singleton   File "/opt/odoo/parts/OCA/mul…

matteoopenf:
Problem with singleton

   res = super().write(values)
  File "/opt/odoo/odoo/odoo/addons/base/models/res_company.py", line 233, in write
    res = super(Company, self).write(values)
  File "/opt/odoo/odoo/odoo/models.py", line 3736, in write
    self._check_company()
  File "/opt/odoo/parts/OCA/multi-company/base_multi_company/models/base.py", line 21, in _check_company
    company_source_id = self.id
  File "/opt/odoo/odoo/odoo/fields.py", line 3825, in __get__
    raise ValueError("Expected singleton: %s" % record)
Exception

PR #551, opened 2023-12-18
[16.0][FIX] base_multi_company: Check multiple companies

SirAionTech:
Before this PR, executing `_check_company` on multiple companies failed with our beloved
> Expected singleton

error

**Steps**

1. Install **stock**: this module sets `_check_company_auto = True` in https://github.com/odoo/odoo/blob/6485872d00f47e351e40741283f3b9219adcd8a8/addons/stock/models/res_company.py#L9
2. Create multiple companies in one shot with `self.env["res.company"].create([ ... ])` syntax

**Actual behavior**
Error traceback ending with something like
```
  File "/path/to/multi-company/base_multi_company/models/base.py", line 21, in _check_company
    company_source_id = self.id
  File "/path/to/odoo/odoo/fields.py", line 5069, in __get__
    raise ValueError("Expected singleton: %s" % record)
ValueError: Expected singleton: res.company(214, 215)
```

**Expected behavior**
No error

### DIFF
--- a/base_multi_company/models/base.py
+++ b/base_multi_company/models/base.py
@@ -18,9 +18,9 @@ class Base(models.AbstractModel):
         for record in self:
             company_source_id = False
             if record._name == "res.company":
-                company_source_id = self.id
+                company_source_id = record.id
             elif "company_id" in record._fields:
                 company_source_id = record.company_id.id
-            self = self.with_context(_check_company_source_id=company_source_id)
-            super()._check_company(fnames=fnames)
+            record = record.with_context(_check_company_source_id=company_source_id)
+            super(Base, record)._check_company(fnames=fnames)
         return


base_multi_company/tests/test_multi_company_abstract.py
```diff
@@ -280,3 +280,18 @@ def test_set_company_id(self):
...
+            [
+                {
+                    "name": "Test multiple creation 1",
+                },
+                {
+                    "name": "Test multiple creation 2",
+                },
+            ],
+        )
+        companies._check_company()
```
pedrobaeza: Calling directly to the involved method is not a very good test strategy, as it's prompt to fail by other reasons apart from the patch provided (or to not fail as exited early), so I wouldn't add tests this time.

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Please check if this is applicable to other versions.

base_multi_company/tests/test_multi_company_abstract.py
SirAionTech: Thanks for the very fast review.
I always try to add a test when fixing something so I'm a bit sad this has to be removed, but you are making a good point: the records might have some data that would trigger an error in `_check_company` and that is allowed as long as `check_company_auto` is `False`.

Also thanks for merging already :smile:

SirAionTech:
> Please check if this is applicable to other versions.

It is applicable to `15.0`:
https://github.com/OCA/multi-company/blob/81092e7ddd53c60723a8b9eb7be49596e42ab0d7/base_multi_company/models/base.py#L21-L25

and `14.0`:
https://github.com/OCA/multi-company/blob/29f8d599e4ca278aa2dd41c6f4d2b7f5de12b353/base_multi_company/models/base.py#L21-L25
here it was partially fixed by https://github.com/OCA/multi-company/pull/548.


Can't find similar code anywhere else.
I'll probably _oca-port_ tomorrow or the day after.

PR #555, opened 2023-12-20
[14.0][FIX] purchase_sale_inter_company: fix demo data

HekkiMelody:
Just changing these fields to make it easier to test the module.

Technically a bug in "stock" module's demo data. Contact Jeff Lawson is also the "address" of warehouse "Chicago 1".
The fields property_stock_customer and supplier are set to "Inter-warehouse transit" on Jeff Lawson for company San Francisco and not for company Chicago as it should be.

Since there are no routes set for location "Inter-warehouse transit", creating an inter-company PO from Chicago to SF fails; when the inter-company SO is automatically created, with delivery address towards the Chicago Warehouse i.e. Jeff Lawson, the creation of the SO's picking fails as there is no route available.


francesco-ooops:
@pedrobaeza anything against? it's too late to fix it on 14 demo data

pedrobaeza (commented):
Well, the proper final solution is to not use demo data at all. Do you have the same problem if you create a partner on the fly for tests instead of using `stock.res_partner_address_41`?

francesco-ooops:
@pedrobaeza demo data are useful for functional testing and bug reporting :)

pedrobaeza:
Yes, that's its purpose. Not for tests.

HekkiMelody:
There may be a misunderstanding here. Automated tests/unit tests already work well.

This change is purely to make functional testing easier.

pedrobaeza (approved):
OK then.

PR #558, opened 2023-12-22
[14][IMP] account_invoice_inter_company : Ensure supplier reference is set on supplier invoices/refund

florian-dacosta:
Even in case the supplier document is the source one.

@bguillot @sebastienbeau


bguillot (approved):
Thanks !

PR #559, opened 2023-12-27
[14.0][FIX] purchase_sale_inter_company: sync confirmed unlocked purchase updates

chienandalu:

cc @Tecnativa TT46773

please take a look @pedrobaeza


purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -186,3 +186,38 @@ def button_cancel(self):
...
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        """Sync lines between an open purchase and an open sale order"""
+        lines = super().create(vals_list)
+        for order in lines.order_id.filtered(lambda x: x.state == "purchase"):
+            sale = self.env["sale.order"].search(
+                [("auto_purchase_order_id", "=", order.id)]
```
pedrobaeza: Why not adding a o2m for avoiding this ugly syntax?

pedrobaeza (commented):
I miss a `write` override for also synchronizing changed quantities/prices.

purchase_sale_inter_company/models/purchase_order.py
chienandalu: It looks like o2m relations are limited by the record rules and I couldn't access it even with `sudo()`. An alternative approach has been a compute field with `compute_sudo`

chienandalu:
> I miss a write override for also synchronizing changed quantities/prices.

I added that logic as well. For the moment only to sync quantities

francesco-ooops:
> When the purchase user updates the PO after the SO is generated and always if both are open, we sync the new lines to update the demand.

thanks for the PR @chienandalu , just to understand better functionally: what states are you referring to with "open"?

chienandalu:
> thanks for the PR @chienandalu , just to understand better functionally: what states are you referring to with "open"?

Basically, both the purchase and its sale order are in states `purchase/sale`

francesco-ooops:
@chienandalu ok, so they're both "confirmed" :)

Will test and review

chienandalu:
> @chienandalu ok, so they're both "confirmed" :)

Yes, but not `done` (locked)

purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -3,13 +3,30 @@
...
 
-from odoo import _, models
+from odoo import _, api, fields, models
 from odoo.exceptions import UserError
 
 
 class PurchaseOrder(models.Model):
     _inherit = "purchase.order"
 
+    auto_sale_order_id = fields.Many2one(
```
pedrobaeza: Let's name properly new fields (and the same for the other):

```suggestion
    intercompany_sale_order_id = fields.Many2one(
```

purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -186,3 +203,107 @@ def button_cancel(self):
...
+            self.env["sale.order.line"].with_user(intercompany_user.id).sudo().create(
+                sale_lines
+            )
+        return lines
+
+    @api.model
+    def _get_purchase_sale_line_sync_fields(self):
+        """Map purchase line fields to the synced sale line peer"""
+        return {
+            "product_qty": "product_uom_qty",
```
pedrobaeza, line 265: Please include price_unit here (and discount should be added in a glue module).

purchase_sale_inter_company/models/stock_picking.py
```diff
@@ -30,13 +30,28 @@ def _action_done(self):
...
-                    raise UserError(
-                        _(
-                            "Mismatch between move lines with the "
-                            "corresponding  PO %s for assigning "
-                            "quantities and lots from %s for product %s"
-                        )
-                        % (purchase.name, pick.name, move.product_id.name)
+                    note = _(
+                        "Mismatch between move lines with the "
+                        "corresponding  PO %s for assigning "
```
pedrobaeza:
```suggestion
                        "corresponding PO %s for assigning "
```

purchase_sale_inter_company/models/purchase_order.py
chienandalu: IMO the price of the sale order is the one that should prevail. We don't really want to sync that value from the PO to the SO (although we could want to do it back but we don't have the mechanism at this moment)

pedrobaeza: OK, please add it to the roadmap.

purchase_sale_inter_company/models/stock_picking.py
chienandalu: done

purchase_sale_inter_company/models/purchase_order.py
chienandalu: done

purchase_sale_inter_company/models/purchase_order.py
chienandalu: done

pedrobaeza (approved):

francesco-ooops (approved):
Sync works well with following use case: 

Chicago purchases from San Francisco, product Acoustic Block Screen qty 5 > confirm
Chicago updates POL qty to 10 
San Francisco finds qty updated to 10 on SOL and picking OUT

It doesn't work well with flow: 

Chicago purchases from San Francisco, product Acoustic Block Screen qty 10 > confirm
Chicago updates POL qty to 5
San Francisco still has qty 10 on SOL and picking OUT

Seems standard behavior for PO tho... I think at least there should be a warning like there is in SO, "You are decreasing the ordered quantity! Do not forget to manually update the delivery order if needed."

chienandalu:
> Chicago purchases from San Francisco, product Acoustic Block Screen qty 10 > confirm
> Chicago updates POL qty to 5
> San Francisco still has qty 10 on SOL and picking OUT

I just tested it on runboat and it works as expected:

- the sale order gets the updated qty to the decreased one and the picking schedules an warning activity for the decreased qty (this is the standard in this version when you decrease the demand from the sale order)

pedrobaeza:
Please rebase, as there's now a conflict.

chienandalu:
- Rebased
- Added a test with @francesco-ooops case

pedrobaeza:
Please fw-port it.

PR #560, opened 2023-12-27
[14.0] [IMP] purchase_sale_inter_company: manage and sync state

renda-dev:
Put inter company incoming picking in "waiting for another operation state" if it hasn't been validated or canceled, otherwise sync it to the other picking state.


francesco-ooops (approved):
Functional ok!

@pedrobaeza @chienandalu what do you say?

PR #561, opened 2023-12-29
[15.0][FIX] purchase_sale_inter_company: sync confirmed unlocked purchase upadtes

chienandalu:
fw from 

- #559 

cc @Tecnativa TT46773

please review @pedrobaeza 

@francesco-ooops @aleuffre @renda-dev I started forwarding your `14.0` prs but the task isn't trivial as the module was split in v15


pedrobaeza (approved):

francesco-ooops:
@chienandalu I tried testing on runboat, installing the module

do you know why there is only one company in the instance?

chienandalu:
> do you know why there is only one company in the instance?

I guess you used the base runboat. If you use the full one, you'll find all the demo data.

francesco-ooops:
@chienandalu ok, unrelated to this PR, but what module/application are the multi-company demo data related to?

chienandalu:
I think that the demo data for Chicago goes with `stock`

francesco-ooops (approved):
Functional ok!

PR #563, opened 2024-01-02
[16.0][FIX] purchase_sale_inter_company: sync confirmed unlocked purchase updates

chienandalu:
fw from 

- #559 #561

cc @Tecnativa TT46773

please review @pedrobaeza @francesco-ooops


pedrobaeza (approved):
Code review

francesco-ooops (approved):
LGTM

PR #569, opened 2024-01-16
[IMP] company_dependent_flag: icon position improved on company dependent fields

dora-jurcevic:


bealdav (approved):
Tested on project with several impact fields. 

Nice works. Thanks

dora-jurcevic:
CSS style modified in order to display the building icon on the same line as the input field on the form. The two elements (the icon and the field) are now better aligned


BEFORE 
Screenshot from 2024-01-16 14-25-18

AFTER

Screenshot from 2024-01-16 14-24-11

legalsylvain (approved):
Hi. Interesting module I discover. Thanks for this important patch.

Note : maybe matter of taste, but I think the icon should be before the label, and not after. If so, it will be less confusing (because less close to the field to be entered)

bealdav:
Thanks a lot @legalsylvain you're right it's better to be near label but it's harder and require more code because label is either near to the field or either before with `<div><label for="field"/></div>`.

Then this patch is a kind quick and welcome fix with very few code, then sure.

Other behavior needs other PR, you're right, but code could be more tricky

bealdav:
Thanks a lot @legalsylvain

PR #570, opened 2024-01-17
intercompany_shared_contact: fix wrong field computation and add test

sebastienbeau:
origin_company_id was not computed correctly when having parent_id

Add test and fix the bug


sebastienbeau:
@Kev-Roche

PR #575, opened 2024-01-24
[14.0][FIX] partner_multi_company: default value for res_partner.company_ids

PicchiSeba:
When `company_ids` is not set that means the contact will be available to all companies. This might create some noise and unwanted behavior.

A user assigned to only one company that creates a new contact without changing the `company_ids` field is going to make the contact available for anyone.

We can set the field to the current company in order to prevent this. If the user really wants to make the contact available to everyone they have to manually change the field.

Moreover the current module behavior goes against what's written in the USAGE file
> On the partner form view, go to the "Sales & Purchases" tab, and put the companies in which you want to use that partner. If none is selected, the partner will be visible in all of them. The default value is the current one.


francesco-ooops (approved):
Functional ok

@pedrobaeza asking for your opinion on this: 

- for sure it's not correct that users with only one allowed company can create partners for all companies, so that should be fixed somehow
- I also don't want to change existing behaviors without notice

How do you suggest we approach this? Should we implement this behavior optionally through a setting?

PicchiSeba:
Updated the docs with the new option

PicchiSeba:
I also added a test to address these changes

PicchiSeba:
Ok, we should be good by now.

Let me know if you see any issue

partner_multi_company/models/res_partner.py
```diff
@@ -17,6 +17,11 @@ class ResPartner(models.Model):
         store=True,
         index=True,
     )
+    company_ids = fields.Many2many(
+        default=lambda self: [(6, 0, [self.env.company.id])]
+        if self.env.company.set_active_company_partner
+        else [(6, 0, [])]
+    )
```
renda-dev: **nitpick(non-blocking):**
```suggestion
    company_ids = fields.Many2many(
        default=lambda self: [(6, 0, [self.env.company.id])]
        if self.env.company.set_active_company_partner
        else False
    )
```

renda-dev (approved):
Code and functional review, LGTM

francesco-ooops:
@pedrobaeza in the end we went for the setting, in order to minimize disruption. Let us know if it's good for merge, thanks!

HekkiMelody (approved):
Code review, LGTM

In the tests, there are probably simpler ways to to fix the issue with the missing accounts, but that's still fine IMO.

francesco-ooops:
@pedrobaeza shall we merge?

PR #581, opened 2024-01-31
[16.0][ADD] sale_stock_partner_warehouse

santostelmo:
### Description
Allow to choose by default a warehouse on SO based on a Partner parameter. 
If the warehouse parameter is completed, then this value would be the default one on the sale. 
If empty, then the warehouse on sale would be the standard default warehouse.


sale_stock_partner_warehouse/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "data": [
+        "views/partner_view.xml",
+    ],
+    "depends": [
+        "sale_stock",
+        "sale_stock_warehouse_multicompany",
```
jbaudoux: You don't need this dependency
```suggestion
```

sale_stock_partner_warehouse/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    warehouse_id = fields.Many2one(
+        compute="_compute_warehouse_id", store=True, precompute=True
+    )
+
+    @api.depends("user_id", "company_id", "partner_id.sale_warehouse_id")
```
jbaudoux: If you change the wh on the partner, you probably don't want to recompute all SO of that partner
```suggestion
    @api.depends("user_id", "company_id", "partner_id")
```

jbaudoux (changes requested):
This module is not related to multi-company and must move to sale-workflow repo

sale_stock_partner_warehouse/__manifest__.py
santostelmo: Removed dependency

santostelmo:
> This module is not related to multi-company and must move to sale-workflow repo

Moving it to sale-workflow: https://github.com/OCA/sale-workflow/pull/3044

santostelmo:
@jbaudoux can you plase check the PR on OCA sale-workflow ?

PR #583, opened 2024-02-09
[14.0] [FIX] purchase_sale_inter_company: wrong field

renda-dev:
This PR fixes a bug that was preventing the module to work properly.

Without this, an error would be raised every time that `action_confirm` is triggered.

### DIFF
--- a/purchase_sale_inter_company/models/sale_order.py
+++ b/purchase_sale_inter_company/models/sale_order.py
@@ -59,7 +59,7 @@ class SaleOrder(models.Model):
                         for move in moves:
                             purchase_move = purchase_moves.filtered(
                                 lambda m: m.product_id.id == move.product_id.id
-                            )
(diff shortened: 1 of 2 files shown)


HekkiMelody:
Seems like the field in question existed on v12 but was changed on v14

https://github.com/OCA/OCB/commit/57f805f71e9357870dfc2498c5ef72ebd8ab7273#diff-55c6314416a6a400da6acd5018d161a55eeeb0e3008fec8828121e3dd12be0eb

HekkiMelody (approved):

HekkiMelody:
The fix is fine.

Would recommend adding a test that covers this use case, since just executing this part of the code once would have caught the problem.

francesco-ooops (approved):
Functional ok!

HekkiMelody (approved):

PR #584, opened 2024-02-09
[16.0][ADD] sale_product_company

santostelmo:
In the product template allows you to select a list of companies in which you can sell it.
Note that if the product's company field is set, it takes priority over the list of companies.

**Note**: We could add a multi-company field  `company_ids`  on product.template with the module `base_multi_company`. But that it is not an option as we don't want to have any side effect on logistics or accounting flows related product selection.


cyrilmanuel (approved):

sale_product_company/models/sale_order.py
```diff
@@ -0,0 +1,29 @@
...
+    @api.model
+    def get_view(self, view_id=None, view_type="form", **options):
+        res = super().get_view(view_id, view_type, **options)
+        if view_type == "form":
+            order_xml = etree.XML(res["arch"])
+            product_id_fields = order_xml.xpath("//field[@name='product_template_id']")
+            if product_id_fields:
+                product_id_field = product_id_fields[0]
+                domain = product_id_field.get("domain", "[]")
+                domain = domain.replace(
```
simahawk: This seems quite fragile. IMO you should parse the domain, find the leaf you want to replace and rebuild the domain.

sale_product_company/tests/test_sale_partner_company.py
```diff
@@ -0,0 +1,35 @@
...
+
+from lxml import etree
+
+from odoo.tests import Form, common
+
+
+class TestSaleProductCompany(common.TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 14: tracking_disable pls

simahawk (changes requested):

sale_product_company/models/sale_order.py
santostelmo: Implemented something equivalent with `base_view_inheritance_extension`

simahawk (approved):

cyrilmanuel (approved):

cyrilmanuel:
hiii @pedrobaeza  can you merge this PR please ?

pedrobaeza:
For me, the module name is incorrect, as it doesn't reflect the usual structure of them. It should be something like `product_sellable_per_company`, starting with `product`.

jbaudoux (approved):

jbaudoux:
> For me, the module name is incorrect, as it doesn't reflect the usual structure of them. It should be something like `product_sellable_per_company`, starting with `product`.

I don't agree. It depends on sale and modify the behavior of the sales order. So it should start with `sale`. I find current name appropriate

cyrilmanuel (approved):

cyrilmanuel:
Hello @pedrobaeza and @jbaudoux , Could you agree on this or on the changes to be made?

pedrobaeza:
Well, I already said about the module name, so I'm refraining from doing anything more. @simahawk can merge if you all agree it's a good module name.

gurneyalex (approved):

Issue #585, opened 2024-02-09
Multi-company product variant access error


HekkiMelody:
## Module
- `base_multi_company`
- `product_multi_company`

## Describe the bug
Users without administrator access have an error when accessing product variants with `company_ids` set, when some of those companies are not active.

## To Reproduce
**Affected versions**:
- 14.0 (Tested on runboat)
- Other versions untested so far

Assuming a DB with demo data such as runboat, with `product_multi_company` installed
1. As admin user, create a product template.
2. In the `Companies` field of the product template, select all 3 companies --> This means that users from any of the 3 companies should be able to see it
3. Even if you only activate company San Francisco, the user can see all 3 companies in the
4. Login as demo user (any user with multi-company access, but without administrator rights), only have company "San Francisco active"
5. Try to access the product template you created earlier --> OK. in the `companies` field only "San Francisco" is shown
6. Try to access the product variant related to the template (from the product variants menu up top, not from the smart button in the product template, they are different views) --> Access error related to trying to access the other 2, inactive companies.

![image](https://github.com/OCA/multi-company/assets/78726989/17f87a86-e2e4-410a-8f9d-3952f4aeb048)

**Expected behavior**
User with only a single company active can see a product variant when the active company is among the allowed companies

**Additional context**

Administrator users can always see all companies even if they don't have them active, so the error is not present. For non-admin users, the error seems to happen when trying to read the field `company_ids` of product.product (and in fact, if the field `company_ids` is removed from the view https://github.com/OCA/multi-company/blob/14.0/product_multi_company/views/product_template_view.xml there is no access error anymore.

Upon inspecting the calls from the browser:

- when reading the `product.template`, only company 1 is returned, even though really multiple companies are set on the fields. The other companies are not shown due to record rules filtering them. This is the correct and expected behaviour
![image](https://github.com/OCA/multi-company/assets/78726989/e5283f09-4fcf-405a-be0e-c3c01bab568e)

- when reading the `product.product`, all companies are returned; the actual error happens when the client makes a further `read` call to read the names of these companies.
![image](https://github.com/OCA/multi-company/assets/78726989/4fbe0f39-d2bc-431a-bca9-20f47beffb31)

My hypothesis is that the problem is due to the nature of `product.product` as a model that uses delegation inheritance. The `company_ids` field is really memorized on `product.template`, and maybe that's why the companies are not being filtered correctly. I haven't had time to investigate further.

francesco-ooops:
[...]

PR #588, opened 2024-02-21
[FIX] product_multi_company: fix access to product variant

aayartsev:
Fix for this issue https://github.com/OCA/multi-company/issues/585

### DIFF
--- a/product_multi_company/models/__init__.py
+++ b/product_multi_company/models/__init__.py
@@ -3,3 +3,4 @@
 
 from . import product_template
 from . import product_category
+from . import product_product
--- a/product_multi_company/models/product_product.py
+++ b/product_multi_company/models/product_product.py
@@ -1,0 +1,9 @@
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import models
+
+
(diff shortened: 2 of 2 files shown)


HekkiMelody (changes requested):
**issue**: missing change to the `__init__.py` file.

**question**: This would add the field `company_ids` to product.product, which would then be independent of the same field on `product.template`. This way, wouldn't you risk having a mismatch in access rights between a `product.template` and its variants?

aayartsev:
@aleuffre thanks, i add mised record in init file. 
As i can see this solution will not have risk because product variant record created from product template record, but product.product model did not have rules for multi-company features. It was a reason why we had mistake

francesco-ooops:
@pedrobaeza what do you think about this?

pedrobaeza (changes requested):
This doesn't seem the correct fix, and the linked issue is weird mentioning `company rule employee`, which is not related to this module.

ilyasProgrammer:
@aleuffre @pedrobaeza please take a look at another solution: https://github.com/OCA/multi-company/pull/590

francesco-ooops:
This can be closed then

PR #574, opened 2024-01-23
[14.0] ir_config_parameter_multi_company

eLBati:
From https://github.com/OCA/multi-company/pull/534

PR #589, opened 2024-02-22
[15.0] ir_config_parameter_multi_company

eLBati:
from https://github.com/OCA/multi-company/pull/574


ir_config_parameter_multi_company/models/ir_config_parameter.py
```diff
@@ -10,39 +9,17 @@ class IrConfigMultiCompany(models.Model):
...
-        return super()._get_param(key)
+        company_id = self.env.company.id
+        if company_id:
+            self.flush(["key", "value", "company_id"])
+            self.env.cr.execute(
+                "SELECT value FROM ir_config_parameter WHERE key = %s "
+                "AND (company_id = %s OR company_id IS NULL)",
+                [key, company_id],
+            )
+            result = self.env.cr.fetchone()
```
yvaucher: This does not guarantee that company value is used if both global and company specific value exists.

You can simply add ORDER BY company_id
 
NULL value will be at the end and will be ignored by the `fetchone`.

yvaucher (commented):
SQL query sholud be fixed to give priority to the company specific param when a global param exists.

ir_config_parameter_multi_company/models/ir_config_parameter.py
eLBati: Hi @yvaucher thanks, added.

I also added `def set_param` to avoid `Expected singleton` while writing a parameter already present for different companies

PR #590, opened 2024-02-23
[14.0][FIX] product_multi_company: recompute variant companies

### LINKED ISSUE #585: Multi-company product variant access error

ilyasProgrammer:
FIX https://github.com/OCA/multi-company/issues/585
Need to force recompute related product.product company_ids field. Otherwise it takes wrong values from the cache.


pedrobaeza:
This doesn't seem the correct approach, as it causes an overhead on all models. Why don't you redefine the field as computed in product variant, like other fields like `weight` do?

ilyasProgrammer:
> This doesn't seem the correct approach, as it causes an overhead on all models. Why don't you redefine the field as computed in product variant, like other fields like `weight` do?

Looks like it worked as well.
Updated

product_multi_company/models/__init__.py
```diff
@@ -1,5 +1,6 @@
 # Copyright 2015 Serv. Tecnol. Avanzados - Pedro M. Baeza
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from . import product_template
+# from . import base
```
HekkiMelody: **chore**: leftover import

product_multi_company/models/product_template.py
```diff
@@ -1,10 +0,0 @@
-# Copyright 2015-2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
-# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
-
-from odoo import models
-
-
-class ProductTemplate(models.Model):
-    _inherit = ["multi.company.abstract", "product.template"]
```
pedrobaeza: But you shouldn't remove this.

ilyasProgrammer: i didnt. Just renamed file.

product_multi_company/models/product.py
```diff
@@ -0,0 +1,27 @@
...
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    company_ids = fields.Many2many(
+        string="Companies",
+        comodel_name="res.company",
+        relation="product_variant_companies_rel",
+        compute="_compute_company_ids",
+        store=True,
```
HekkiMelody, line 15: **question**: do you think a `compute_sudo=True` is warranted here?

HekkiMelody (commented):
Code review, I think this is the correct solution for the problem, just a few minor.

product_multi_company/models/product.py
ilyasProgrammer: I think it needs to be computed with current user.

HekkiMelody: From what I understand of the base module, `company_ids` contains all the companies (including the ones the current user cannot see) and it's `company_id` that needs to be computed with the current user.

In general, if it needs to be computed with the current user it shouldn't be stored, and the other way around as well, if it's stored, the result shouldn't change depending on the user that last triggered the computation.

HekkiMelody: (Could this just be a related field? I don't know if it would work that way)

ilyasProgrammer: It is related by default because of inheritance. Even if you make that relation explicit it doesnt work. As well as with store=False.

HekkiMelody: But by default it's `related` with `store=False`, which doesn't work for direct DB queries. What if we made it `related` with `store=True`?.

pedrobaeza: But why do you mean by "direct queries"? Odoo handles this in ORM space, but not for SQL.

HekkiMelody: Mmh, maybe I misspoke. But I think that somehow the root cause of the issue is that `company_ids` is not a stored field for `product.variant`. Either way, if this solution already works I'm okay with it, I was only wondering if it could be even simpler.

HekkiMelody: It still sounds weird to me that the field `company_ids` would depend on `uid` somehow, and that it doesn't need `compute_sudo`, though. If it works, it works, of course.

pedrobaeza: It shouldn't depend on the current user for sure. The problem is that I'm giving you tips blindly, as I don't understand the reported issue and it doesn't seem related to this.

HekkiMelody: Is there something I can improve in the issue itself to make it clearer? Is there anything in particular that you think is underexplained?

pedrobaeza: The problem is that I can't reproduce the issue in this runboat:

http://oca-multi-company-14-0-d248a27cd7a5.runboat.odoo-community.org/web#id=40&action=145&model=product.product&view_type=form&cids=&menu_id=250

image

and the error message about company employee rule indicates an issue with other module, not this one.

francesco-ooops: https://github.com/OCA/multi-company/issues/585#issuecomment-1961662395

product_multi_company/readme/CONTRIBUTORS.rst
```diff
@@ -4,3 +4,6 @@
 
   * Vicent Cubells <vicent.cubells@tecnativa.com>
 * Denis Roussel <denis.roussel@acsone.eu>
+
+* Ooops404 <https://ooops404.com>
+    * Ilyas <irazor147@gmail.com>
```
pedrobaeza:
```suggestion
* Ooops404 <https://ooops404.com>

  * Ilyas <irazor147@gmail.com>
```

product_multi_company/models/product.py
pedrobaeza: OK, reproduced in base DB.

pedrobaeza (commented):
Can you avoid to store the field to see if the error persists?

ilyasProgrammer:
With store=False error is still there

pedrobaeza:
Uhm, that's very weird. This should be an ORM fault, as there's nothing in the business side leading to the problem. OK, restore it and clean commit history, and let me check once again.

pedrobaeza (approved):
It seems to work.

I think the inner problem is that in v13 it was falsely assumed that there's no need now to remove the auxiliary model:

https://github.com/OCA/multi-company/blob/12.0/base_multi_company/models/res_company_assignment.py

HekkiMelody (approved):

PR #593, opened 2024-03-06
[14.0] [FIX] product_multi_company: inherit from multi.company.abstract

renda-dev:
Required behavior:
Searching a record with a product that has `check_company=True` without any company assigned must be visible by every company.

Without inheriting `multi.company.abstract` the `product.product` model doesn't use the patched `_name_search` and `search_read`, which is essential if you want to have a relation towards `product.product` with `check_company=True`.

I've also added a test case that practically demonstrate the purpose of this PR and that it would fail without these changes.

### DIFF
--- a/product_multi_company/models/product.py
+++ b/product_multi_company/models/product.py
@@ -5,7 +5,8 @@ from odoo import api, fields, models
 
 
 class ProductProduct(models.Model):
-    _inherit = "product.product"
+    _inherit = ["multi.company.abstract", "product.product"]
+    _name = "product.product"
 
     company_ids = fields.Many2many(
         string="Companies",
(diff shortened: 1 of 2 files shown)


francesco-ooops (approved):
Functional is ok!

@pedrobaeza what do you think?

product_multi_company/models/product.py
```diff
@@ -5,7 +5,9 @@
 
 
 class ProductProduct(models.Model):
-    _inherit = "product.product"
+    _inherit = ["multi.company.abstract", "product.product"]
+    _name = "product.product"
+    _description = "Product (Multi-Company)"
```
pedrobaeza: Don't change the description.

renda-dev: Removed, thanks for the feedback!

PicchiSeba (approved):
Code Review: LGTM

francesco-ooops:
@pedrobaeza good to go? :)

PR #594, opened 2024-03-08
[14.0][MIG] company_dependent_flag: Migration to 14.0

toita86:


company_dependent_flag/models/base.py
```diff
@@ -27,12 +27,45 @@ def _update_company_dependent_css(self, arch):
...
+                    div.setAttribute("class", "o_row")
 
-    def _get_company_dependent_css_class(self):
-        """Inherit to apply your own class"""
+                    # Create a new span element
+                    span = arch.createElement("span")
+                    span.setAttribute(
+                        "class",
+                        "fa fa-lg fa-building-o title='Values set here are company-specific.'",
+                    )
```
HekkiMelody: **issue**: `title='Values set here are company-specific.'` <-- this is not part of `class`, but a new attribute.

toita86: Just fixed, thank you for pointing it out!

francesco-ooops (approved):
Functional ok!

company_dependent_flag/models/base.py
```diff
@@ -0,0 +1,72 @@
...
+                    # Check if a label already exists for the field
+                    existing_labels = arch.getElementsByTagName("label")
+
+                    if not any(
+                        label.getAttribute("for") == field.getAttribute("name")
+                        for label in existing_labels
+                    ):
+                        # Create a new label element
+                        td = arch.createElement("td")
+                        td.setAttribute("class", "o_td_label")
```
renda-dev: **question:** since there isn't any `<tr>`, are you sure that the `<td>` is needed?

company_dependent_flag/tests/test_module.py
```diff
@@ -0,0 +1,20 @@
...
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.partner_view = cls.env.ref("base.view_partner_form")
+
+    def test_class_company(self):
+        res = self.env["res.partner"]._fields_view_get(view_id=self.partner_view.id)
+        arch = minidom.parseString(res["arch"])
+        for field in arch.getElementsByTagName("phone"):
+            self.assertIn("building", field.getAttribute("class"))
```
renda-dev: **issue:** You are looking for a `<phone>` tag while you should be looking for a `<field name="phone"/>` and make sure to have it inside a `<div>` with also a `<span>` inside that same div

renda-dev (changes requested):
Beautiful work, just a couple of changes and it should be good to go

company_dependent_flag/models/base.py
toita86: Thank you for pointing that out, looking a little more into the code I noticed it was not necessary. 
The `td` element for the label was created two times, now its a lot better.

company_dependent_flag/tests/test_module.py
toita86: Fixed, following your suggestion now it searches for the `<span>` and checks if the icon is there.

renda-dev (approved):
Code and functional review, LGTM!

francesco-ooops:
@pedrobaeza good to go? :)

pedrobaeza:
As this is a modification in the web client, it should be called `web_field_company_dependent_flag` or similar.

I also miss that the image is more focused on the adaptation and remark (with a rectangle or so) which is the change.

francesco-ooops:
@pedrobaeza do you think it's convenient to rename a module existing on a latter version? I think that can be confusing and I'd rather add the requested name change to ROADMAP on most recent module version. 

> I also miss that the image is more focused on the adaptation and remark (with a rectangle or so) which is the change.

Could you clarify which change should/should not be applied? thanks!

company_dependent_flag/static/description/demo.png
pedrobaeza, line 1: I mean something like this:

imagen

francesco-ooops: @pedrobaeza sorry, what is the difference with current version?

image

pedrobaeza: What I'm saying is to include the image with the arrows in the module. Look the current inserted image, which is not significant for the user seeing the README.

francesco-ooops: oh, clear now, thanks!

francesco-ooops:
@pedrobaeza doc image improved, opened PR for roadmap [here](https://github.com/OCA/multi-company/pull/606)

shall we merge?

PR #595, opened 2024-03-15
[17.0][MIG] account_invoice_inter_company

cuongnmtm:
**Note:**

The test `test_confirm_invoice_with_product_and_shared_catalog` is failed. 
The reason is that Odoo v17 now has a `check_company=True` attribute on the `product_id` field of the `account.move.line`, which was not present in v16.

Although the "Common Product Catalog" feature of this module seems appealing, in order to make it work, we need to bypass the Odoo `check_company` attribute. This makes the feature less appealing and we should consider removing it from the module. Instead, users can remove the company value on some products to make them shared products, which is a normal procedure for inter-company.

=> I will remove the "Common Product Catalog" option.


cuongnmtm:
The test `test_confirm_invoice_with_product_and_shared_catalog` is failed. 
The reason is that Odoo v17 now has a `check_company=True` attribute on the `product_id` field of the `account.move.line`, which was not present in v16.

Although the "Common Product Catalog" feature of this module seems appealing, in order to make it work, we need to bypass the Odoo `check_company` attribute. This makes the feature less appealing and we should consider removing it from the module. Instead, users can remove the company value on some products to make them shared products, which is a normal procedure for inter-company.

account_invoice_inter_company/tests/test_inter_company_invoice.py
```diff
@@ -0,0 +1,572 @@
...
+        """With native security rule, products with no access in both companies
+        must prevent child invoice creation.
+        """
+        # ensure the catalog is shared even if product is in other company
+        self.env.ref("product.product_comp_rule").write({"active": True})
+        # Product is set to a specific company
+        self.product_a.sudo().write({"company_id": self.company_a.id})
+        # If product_multi_company is installed
+        if "company_ids" in dir(self.product_a):
+            self.product_a.suwrite({"company_ids": [(6, 0, [self.company_a.id])]})
```
tien-ld:
```suggestion
            self.product_a.sudo().write({"company_ids": [(6, 0, [self.company_a.id])]})
```

pomazanbohdan:
@cuongnmtm 
Would you be so kind as to repeat the note to the bot?
> /ocabot migration account_invoice_inter_company

pomazanbohdan:
@pedrobaeza 
pls help #604  #605

pedrobaeza:
That PRs don't have the guarantee of having a correct CI. 

https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

cuongnmtm:
> That PRs don't have the guarantee of having a correct CI.
> 
> https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

I will check it. Thanks.

carolinafernandez-tecnativa:
@cuongnmtm Hi, could you please rebase and squash administrative commits ?
So i can take a look. 

Thanks!

cuongnmtm:
@carolinafernandez-tecnativa I rebased, and squashed administrative commits.

carolinafernandez-tecnativa (approved):
Functional review LGTM :) Thanks
There are still some administrative commits such as git oca bot that should be squashed.
ping @pedrobaeza

account_invoice_inter_company/tests/test_inter_company_invoice.py
```diff
@@ -3,36 +3,60 @@
...
-                _("No Chart of Account Template has been defined !")
-            )
 
         cls.company_a = cls.env["res.company"].create(
             {
                 "name": "Company A",
                 "currency_id": cls.env.ref("base.EUR").id,
                 "country_id": cls.env.ref("base.fr").id,
-                "parent_id": cls.env.ref("base.main_company").id,
+                # "parent_id": cls.env.ref("base.main_company").id,
```
pedrobaeza: Please remove these commented lines instead of commenting them (there are several).

account_invoice_inter_company/tests/test_inter_company_invoice.py
```diff
@@ -3,36 +3,60 @@
...
-            raise ValidationError(
-                # translation to avoid pylint warnings
-                _("No Chart of Account Template has been defined !")
-            )
 
         cls.company_a = cls.env["res.company"].create(
             {
                 "name": "Company A",
                 "currency_id": cls.env.ref("base.EUR").id,
                 "country_id": cls.env.ref("base.fr").id,
```
pedrobaeza: Why putting France as country? I think we should remove this and the patching and let the system works as expected without country.

cuongnmtm: I removed the code that set currency_id, country_id and the patching.

account_invoice_inter_company/tests/test_inter_company_invoice.py
cuongnmtm: Done

pedrobaeza (approved):
/ocabot merge nobump

PR #596, opened 2024-03-18
[14.0] [IMP] purchase_sale_inter_company: remove user error on lines mismatch

renda-dev:
We would like to propose the remove of the `UserError` from this check.

The first main reason is that the `ir.config.parameter` to make it non-blocking it's not really user-friendly to configure.
Also, there could be a lot of cases where this mismatch could happen and it would be pretty hard to go through.

cc: @chienandalu

### DIFF
--- a/purchase_sale_inter_company/models/stock_picking.py
+++ b/purchase_sale_inter_company/models/stock_picking.py
@@ -50,19 +50,12 @@ class StockPicking(models.Model):
                         == ic_pick
                     ).mapped("move_line_ids")
                 )
-                if not len(move_lines) == len(po_move_lines):
+                if len(move_lines) != len(po_move_lines):
                     note = _(
                         "Mismatch between move lines with the "
                         "corresponding PO %s for assigning "
                         "quantities and lots from %s for product %s"
                     ) % (purchase.name, pick.name, move.product_id.name)
-                    # Configurable parameter so we don't lock the picking validation
-                    if (
-                        not self.env["ir.config_parameter"]
-                        .sudo()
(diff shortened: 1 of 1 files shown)


chienandalu (commented):
If I remember well, I put the parameter to avoid breaking the default behavior, which was too blocking for our use cases...

So we won't miss it if it's dropped...

Another way to address it could be:

- Adding this parameter to res.config.settings 
- Extra documenting the feature

HekkiMelody (approved):
Code review, LGTM

HekkiMelody:
> If I remember well, I put the parameter to avoid breaking the default behavior, which was too blocking for our use cases...
> 
> So we won't miss it if it's dropped...
> 
> Another way to address it could be:
> 
>     * Adding this parameter to res.config.settings
> 
>     * Extra documenting the feature

As @renda-dev said, there are multiple quite common instances where the number of lines may not match, such as a partial picking with a backorder, so I think it's fair to drop the check entirely

chienandalu (approved):
I won't miss it either :D

pedrobaeza:
Please fw-port it to upper versions:

/ocabot merge patch

renda-dev:
> Please fw-port it to upper versions

I've just checked and it seems not needed since there isn't this check in upper versions @pedrobaeza

pedrobaeza:
But I think because there are more than this PR to forward-port to upper versions. You have some of them unported in fact.

francesco-ooops:
@renda-dev in which PR was the check added?

PR #600, opened 2024-03-25
[14.0][IMP] purchase_sale_inter_company: lock incoming picking validation if intercompany transfer

toita86:


SicurSam (approved):

purchase_sale_inter_company/models/res_company.py
```diff
@@ -36,3 +36,7 @@ class ResCompany(models.Model):
         help="Sync the receipt from the destination company with the "
         "delivery from the source company",
     )
+    block_po_manual_picking_validation = fields.Boolean(
+        string="Block manual validation of picking in the destination company",
+        help="Block manual validation of picking in the destination company",
```
PicchiSeba: **nitpick**: The help attribute is used to display additional information to the user.
Change this string to add information or remove it if you deem it necessary

purchase_sale_inter_company/models/res_config.py
```diff
@@ -47,3 +47,9 @@ class InterCompanyRulesConfig(models.TransientModel):
         "the delivery from the source company",
         readonly=False,
     )
+    block_po_manual_picking_validation = fields.Boolean(
+        related="company_id.block_po_manual_picking_validation",
+        string="Block manual validation of picking in the destination company",
+        help="Block manual validation of picking in the destination company",
```
PicchiSeba: **nitpick**: unless you're redefining these attributes I would omit them. A field related to another one gets its values.

purchase_sale_inter_company/models/stock_picking.py
```diff
@@ -106,6 +106,23 @@ def button_validate(self):
...
+        # if the flag is set, block the validation of the picking in the destination company
+        if self.env.company.block_po_manual_picking_validation:
+            for record in self.sudo():
+                dest_company = record.partner_id.commercial_partner_id.ref_company_ids
+                if (
+                    dest_company
+                    and record.picking_type_code == "incoming"
+                    and record.state == "assigned"
+                    or record.state == "done"
+                    or record.state == "waiting"
```
PicchiSeba: **question**: If I'm understanding this correctly you mean `(dest_company AND picking_type_code AND state == "assigned") OR state == "done" OR state == "waiting"`

And not `(dest_company AND picking_type_code) AND state == ("assigned" OR "done" OR "waiting")`.
Right?


**suggestion**: If so I would suggest to add brackets to help visualize the logical expression

```suggestion
                    (
                        dest_company
                        and record.picking_type_code == "incoming"
                        and record.state == "assigned"
                    )
                    or record.state in ["done", "waiting"]
```

purchase_sale_inter_company/models/stock_picking.py
```diff
@@ -106,6 +106,23 @@ def button_validate(self):
                         dest_company,
                         record.sale_id,
                     )
+        # if the flag is set, block the validation of the picking in the destination company
+        if self.env.company.block_po_manual_picking_validation:
+            for record in self.sudo():
```
PicchiSeba: **question**: what would happen if we were to remove this `sudo()`?

Would we get an access error of some kind?

PicchiSeba (commented):
Code review: a couple of questions and a few non-blocking fixes

purchase_sale_inter_company/models/stock_picking.py
toita86: My intent was the second one, I have rewritten the expression based on your suggestion.

purchase_sale_inter_company/models/stock_picking.py
toita86: Thank you for pointing this out, there are no access errors so it was just useless

PicchiSeba (approved):
Code review: LGTM

francesco-ooops:
@pedrobaeza merge?

Rationale for this change: do not allow purchase picking to be validated before corresponding intercompany sale picking is validated

pedrobaeza:
I don't get this need. Isn't the sales order auto-confirmed when the PO is confirmed?

francesco-ooops:
@pedrobaeza the sale order is confirmed but the sale order picking is not (rightfully)

so what we're doing here is to prevent Purchasing company from receiving products that Selling company hasn't shipped

pedrobaeza:
OK, understood.

/ocabot merge minor

PR #602, opened 2024-03-26
[16.0][IMP] purchase_sale_inter_company: add more ctx to update locked sales

QuocDuong1306:


santostelmo:
What is the purpose of this improvement. Can you please write some context ?

purchase_sale_inter_company/models/purchase_order_line.py
```diff
@@ -87,7 +87,10 @@ def write(self, vals):
         ).sudo()
         if not sale_lines:
             return res
-        closed_sale_lines = sale_lines.filtered(lambda x: x.state != "sale")
+        state = ["sale"]
+        if self.env.context.get("allow_update_locked_sales", False):
```
santostelmo: Can you please add a test where you tests that you are able to write on po line with context key `allow_update_locked_sales` ?

santostelmo (changes requested):

purchase_sale_inter_company/models/purchase_order_line.py
QuocDuong1306: Hello @santostelmo , I updated

santostelmo (approved):

purchase_sale_inter_company/models/purchase_order_line.py
```diff
@@ -31,10 +31,13 @@ def create(self, vals_list):
         """Sync lines between an confirmed unlocked purchase and a confirmed unlocked
         sale order"""
         lines = super().create(vals_list)
+        not_allowed_states = ["cancel", "done"]
+        if self.env.context.get("allow_update_locked_sales", False):
+            not_allowed_states.remove("done")
         for order in lines.order_id.filtered(
```
gurneyalex, line 35: I would rather have this in a method that would allow overriding / extending the logic

gurneyalex (approved):

purchase_sale_inter_company/models/purchase_order_line.py
QuocDuong1306: Thank @gurneyalex , I updated

santostelmo (approved):

PR #612, opened 2024-04-24
[16.0][IMP] mail_multicompany: search most appropriate server based on the from email

MarinaAForgeFlow:


JordiMForgeFlow (approved):
code review

mail_multicompany/models/mail_message.py
```diff
@@ -19,14 +19,20 @@ def create(self, vals_list):
...
                 )
+                vals["mail_server_id"] = mail_server.id if mail_server else False
+                vals["email_from"] = email_from
         return super(MailMessage, self).create(vals_list)
+
+    def _find_company_mail_server(self, email_from, company_id):
+        # We want to find the most appropriate mail server for the given email_from
+        # but limiting it to the current active company mail servers.
+        mail_servers = (
+            self.sudo().env["ir.mail_server"].search([("company_id", "=", company_id)])
```
AaronHForgeFlow, line 38: limit=1 ?

MarinaAForgeFlow: In this case I want to retrieve all the outgoing mail servers that have as company_id the current active company. Then the _find_mail_server method will make sure only one is retrieved, the one that is a better match taking into account the from email.

AaronHForgeFlow: Ok I see. Thanks

AaronHForgeFlow (approved):
LGTM

mail_multicompany/models/mail_message.py
```diff
@@ -19,14 +19,20 @@ def create(self, vals_list):
...
+        )
+        if not mail_servers:
+            return False, email_from
```
MiquelRForgeFlow:
```suggestion
            return None, email_from
```
as the same odoo does in _find_mail_server

MiquelRForgeFlow (approved):

mail_multicompany/models/mail_message.py
```diff
@@ -19,14 +19,20 @@ def create(self, vals_list):
...
-                        [("company_id", "=", vals.get("company_id", False))],
-                        order="sequence",
-                        limit=1,
-                    )
-                    .id
+                mail_server, email_from = self._find_company_mail_server(
+                    vals.get("email_from", False), vals.get("company_id", False)
                 )
+                vals["mail_server_id"] = mail_server.id if mail_server else False
+                vals["email_from"] = email_from
```
luisg123v: instead of:

```
vals["key1"] = "val1"
vals["key2"] = "val2"
...
```

It's better to:

```
vals.update({
    "key1": "val1",
    "key2": "val2",
    ...
})
```

As a single update would be faster.

luisg123v (changes requested):
Missing commit description.

mail_multicompany/models/mail_message.py
MarinaAForgeFlow: You're right, thanks.

PR #614, opened 2024-04-30
[15.0][MIG] login_all_company: Migration to 15.0

amh-mw:


amh-mw:
Per https://github.com/odoo/odoo/commit/0573acae2306bf5da2005852da9323ddc59e5431, [web.AbstractWebClient](addons/web/static/src/legacy/js/chrome/abstract_web_client.js)  was a casualty of the initial OWL refactoring.

amh-mw:
@etobella I've migrated your module to 15. Would you mind reviewing?

login_all_company/static/src/js/company_service.esm.js
```diff
@@ -0,0 +1,20 @@
...
+        if (!cookie.current.cids && !router.current.hash.cids) {
+            var companies = [session.company_id];
+            _.forEach(session.user_companies.allowed_companies, function (company) {
+                if (company.id !== session.company_id) {
+                    companies.push(company.id);
+                }
+            });
+            cookie.setCookie("cids", companies.join(","));
+        }
+        return this._super(env, {user, router, cookie});
```
etobella: use this._super(...arguments) this will ensure that all arguments are passed if some are not used by your function

etobella (approved):
Tested functionally.

Thanks for the migration

/ocabot migration login_all_company

login_all_company/static/src/js/company_service.esm.js
amh-mw: If you don't mind, I'd prefer not to change this, mostly because we've both already tested it as is, but also because this matches the signature of the method in Odoo 15:

https://github.com/odoo/odoo/blob/670012a04f9d849ca8f0a42796ae47ecf138eaf8/addons/web/static/src/webclient/company_service.js#L34

amh-mw: Eh... even as I write that, I see the value in not blocking off avenues of extension of the company service that I cannot imagine.

amh-mw: Initial attempt failing commit with
```
eslint...................................................................Failed
  8:17  error  'user' is defined but never used  no-unused-vars
```

etobella: yes, you should remove user on the signature, as you only declare the necessary parameters, the rest will be managed by arguments. I am just trying to explain the standard usage of JS.

amh-mw: The obvious amelioration of appending `// eslint-disable-line no-unused-vars` to the line isn't working, as the `prettier` code formatter hook is moving the comment to the next line.

amh-mw: Thanks, I wouldn't have thought of that; I don't write much Javascript.

etobella: Don't worry, that's part of the review process :smile:

thomaspaulb (approved):

PR #618, opened 2024-05-08
[15.0][ADD] mail_multicompany_force_company

iTecan:
After some testing we have found this issue with mail_multicompany and how odoo detects the mail servers.

We assume that normally the different companies configured will use domains/subdomains which will match the domains/subdomains of their users.

In this case the functionality of mail_multicompany is correct and a user cannot send through an outgoing mail server that does not correspond to him.

We have found another case, a companyA with domainA and another companyB with domainB configured and assigned respectively, if an employee of companyB has indicated the email employee@domainA in his profile, Odoo in its internal configuration will look for the best matching domain and will use the outgoing mail server of companyA regardless of the fact that it is not configured for companyB.

This module forces internally that the mails of each company go through the outgoing mail server assigned to them, preventing them to use any other.

In case there is an outgoing mail server without an assigned company, any company will be able to use it following the normal usage.


mail_multicompany_force_company/models/ir_mail_server.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        domain = []
+        if not self._context.get("show_all"):
+            domain = [
+                "|",
+                ("company_id", "=", self.env.company.id),
+                ("company_id", "=", False),
+            ]
```
thomaspaulb, line 16: It seems to me a bit risky to do it this way: other parts of Odoo, or modules, might depend on the search returning mail servers for all companies, and you'd change that behavior globally.

Wouldn't it be better to override [find_mail_server](https://github.com/OCA/OCB/blob/361d7db2ea75b1d01623ea0d40d4f371fd78ac99/odoo/addons/base/models/ir_mail_server.py#L652) and inject a context key `dont_return_all` there, that you catch here and if it's there, you limit the results?

mail_multicompany_force_company/models/ir_mail_server.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo.osv import expression
+
+
+class IrMailServer(models.Model):
+    _inherit = "ir.mail_server"
+
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        domain = []
+        if not self._context.get("show_all"):
```
thomaspaulb, line 11: For such context keys it's safer to prefix them with something unique, so that it stands a lower chance of conflicting with something from another module or custom code. eg `mail_multicompany_[dont_]force_company=True`

mail_multicompany_force_company/models/ir_mail_server.py
```diff
@@ -0,0 +1,21 @@
+from odoo import api, models
+from odoo.osv import expression
+
+
+class IrMailServer(models.Model):
+    _inherit = "ir.mail_server"
+
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
```
thomaspaulb, line 9: You don't actually use these named parameters, so you can replace them by `**kwargs`, also in the `super` call

mail_multicompany_force_company/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "Email Multi Company Force Company Server",
+    "category": "Extra Tools",
+    "description": """
+       Email Multi Company Force Company Server:
+        This module forces to use the mail server that is configured for the company in outgoing mail server configuration.\n
+        If there is a mail server without a company assigned, it will be used by all companies.\n
+        In case there is a mail server without company and your company has assigned a mail server, you will use your company's mail server
```
thomaspaulb, line 8: 1. We don't use description in manifest, this should be in one of the `readme/*.rst` files. Check the OCA guidelines
2. The description on your PR was clearer, maybe you can use that one

mail_multicompany_force_company/README.rst
```diff
@@ -0,0 +1,93 @@
...
+|badge1| |badge2| |badge3|
+
+This module forces to use the mail server that is configured for the company in outgoing mail server configuration.
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Configuration
```
thomaspaulb, line 30: This should be in one of the `readme/*.rst` files, then the README.rst will be autogenerated

thomaspaulb (changes requested):

thomaspaulb:
@iTecan I see that you are first-time contributor! Feel free to ask questions or ping for any help, explanation about guidelines etc. While being strict, we also want to keep people like you aboard :-)

PR #619, opened 2024-05-13
Event tag multi company

Thibagon:
This module is useful to handle event tags in a multi company environment. It means tags from company A are not accessible to company B but common tags (tags without associated company) can be seen and used by A & B.

This module spread on tags and tags category due to their linkage


thomaspaulb:
@Thibagon Same here, pre-commit and runboat

remi-filament:
Thank  you @Thibagon 
please rename your PR [16.0][ADD] event_tag_multi_company

tests failing are not related to this module and should be fixed once #668 will be merged

PR #621, opened 2024-05-14
[14.0][FIX]company_dependent_flag: fixed missing label

toita86:
Before this fix in the pop-up window for creating contacts, the phone field was missing the label.
By not updating at every iteration the existing labels its possible to have more than two element with the same label have the flag

### DIFF
--- a/company_dependent_flag/README.rst
+++ b/company_dependent_flag/README.rst
@@ -39,6 +39,12 @@ With this module all company dependent fields are decorated with an adhoc class.
 .. contents::
    :local:
 
+Known issues / Roadmap
+======================
+
+This moudle won't work cleanly if the field in question is present twice in the view.
+It's very rare, but it could happen.
+
 Bug Tracker
 ===========
 
--- a/company_dependent_flag/models/base.py
+++ b/company_dependent_flag/models/base.py
@@ -21,14 +21,50 @@ class Base(models.AbstractModel):
         return res
 
     def _update_company_dependent_css(self, arch):
+        def _subview_check(field):
+            parent_el = field.parentNode
+            while parent_el.tagName != "form":
+                if parent_el.tagName == "field":
+                    return False  # we are in a subview
+                parent_el = parent_el.parentNode
+
+                if parent_el is None:
+                    return False
+
+            return parent_el
+
         cpny_dep_fields = [
             field_name
             for field_name, field_rec in self.env[self._name]._fields.items()
             if field_rec.company_dependent
         ]
+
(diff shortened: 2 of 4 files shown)


francesco-ooops (approved):
LGTM

image

company_dependent_flag/models/base.py
```diff
@@ -26,30 +26,28 @@ def _update_company_dependent_css(self, arch):
...
 
-                    # Create a new span element
-                    span = arch.createElement("span")
-                    span.setAttribute(
-                        "class",
-                        "fa fa-lg fa-building-o",
-                    )
-                    span.setAttribute("title", "Values set here are company-specific.")
+                    if parent_form.tagName != "form":
+                        continue  # Skip this field as it's in a subview
```
HekkiMelody: suggestion(non-blocking): this could be cleaner and more elegant if it was in its own method, IMO.

HekkiMelody (approved):
Code review, LGTM.

Another non-blocking suggestion is to add a note in `Known Issues` that this won't work cleanly if the field in question is present twice in the view.  It's very rare, but it could happen and I think it's worth mentioning.

HekkiMelody (approved):
Code review, LGTM

francesco-ooops:
@pedrobaeza what do you say?

@bealdav as module author too :)

company_dependent_flag/README.rst
```diff
@@ -39,6 +39,12 @@ With this module all company dependent fields are decorated with an adhoc class.
 .. contents::
    :local:
 
+Known issues / Roadmap
+======================
+
+This moudle won't work cleanly if the field in question is present twice in the view.
```
bealdav, line 45:
```suggestion
This module won't work cleanly if the field in question is present twice in the view.
```

company_dependent_flag/readme/ROADMAP.rst
```diff
@@ -0,0 +1,2 @@
+This moudle won't work cleanly if the field in question is present twice in the view.
```
bealdav, line 1:
```suggestion
This module won't work cleanly if the field in question is present twice in the view.
```

bealdav (approved):
LGTM looks good except typo.

Needs to forward port in 16.0

Thanks a lot

toita86:
@bealdav 
> Needs to forward port in 16.0

By checking on the runboat, the issue fixed by this PR is absent, Imo there is no need to increase the complexity of the module by doing a fwp.

bealdav:
Nice to checked, thanks a lot @toita86 .

PR #622, opened 2024-05-16
[14.0] [ADD] partner_multi_company_pos

renda-dev:
Glue module between `point_of_sale` and `partner_multi_company`.

Allows to automatically set the company when creating a partner from the POS interface if the setting 'Set Active Company Partner' is enabled.


SicurSam (approved):

GSLabIt (approved):

francesco-ooops:
@pedrobaeza is this clear enough?

pedrobaeza:
The need of this module seems a flaw in the original `partner_multi_company` module, as it should be done automatically without the glue one. The flag `set_active_company_partner` seems also an amend to cover something outside of the module itself. I have been using `partner_multi_company` in 13.0 and 15.0 without the need of this.

francesco-ooops:
@pedrobaeza I tried to look through this repo's merged PRs but couldn't find the commit where this issue was fixed for v15, we would have ported that in that case instead of making this fix: https://github.com/OCA/multi-company/pull/575

Anyway, since it's working correctly in v15 and v14 is EOL, to avoid unnecessary work I think this one could be just merged :)

pedrobaeza:
OK, but I think you are adding yourself some burden: 

/ocabot merge nobump

PR #624, opened 2024-05-21
[17.0] [MIG] base_multi_company: Migration to 17.0

rven:


mohammedshahil:
I had a quick look at this and the domain seems to be not working properly, if there is no company set. I had to rewrite the domain to "['|', ('company_ids', '=', False), ('company_ids', 'in', company_ids)]" for it to work properly. I then had a look at the Odoo's code and Odoo also seems to use the one i gave for m2m in the iap module. @rven

rven:
> I had a quick look at this and the domain seems to be not working properly, if there is no company set. I had to rewrite the domain to "['|', ('company_ids', '=', False), ('company_ids', 'in', company_ids)]" for it to work properly. I then had a look at the Odoo's code and Odoo also seems to use the one i gave for m2m in the iap module. @rven

I didn't change this domain, it has been updated within the migration of v16 https://github.com/OCA/multi-company/commit/91088f113d57eb13d994eb85896d2366edfa8bba

pedrobaeza:
It's incorrect on v16. To be fixed. Please fix it here.

rven:
@mohammedshahil @pedrobaeza Can you recheck the module with this change?

pedrobaeza (approved):

rven:
@pedrobaeza Are you sure the domain was incorrect in v16? https://github.com/odoo/odoo/commit/c18b9f44be3d666ca1e082807a2fdfc7ef2049d2

pedrobaeza:
Yes, in what you put, the left part is a many2one, not a many2many. This was discovered in #628. The question is that m2m doesn't handle correctly the `False` value in the same list.

LoisRForgeFlow (approved):
LGTM :+1:

PR #628, opened 2024-05-23
[16.0][MIG] product_multi_company: Migration to version 16.0

pilarvargas-tecnativa:
Superseed #459 because the PR is very old and the creator stopped taking care of it, this PR recovers the changes made in its migration but following all the steps of a conventional migration and finalising the migration.

cc @Tecnativa TT48972


product_multi_company/readme/USAGE.rst
```diff
@@ -0,0 +1,3 @@
+On the product form view, go to the "Information" tab, and put the companies
```
carolinafernandez-tecnativa: tab "Information" renamed for "General Information"

product_multi_company/models/product_category.py
```diff
@@ -0,0 +1,29 @@
...
+    route_ids = fields.Many2many(
+        "stock.route",
+        "stock_location_route_categ",
+        "categ_id",
+        "route_id",
+        "Routes",
+        domain=lambda self: [
+            "|",
+            ("company_id", "=", False),
+            ("company_id", "in", self.env.companies.ids),
```
carolinafernandez-tecnativa: better use [('company_id', 'in', self.env.companies.ids + [False])]

product_multi_company/models/product_category.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class ProductCategory(models.Model):
+    _inherit = "product.category"
+
+    total_route_ids = fields.Many2many(
+        domain=lambda self: [
+            "|",
+            ("company_id", "=", False),
+            ("company_id", "in", self.env.companies.ids),
```
carolinafernandez-tecnativa: better use [('company_id', 'in', self.env.companies.ids + [False])]

product_multi_company/hooks.py
```diff
@@ -0,0 +1,34 @@
...
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    # Change access rule
+    rule = env.ref("product.product_comp_rule")
+    rule.write(
+        {
+            "domain_force": (
+                "['|', ('company_ids', 'in', company_ids),"
+                "('company_ids', '=', False)]"
+            ),
```
carolinafernandez-tecnativa, line 25: same here

carolinafernandez-tecnativa:
Functional review LGTM :)

Please rebase and check for CI.
Do not forget to squash administrative commits. Thanks!

amkarthik:
@carolinafernandez-tecnativa 

We are getting this issue in this PR:
https://github.com/OCA/multi-company/issues/252

Please check the attached video below and that is recorded in PR runboat.

https://github.com/OCA/multi-company/assets/12694917/024b9513-779c-4266-9e83-8a5e8d6591a8

Thanks

pilarvargas-tecnativa:
> We are getting this issue in this PR: #252

It may have been something transient in the runbot, since duplicating the user as ‘User B’ and repeating the same operation does not cause this problem.

product_multi_company/models/product.py
```diff
@@ -1,19 +0,0 @@
-# Copyright 2023 Karthik <karthik@sodexis.com>
-# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
-
-from odoo import fields, models
-
-
-class ProductProduct(models.Model):
-    _inherit = "product.product"
-
-    company_ids = fields.Many2many(
```
pedrobaeza: Why removing this?

pilarvargas-tecnativa: It was not in v15, it is not necessary

pilarvargas-tecnativa: The fact is that by extending the abstract model in product.template, product.product acquires that field.

amkarthik:
> > We are getting this issue in this PR: #252
> 
> It may have been something transient in the runbot, since duplicating the user as ‘User B’ and repeating the same operation does not cause this problem.

Hi I tested again with the User B. (changed the password of userb to userb).
Same issue happens to me.
Please check.


https://github.com/OCA/multi-company/assets/12694917/7c930b43-2e2a-4a04-8f16-6db26143eda7

product_multi_company/models/product.py
amkarthik: @pilarvargas-tecnativa 
From what I know, the above concept is the reason for the user A and user B issue.
As product.product doesn't hold the company_ids directly in it, if someone tries to access the product.product then it tries to compute it from the product.template.
This causing the above issue as the user doesn't have access to all the companies that is set on the product.template.

pilarvargas-tecnativa:
> Hi I tested again with the User B. (changed the password of userb to userb). Same issue happens to me. Please check.

I have retested the test locally, checking in shell the data contained in the bd if it contains the company_ids field. 
image
image

I have replicated the error but it only happens in the case of new companies created, it does not happen with the demo companies already defined in odoo. I am still investigating the error.

amkarthik:
> > Hi I tested again with the User B. (changed the password of userb to userb). Same issue happens to me. Please check.
> 
> I have retested the test locally, checking in shell the data contained in the bd if it contains the company_ids field. image image
> 
> I have replicated the error but it only happens in the case of new companies created, it does not happen with the demo companies already defined in odoo. I am still investigating the error.

Hi,
Can you please confirm the user that you are using in the shell?

pilarvargas-tecnativa:
@amkarthik Actually, the access rule to the ‘product.product’ model was not being applied and it is necessary to do it as it was done in v13 and v14. This same error already occurs in v15 where this part of the code has been removed. This error should no longer appear and I have added the corresponding unit tests so that it does not happen again. Thanks

product_multi_company/models/product_category.py
pilarvargas-tecnativa: thanks

carolinafernandez-tecnativa:
Please check CI, still having problems with multi-company rules, thanks!

amkarthik:
> @amkarthik Actually, the access rule to the ‘product.product’ model was not being applied and it is necessary to do it as it was done in v13 and v14. This same error already occurs in v15 where this part of the code has been removed. This error should no longer appear and I have added the corresponding unit tests so that it does not happen again. Thanks

Thanks

product_multi_company/models/product_product.py
amkarthik, line 1: @pilarvargas-tecnativa 
Why we are going back to using create and write method as in v13.
Why don't we consider creating related field instead?
        comodel_name="res.company",
        column1="product_id",
        column2="company_id",
        relation="product_product_company_rel",
        related="product_tmpl_id.company_ids",
        compute_sudo=True,
        readonly=False,
        store=True,
    )

This is in the PR #459

pilarvargas-tecnativa: Having debugged all the code after all the conflicts with tests of other modules that uncovered the errors that occurred in this one, it is certain that this option is the most appropriate. The tests to check that the variants have their access permissions correctly set are also maintained. Thanks

carolinafernandez-tecnativa:
With these changes, when creating new companies and accessing with created users to product variant the error seems to disappear.
I also have checked accessing in product template and now it seems to work fine, Thanks !

carolinafernandez-tecnativa (approved):

product_multi_company/__manifest__.py
```diff
@@ -7,10 +7,12 @@
     "author": "Tecnativa," "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/multi-company",
     "category": "Product Management",
-    "version": "15.0.1.0.0",
+    "version": "16.0.1.0.0",
     "license": "AGPL-3",
     "depends": ["base_multi_company", "product", "stock"],
-    "data": ["views/product_template_view.xml"],
+    "data": [
+        "views/product_template_view.xml",
```
pedrobaeza: This change is not needed. Please undo in the corresponding commit.

product_multi_company/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+
+{
+    "name": "Product multi-company",
+    "summary": "Select individually the product template visibility on each " "company",
+    "author": "Tecnativa," "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/multi-company",
+    "category": "Product Management",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "depends": ["base_multi_company", "product", "stock"],
```
pedrobaeza: The dependency on stock must be removed.

product_multi_company/models/product_category.py
```diff
@@ -8,22 +8,14 @@ class ProductCategory(models.Model):
...
-            "|",
-            ("company_id", "=", False),
-            ("company_id", "in", self.env.companies.ids),
-        ]
+        domain=lambda self: [("company_id", "in", [False] + self.env.companies.ids)]
     )
 
     route_ids = fields.Many2many(
-        "stock.location.route",
+        "stock.route",
```
pedrobaeza: Remove these overrides. That should go if ever done on a module `product_multi_company_stock`.

pedrobaeza (changes requested):

product_multi_company/models/product_category.py
pilarvargas-tecnativa: All done! thanks

pedrobaeza (approved):
/ocabot merge nobump

amkarthik (approved):

cyrilmanuel:
Hello @pedrobaeza, could you merge this PR ?

PR #630, opened 2024-05-23
[16.0][FIX] sale_product_company: fix test_sale_partner_company

pilarvargas-tecnativa:
Make tests compatible with other modules that hide the company_id field. For example, product_multi_company hides the company_id field in the form and using Form to run the tests causes a conflict. To solve this, the functionality of the form is replicated directly in the tests without using Form.

Error: https://github.com/OCA/multi-company/actions/runs/9211021131/job/25339390299#step:8:286

cc @Tecnativa TT48972

### DIFF
--- a/sale_product_company/tests/test_sale_partner_company.py
+++ b/sale_product_company/tests/test_sale_partner_company.py
@@ -3,7 +3,7 @@
 
 from lxml import etree
 
-from odoo.tests import Form, common
+from odoo.tests import common
 
 from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
As this is caused by a new module introduce via PR, the solution should come in the same PR with an extra commit, not a new PR, so closing.

Anyway, I'm trying this hack for not doing this ugly thing or requiring constantly to change things in other modules: https://github.com/OCA/multi-company/compare/ee58c53d73146e88b89d2822573a4d6644d072b5..f73a885470bd41c875cf578e157690da0b77c0f2

PR #633, opened 2024-06-07
[16.0][ADD] product_default_code_res_company_code

quentinDupont:
## Description

This module extends the functionality of product module to generate automatically internal reference
(``default_code``) for product.

Note, for that purpose, the field become readonly, and a sequence is created automatically per company (that's why it's in multi-company repo)

sequence_form

The new products will have a default code with the following pattern ``COD-XXXXXX``, where ``COD`` is the code of the company of the product, and ``XXXXXX`` is a autoincremented value.

product_product_form

This new feature is enabled only for products that belong to a company.

## This PR

We use this module since several years in our company, it was a custom module that we want to share with the OCA now that we know it fits well users need :tada: 

## Not compatible with..
With https://github.com/OCA/product-variant/tree/16.0/product_variant_default_code since both modules writes on default_code with different business logic


hparfr (approved):
Do you know about : https://github.com/OCA/product-variant/tree/16.0/product_variant_default_code
?
It can be interesting to specify if there is some incompatibilities.

quentinDupont:
> Do you know about : https://github.com/OCA/product-variant/tree/16.0/product_variant_default_code ? It can be interesting to specify if there is some incompatibilities.

I did not know about it, thank you !
I just test it and indeed, both modules can't work together
I add it to exclude module then

petrus-v (approved):
Code reviews only, looks good to me !

Thanks

PR #637, opened 2024-06-11
[14.0][FIX] Picking sync should also work for dropshipping

thomaspaulb:
In case of dropshipping, picking.partner_id is the end customer (picking.sale_id.partner_shipping_id) so this partner's ref_company_ids will be empty and the picking will not sync. Use sale_id.partner_id instead, which is always the intercompany partner.


HekkiMelody (approved):
Code Review, LGTM

christian-ramos-tecnativa (approved):
LGTM

Deriman-Alonso:
Code review LGTM 👍.

Deriman-Alonso (approved):

PR #639, opened 2024-06-12
[16.0][FIX] check if purchase come from intercompany sale

Kev-Roche:
ensure compatibility with sale_purchase_inter_company module
cc @chafique-delli , @metaminux , @sebastienbeau

### DIFF
--- a/purchase_sale_inter_company/models/purchase_order.py
+++ b/purchase_sale_inter_company/models/purchase_order.py
@@ -37,9 +37,18 @@ class PurchaseOrder(models.Model):
                 purchase_order.partner_id.commercial_partner_id.ref_company_ids
             )
             if dest_company and dest_company.so_from_po:
-                purchase_order.with_company(
-                    dest_company.id
-                )._inter_company_create_sale_order(dest_company)
+                # ensure compatibility with sale_purchase_inter_company
+                # module
+                create_sale_order = False
+                if "auto_sale_order_id" in self._fields:
+                    if not self.auto_sale_order_id:
+                        create_sale_order = True
+                else:
+                    create_sale_order = True
+                if create_sale_order:
+                    purchase_order.with_company(
+                        dest_company.id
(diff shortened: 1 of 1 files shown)


purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -37,9 +37,18 @@ def button_approve(self, force=False):
...
-                    dest_company.id
-                )._inter_company_create_sale_order(dest_company)
+                # ensure compatibility with sale_purchase_inter_company
+                # module
+                create_sale_order = False
+                if "auto_sale_order_id" in self._fields:
+                    if not self.auto_sale_order_id:
+                        create_sale_order = True
+                else:
+                    create_sale_order = True
```
thomaspaulb, line 47: Couldn't this be written more concisely as:

```
# compatibility with sale_purchase_inter_company
if getattr(self, "auto_sale_order_id", False):
    continue
purchase_order.with_company(...).inter_company_create...
```

purchase_sale_inter_company/models/purchase_order.py
```diff
@@ -172,14 +181,25 @@ def _prepare_sale_order_line_data(self, purchase_line, dest_company, sale_order)
         return new_line._convert_to_write(new_line._cache)
 
     def button_cancel(self):
-        sale_orders = (
```
thomaspaulb, line 175: Here, to be less invasive to the code and also not introduce a loop, you could do instead:

```
purchase_orders = self.filtered(lambda p: not getattr(p, "auto_sale_order_id", False))
(...)
.search([("auto_purchase_order_id", "in", purchase_orders.ids)])
(...)
purchase_orders.write({"partner_ref": False})
```

thomaspaulb (approved):
Reviews non blocking, I guess

metaminux:
Hi @Kev-Roche 

Can you look at that PR and the comments from @thomaspaulb ?
Do you want me to make a PR on your branch instead ?

PR #646, opened 2024-06-18
[14.0][FIX] purchase_stock_inter_company: support 'no backorder' case

thomaspaulb:
When an incomplete sale order is delivered and "No backorder" is chosen, SO moves split in two while PO stays the same.

Aggregating writes per each PO move makes sure qty does not get overwritten.

Usually second qty is zero so the destination move becomes quantity zero and the picking was even cancelled.

Sad flow result on runboat 14.0:

image

### DIFF
--- a/purchase_sale_inter_company/models/stock_picking.py
+++ b/purchase_sale_inter_company/models/stock_picking.py
@@ -129,6 +129,7 @@ class StockPicking(models.Model):
             raise UserError(_("PO does not exist or has no receipts"))
         if self.intercompany_picking_id:
             dest_picking = self.intercompany_picking_id.with_user(intercompany_user.id)
+            dest_move_qty_update_dict = {}
             for move in self.move_ids_without_package.sudo():
                 # To identify the correct move to write to,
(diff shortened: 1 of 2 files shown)


christian-ramos-tecnativa (approved):
Functional and code review. LGTM!

HekkiMelody (approved):
Code review, LGTM

HekkiMelody:
Thanks @thomaspaulb!

Also hi, fellow red panda enjoyer @Christian-RB :heart: 

/ocabot merge patch

francesco-ooops:
@thomaspaulb can you try fw-porting this to 15 and 16? I am afraid the diff between versions may increase too much. Thanks!

thomaspaulb:
Thanks for the review and merge, everyone.

@francesco-ooops I will, although it seems to be a big work, will need to block some time for that

francesco-ooops:
@thomaspaulb yes, I'm getting convinced that the burden of fw-porting for those working on older versions should be shared with the burden of bk-porting for those working on newer versions, otherwise that's the situation we end up in.

PR #627, opened 2024-05-22
[16.0][ADD] partner_x_multicompany_default

yvaucher:
Adds new modules to manage property fields on partner accross companies.

Refactor of partner_account_multicompany_default to extract the base methods in partner_base_multicompany_default

Following https://github.com/OCA/multi-company/pull/615

PR #648, opened 2024-06-20
[16.0][ADD] partner_account_payment_mode_multicompany_default

santostelmo:
This module extends the functionality of partner and account_payment_partner to allow you
to propagate multi-company partner account payment modes from the
current company to all other companies where you have access rights.

Depends: https://github.com/OCA/multi-company/pull/627


partner_account_payment_mode_multicompany_default/models/partner.py
```diff
@@ -0,0 +1,34 @@
...
+        return "name"
+
+    def propagate_multicompany_customer_payment_mode(self):
+        self._propagate_multicompany_field("customer_payment_mode_id")
+
+    def propagate_multicompany_supplier_payment_mode(self):
+        self._propagate_multicompany_field("supplier_payment_mode_id")
+
+    def _propagate_property_fields(self):
+        """Propagate accounts to other companies always."""
```
cyrilmanuel, line 19: perhaps add on the comment that is Overwrite to propagate on the parent (if it's not i suppose it can be remove)

cyrilmanuel (approved):
not an expert but LGTM

partner_account_payment_mode_multicompany_default/models/partner.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2024 Camptocamp SA
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+from odoo import models
+
+
+class Partner(models.Model):
+    _inherit = "res.partner"
+
+    def _get_multicompany_account_payment_mode_key(self):
+        return "name"
```
cyrilmanuel, line 10: Ask : shall return the "Code" ?

santostelmo: Payment mode is mapped by name inter-company. There's no code field

partner_account_payment_mode_multicompany_default/models/partner.py
santostelmo: Wrong string. Actually this is about payment_mode and not accounts

cyrilmanuel (approved):

PR #649, opened 2024-06-20
[16.0][IMP] partner_account_multicompany_default map fiscal position by code

santostelmo:
Often between companies, fiscal positions do not share the same name.
In this improvement we've added a code field that allows you to map fiscal positions between companies.


partner_account_multicompany_default/views/partner_view.xml
```diff
@@ -82,4 +82,14 @@
         </field>
     </record>
 
+    <record id="view_account_position_form" model="ir.ui.view">
+        <field name="name">account.fiscal.position.form.inherit</field>
+        <field name="model">account.fiscal.position</field>
+        <field name="inherit_id" ref="account.view_account_position_form" />
```
cyrilmanuel, line 88: ask : tree view and search view not need to be update with the code ?

cyrilmanuel (approved):

partner_account_multicompany_default/views/partner_view.xml
santostelmo: I think it's fine to show only in the form. But it can be an improvement indeed

leemannd (approved):

gurneyalex (approved):

PR #655, opened 2024-07-03
[FIX] purchase_sale_stock_inter_company: default warehouse from same …

cyrilmanuel:
…partner_id

standard default warehouse will apply based on company dest.

In our case, the company has multi warehouses ) we need to apply the right warehouse as default warehouse base on partner_id and belong the dest_company.

### DIFF
--- a/purchase_sale_stock_inter_company/models/purchase_order.py
+++ b/purchase_sale_stock_inter_company/models/purchase_order.py
@@ -22,11 +22,18 @@ class PurchaseOrder(models.Model):
         )
         if delivery_address:
             new_order.update({"partner_shipping_id": delivery_address.id})
+
+        potential_warehouse = self.env["stock.warehouse"].search(
+            [("partner_id", "=", self.partner_id.id)]
+        )
+
         warehouse = (
-            dest_company.warehouse_id.company_id == dest_company
+            potential_warehouse
+            or dest_company.warehouse_id.company_id == dest_company
             and dest_company.warehouse_id
             or False
         )
+
         if warehouse:
             new_order.update({"warehouse_id": warehouse.id})
         return new_order


purchase_sale_stock_inter_company/models/purchase_order.py
```diff
@@ -22,11 +22,18 @@ def _prepare_sale_order_data(
         )
         if delivery_address:
             new_order.update({"partner_shipping_id": delivery_address.id})
+
+        potential_warehouse = self.env["stock.warehouse"].search(
+            [("partner_id", "=", self.partner_id.id)]
+        )
```
jbaudoux, line 28: To prevent singleton error:
```suggestion
        limit=1)
```

jbaudoux (changes requested):
LG, just a small remark

purchase_sale_stock_inter_company/models/purchase_order.py
```diff
@@ -22,11 +22,18 @@ def _prepare_sale_order_data(
         )
         if delivery_address:
             new_order.update({"partner_shipping_id": delivery_address.id})
+
+        potential_warehouse = self.env["stock.warehouse"].search(
+            [("partner_id", "=", self.partner_id.id)]
```
jbaudoux, line 27: You forgot the last part of your PR description (belongs to dest company)
```suggestion
            [
            ("partner_id", "=", self.partner_id.id),
            ("company_id", "=", dest_company.id),
            ]
```

jbaudoux (changes requested):

jbaudoux (commented):
Can you add the unit test for the case you are fixing?

jbaudoux:
@cyrilmanuel Where do you stand with this fix? Would be nice to integrate it

cyrilmanuel:
superseeded at https://github.com/OCA/multi-company/pull/948

PR #662, opened 2024-07-11
[FIX] purchase_sale_inter_company: support discount

thomaspaulb:
Currently when the sale order line has received a discount, and the SO is confirmed, the undiscounted price is copied over to the PO, leading to a price discrepancy between the SO and PO. With this commit, discount is respected.


HekkiMelody (approved):
Code review, LGTM

PR #665, opened 2024-07-16
[16.0][MIG] crm_tag_multi_company: Migration to 16.0

julienCozo:
Migration from 12.0.
Model change, so module name change from crm_lead_tag_multi_company to crm_tag_multi_company


remi-filament (changes requested):
Thanks for this migration @julienCozo 
You should rename the PR as follow : [16.0][MIG] crm_tag_multi_company: Migration to 16.0 
Apart from this, tests are failing because of another module which I try to fix here : #668

theo-le-filament (approved):

remi-filament (approved):
Tested OK

Issue #182, opened 2020-01-07
When a user switches from one company to another the companies of the related partner are rewritten in stead of amended (partner_multi_company)   


gbeuning:
Hi,

I'm experiencing some issues with `partner_multi_company` when a user switches from one company to another.

**Use case:**

- User 1 has access to both company A and B.
- The related partner of this user also has access to both company A and B. 
- This user switches from company A to company B.

This results in a `res_users` write action (because `res_users.company_id` is changed). The companies (`company_ids`) of the related partner are now overwritten, containing company B only.
This may lead to unexpected results such as user 1 not showing up anymore in the list of users in the UI when this list is shown for user 2 with active company A.

**Odoo version:**
Tested with Odoo v12 CE, `partner_multi_company` v12.0.1.0.0, `base_multi_company` v12.0.1.0.1

**Proposed solution:**
Adapt `write(self, vals)` in res_users.py 

- Append to `user.partner_id.company_ids` in stead of replacing the list of linked records.
- Take new `visible_for_all_companies` into account.
- It would probably also be better to keep the companies for the related partner synchronized with `res_users.company_ids`.

Something like:
```
    @api.multi
    def write(self, vals):
        res = super(ResUsers, self).write(vals)
        if 'company_id' in vals:
            for user in self.sudo():
                if user.partner_id and not(user.partner_id.visible_for_all_companies) and user.partner_id.company_ids:
                    user.partner_id.company_ids = [(4, vals['company_id'])]       
        if 'company_ids' in vals:
            for user in self.sudo():
                if user.partner_id and not(user.partner_id.visible_for_all_companies) and user.partner_id.company_ids:
                    user.partner_id.company_ids = [(6, 0, vals['company_ids'][0][2])]
        return res
```

JordiBForgeFlow:
I agree with this. @MiquelRForgeFlow can you tackle this one? cc @LoisRForgeFlow

MiquelRForgeFlow:
Done in https://github.com/OCA/multi-company/pull/190

MiquelRForgeFlow:
Can be closed.

PR #667, opened 2024-07-16
[15.0][IMP] partner_multi_company: prevent inconsistencies between user and partner

ArnauCForgeFlow:
This improvement prevent some inconsistencies between user and partner company_ids

- If you try to modify the partner companies and the user companies are not included in the partner companies,  an error will be raised.
- If the user's companies are modified, they will be added to the partner while retaining the companies the partner already has.
- Don't delete all partner company_ids when the user's related company_id is modified.

The change on the res_company_code was needed in order to pass odoo tests

### DIFF
--- a/partner_multi_company/migrations/15.0.1.1.0/post-migration.py
+++ b/partner_multi_company/migrations/15.0.1.1.0/post-migration.py
@@ -1,0 +1,10 @@
+# Copyright 2024 ForgeFlow S.L. (http://www.forgeflow.com)
+# Copyright 2024 ForgeFlow S.L. (http://www.forgeflow.com)
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+# pylint: disable=W7950
+from odoo.addons.partner_multi_company import hooks
+
+
+def migrate(cr, version):
+    hooks.fix_user_partner_companies(cr)
--- a/partner_multi_company/models/res_users.py
+++ b/partner_multi_company/models/res_users.py
@@ -12,16 +12,28 @@ class ResUsers(models.Model):
         res = super(ResUsers, self).create(vals)
         if "company_ids" in vals:
             res.partner_id.company_ids = vals["company_ids"]
-        if "company_id" in vals and res.partner_id.company_ids:
-            res.partner_id.company_id = vals["company_id"]
         return res
 
     def write(self, vals):
-        res = super(ResUsers, self).write(vals)
+        res = super(ResUsers, self.with_context(from_res_users=True)).write(vals)
         if "company_ids" in vals:
             for user in self.sudo():
-                if user.partner_id.company_ids:
-                    user.partner_id.company_ids = vals["company_ids"]
+                new_company_ids = []
+                company_ids_data = vals["company_ids"]
+                if isinstance(company_ids_data, list) and user.partner_id.company_ids:
+                    for item in company_ids_data:
+                        if isinstance(item, tuple):
+                            if item[0] == 4:
+                                new_company_ids.append(item[1])
+                            elif item[0] == 6:
+                                new_company_ids.extend(item[2])
+                        elif isinstance(item, list):
+                            new_company_ids = item[2]
+                        else:
+                            new_company_ids.append(item)
+                    user.partner_id.company_ids = [
+                        (4, company_id) for company_id in new_company_ids
+                    ]
         if "company_id" in vals:
             for user in self.sudo():
                 if user.partner_id.company_ids:
--- a/partner_multi_company/__manifest__.py
+++ b/partner_multi_company/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Partner multi-company",
     "summary": "Select individually the partner visibility on each company",
-    "version": "15.0.1.0.0",
+    "version": "15.0.1.1.0",
     "license": "AGPL-3",
     "depends": ["base_multi_company"],
     "author": "Tecnativa, " "Odoo Community Association (OCA)",
--- a/partner_multi_company/hooks.py
+++ b/partner_multi_company/hooks.py
@@ -16,6 +16,19 @@ def post_init_hook(cr, registry):
         "base.res_partner_rule",
         "res.partner",
     )
+    fix_user_partner_companies(cr)
+
+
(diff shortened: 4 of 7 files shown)


LoisRForgeFlow (commented):
@ArnauCForgeFlow  Functional review

There is an issue when creating a new company, the constraint raises and it is blocked:

image

LoisRForgeFlow (approved):
After discussion, the problem is only due to demo data coming from different modules. It is a status impossible to reach in a blank database.

Therefore, LGTM :+1:

partner_multi_company/migrations/15.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2024 ForgeFlow S.L. (http://www.forgeflow.com)
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo import SUPERUSER_ID, api
+
+
+def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    for user in env["res.users"].search([]):
```
pedrobaeza: Don't duplicate code. Just import with `from ...hooks import post_init_hook` and call it here.

partner_multi_company/models/res_users.py
```diff
@@ -12,16 +12,28 @@ def create(self, vals):
         res = super(ResUsers, self).create(vals)
         if "company_ids" in vals:
             res.partner_id.company_ids = vals["company_ids"]
-        if "company_id" in vals and res.partner_id.company_ids:
```
pedrobaeza, line 15: Why removing this?

partner_multi_company/models/res_users.py
```diff
@@ -12,16 +12,28 @@ def create(self, vals):
...
         return res
 
     def write(self, vals):
-        res = super(ResUsers, self).write(vals)
+        res = super(ResUsers, self.with_context(from_res_users=True)).write(vals)
         if "company_ids" in vals:
             for user in self.sudo():
-                if user.partner_id.company_ids:
-                    user.partner_id.company_ids = vals["company_ids"]
+                new_company_ids = []
```
pedrobaeza, line 21: Please explain this change of behavior.

pedrobaeza (changes requested):
Standard doesn't put any company to the partners of the users. Why changing the strategy?

partner_multi_company/migrations/15.0.1.1.0/post-migration.py
LoisRForgeFlow: Good remark

cc @ArnauCForgeFlow

partner_multi_company/models/res_users.py
LoisRForgeFlow: Because this was totally legacy, from the time in which the active company was self.env.user.company_id, since v13 this doesn't make sense anymore as self.env.user.company_id is the default company that you should not touch much, and the environment company is what matters (self.env.company)

Ref https://github.com/OCA/multi-company/issues/182

partner_multi_company/models/res_users.py
LoisRForgeFlow: The idea here is that you add allowed companies to an user, the underlying related partner gets updated if needed.

Before this you can have a user with companies A,B and the related partner with companies A,B and then add company C to the user. This generates an inconsistency as related partner should also have company C, otherwise multicompany rules will raise in different siutations.

pedrobaeza: But as said, Odoo by default don't assign any company to any user's partner, so I think that should keep the same.

partner_multi_company/models/res_users.py
pedrobaeza: Well, not really, as this is only on the `create` method, so switching between companies doesn't affect this, but well, with the other reasoning of keeping empty the companies on the partner (if we go there), this should be removed indeed.

partner_multi_company/models/res_users.py
LoisRForgeFlow: If the partner has no companies, this does nothing. If partner has no companies, nothing will be assigned.

partner_multi_company/models/res_users.py
LoisRForgeFlow: Yes, I see what you say and I agree that partner companies should be kept empty. Actually the meaning of company_id in user is not exactly the same as in other models since v13.

partner_multi_company/migrations/15.0.1.1.0/post-migration.py
ArnauCForgeFlow: Done!

LoisRForgeFlow:
@pedrobaeza let us know if it looks good to you now

LoisRForgeFlow:
Hi @pedrobaeza 

Does it looks better now? Could you update your review?

Thanks!

partner_multi_company/migrations/15.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 ForgeFlow S.L. (http://www.forgeflow.com)
```
pedrobaeza, line 1: Double copyright

partner_multi_company/migrations/15.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,10 @@
...
+# Copyright 2024 ForgeFlow S.L. (http://www.forgeflow.com)
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+# pylint: disable=W7950
+from odoo.addons.partner_multi_company import hooks
```
pedrobaeza, line 6:
```suggestion
from ...hooks import fix_user_partner_companies
```

and change the call

pedrobaeza (approved):
/ocabot merge nobump

Please fw-port it to upper versions.

partner_multi_company/migrations/15.0.1.1.0/post-migration.py
pedrobaeza: This is not critical, so I launched the merge command.

PR #670, opened 2024-07-23
[16.0][IMP] partner_multi_company: prevent inconsistencies between user and partner

ArnauCForgeFlow:
fwport https://github.com/OCA/multi-company/pull/667

### DIFF
--- a/partner_multi_company/migrations/16.0.1.1.0/post-migration.py
+++ b/partner_multi_company/migrations/16.0.1.1.0/post-migration.py
@@ -1,0 +1,9 @@
+# Copyright 2024 ForgeFlow S.L. (http://www.forgeflow.com)
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+# pylint: disable=W8150
+from odoo.addons.partner_multi_company import hooks
+
+
+def migrate(cr, version):
+    hooks.fix_user_partner_companies(cr)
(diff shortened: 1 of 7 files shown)


LoisRForgeFlow (approved):
Ty

LoisRForgeFlow:
cc @pedrobaeza

partner_multi_company/migrations/16.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2024 ForgeFlow S.L. (http://www.forgeflow.com)
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+# pylint: disable=W8150
+from odoo.addons.partner_multi_company import hooks
```
pedrobaeza, line 5: We can do it here correctly:

```suggestion
from ...hooks import fix_user_partner_companies
```

ArnauCForgeFlow: Hello @pedrobaeza ,

I attempted to use the following import statement: `from ...hooks import fix_user_partner_companies`, but I encountered the following error:
`ImportError: attempted relative import with no known parent package`.

I reviewed other modules that utilize functions from hooks in their migration scripts, and they all employ absolute imports, disabling pylint. Perhaps you have encountered this error before and have found a solution?

pedrobaeza: OK, don't worry. Let's go this way. Maybe the migration manager is not doing imports 100% Python compatible.

pedrobaeza (approved):
/ocabot merge nobump

PR #671, opened 2024-07-23
[15.0][FIX] product_multi_company: do not allow to remove company if there is stock or moves in that company

JoanSForgeFlow:
This fix addresses the issue where companies could be removed from a product's company_ids field even when there are existing stock quantities or stock moves associated with that company. The update adds validation checks to prevent the removal of a company under these conditions, ensuring data integrity and preventing potential inconsistencies


BernatPForgeFlow (commented):
Shouldn't the check for stock or stock moves be in the write and the unlink?

pedrobaeza (changes requested):
The addition of the `stock` dependency in this module wasn't correct in the migration (blaming @AaronHForgeFlow), and must be removed, so don't add more things coupled to stock. You must do it in a new module `product_multi_company_stock`.

PR #672, opened 2024-07-23
[13.0][FIX] purchase_sale_inter_company: Use the correct function name to link the invoice with purchase order.

carlos-lopez-tecnativa:
Before this commit, when an invoice was created, it was never linked with the  purchase order because the function name was incorrect After this commit, the invoice and purchase are linked correctly see function name here https://github.com/OCA/multi-company/blob/473aa18b9286dcd8650d7e51dcdb07295e972aac/account_invoice_inter_company/models/account_move.py#L82

@Tecnativa TT46194
@carolinafernandez-tecnativa  @pedrobaeza could you please review this PR

### DIFF
--- a/purchase_sale_inter_company/models/account_move.py
+++ b/purchase_sale_inter_company/models/account_move.py
@@ -9,13 +9,9 @@ from odoo import _, models
 class AccountMove(models.Model):
     _inherit = "account.move"
 
-    def inter_company_create_invoice(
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

carolinafernandez-tecnativa (approved):
LGTM Thanks :)

PR #673, opened 2024-07-24
[17.0][MIG] product_tax_multicompany_default: Migration to version 17.0

carlos-lopez-tecnativa:
@Tecnativa TT49882

NOTE: It is necessary to use `sudo` in test because the `account.group_account_manager` group does not have permission to create products. See https://github.com/odoo/odoo/commit/512574861691f425ec6a17f20fe4b586bb88a299#r144593878


carolinafernandez-tecnativa:
I have tested and check  this example:

1- I have 2 companies
2- Select with user two companies
3- Create new product selecting two companies with user.

Taxes for first company is added twice
image

If i edit product then it is fixed.

Is it okay this behaviour?

carlos-lopez-tecnativa:
@carolinafernandez-tecnativa could you please test again?, thanks

carolinafernandez-tecnativa (approved):
Functional review LGTM :) Thanks!

product_tax_multicompany_default/tests/test_product_tax_multicompany.py
```diff
@@ -101,27 +120,35 @@ def test_multicompany_default_tax(self):
...
-        product = self.env["product.product"].create(
-            {
-                "name": "Test Product",
-                "taxes_id": [(6, 0, self.tax_20_cc1.ids)],
-                "supplier_taxes_id": [(6, 0, self.tax_20_sc1.ids)],
-                "company_id": False,
-            }
+        product = (
+            self.env["product.product"]
+            .sudo()
```
pedrobaeza, line 128: Why do you need the sudo? It shouldn't be needed.

carlos-lopez-tecnativa: @pedrobaeza In the PR description, I justified these changes. Basically, the `account.group_account_manager` group does not have permission to create products.

I also added comments in the code to clarify and removed some unnecessary `sudo()` calls after create. Let me know if you have any other comments. Thank you.

pedrobaeza (approved):
/ocabot merge nobump

PR #678, opened 2024-08-08
[15.0][IMP] Don't update session_info if display_switch_company_menu unset.

amh-mw:
This adds some defensive code to avoid an error updating session_info values that have not been populated.

### DIFF
--- a/res_company_code/models/ir_http.py
+++ b/res_company_code/models/ir_http.py
@@ -13,24 +13,22 @@ class Http(models.AbstractModel):
         result = super().session_info()
         user = request.env.user
 
-        display_switch_company_menu = (
-            user.has_group("base.group_multi_company") and len(user.company_ids) > 1
-        )
+        if user.has_group("base.group_multi_company") and len(user.company_ids) > 1:
+            # Replace company name by company complete name in the session
+            # The values are used in the switch_company_menu widget (web module)
+            result["user_companies"]["current_company"] = user.company_id.id
 
-        # Replace company name by company complete name in the session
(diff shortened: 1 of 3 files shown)


res_company_code/models/ir_http.py
```diff
@@ -13,24 +13,22 @@ def session_info(self):
         result = super().session_info()
         user = request.env.user
 
-        display_switch_company_menu = (
-            user.has_group("base.group_multi_company") and len(user.company_ids) > 1
-        )
+        if result.get("display_switch_company_menu"):
```
amh-mw: DRY https://github.com/odoo/odoo/blob/15.0/addons/web/models/ir_http.py#L93

res_company_code/models/ir_http.py
```diff
@@ -13,24 +13,22 @@ def session_info(self):
...
+            # Replace company name by company complete name in the session
+            # The values are used in the switch_company_menu widget (web module)
+            result["user_companies"]["current_company"] = user.company_id.id
```
amh-mw, line 19: This is the line that errors. Reproducible running stock odoo unit tests with `--init mail,res_company_code`. The `user_companies` value is not set for users without group `base.group_user`.

amh-mw: Covered with unit tests.

christian-ramos-tecnativa (approved):
Make sense, LGTM

amh-mw:
@OCA/intercompany-maintainers Only one approval, but merge?

PR #652, opened 2024-06-25
[16.0][FIX] account_invoice_inter_company : partner_shipping_id shouldn't b…

metaminux:
…e propagated to supplier invoices

  fixes OCA/multi-company#651

Issue #651, opened 2024-06-25
[16.0] account_invoice_inter_company : partner_shipping_id propagation leads to wrong fiscal position


metaminux:
I think we shouldn't propagate the **partner_shipping_address** field in supplier invoices

## Module
account_invoice_inter_company

## Describe the bug
As of 44a3178f365bc9ec6148f4eeec4c88af23d8be36 the **partner_shipping_id** field is propagated from the customer invoice of company A to the inter company Vendor Bill of company B
But shipping address is used in invoices to compute the fiscal position (see https://github.com/odoo/odoo/blob/54f84aedf8da14f4f3298febc44c4fec103e3a47/addons/account/models/account_move.py#L773-L781), leading to wrong tax rates

## To Reproduce
**16.0**:

1. Create a brand new database with demo data
2. Install module **account_invoice_inter_company**
3. Install module **sale** to be sure partner_shipping_id field exists in invoice
4. Install module **l10n_fr** with demo data (to get a french company with fiscal position and taxes already set)
5. Select the main company _YourCompany_
6. Create a customer invoice with **FR Company** as customer
7. Add a product (for example E-COM07, qty = 1, unit price $320)
8. Confirm the invoice

**Current behavior**
The generated inter company Supplier Bill from _YourCompany (San Francisco)_ in _FR Company_ company has a fiscal position **EU privé** (which is wrongly computed from _FR Company partner_ in **partner_shipping_id** field in the _FR Company company_ context)
This leads to a VAT 20% rate and a total amount of $384 USD
A message is created in the chatter to inform that the total amount is different from the original inter company invoice...

**Expected behavior**
The generated inter company Supplier Bill from _YourCompany (San Francisco)_ in _FR Company_ company has a fiscal position **Import/Export Hors Europe + DOM-TOM** (which is the right position for exportation from US to France)
This leads to a Export 0% rate and a total amount of $320 USD

**Additional context**
- by default, tax rate computed in customer invoice for France is 15% which I think is wrong but is not relevant here
- as account_invoice_inter_company doesn't depend on sale, I'm not sure how to proceed to write a relevant test for this use case
- i will create a PR to remove the propagation

metaminux:
cc @Kev-Roche

PR #688, opened 2024-09-02
[16.0][FIX] account_invoice_inter_company : clean_context not working correctly

Kev-Roche:
In some cases,  some default values appear in context (default_partner_shipping_id, ...) and are propagate in the counterpart invoice. 
As discuss here https://github.com/OCA/multi-company/issues/651 and https://github.com/OCA/multi-company/pull/652, we don't want to propagate these values between invoices. 

For example : 
When we open an invoice via the method `action_view_invoice`  from a sale or purchase order. 

Also, calling `clean_context` with ** not working, and without ** precommit fails. 

cc @pedrobaeza , @metaminux

### DIFF
--- a/account_invoice_inter_company/models/account_move.py
+++ b/account_invoice_inter_company/models/account_move.py
@@ -135,6 +135,7 @@ class AccountMove(models.Model):
             force_number = inter_invoice.name
             inter_invoice.with_context(force_delete=True).unlink()
         # create destination invoice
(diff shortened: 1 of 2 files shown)


emiliesoutiras (approved):

emiliesoutiras:
Hi @Kev-Roche, please can you fix pre-commit ?

PR #700, opened 2024-09-25
[16.0][ADD] check_current_company_for_validation module

Kev-Roche:
Small mixin checkin if the current company is the same as the record company. 
It's come with the check on sale.order


check_current_company_for_validation/models/check_current_company_mixin.py
```diff
@@ -0,0 +1,16 @@
...
+from odoo import _, models
+from odoo.exceptions import UserError
+
+
+class CheckCurrentCompanyMixin(models.AbstractModel):
+    _name = "check.current.company.mixin"
+    _description = "Check Current Company Mixin"
+
+    def check_current_company(self):
+        if self.env.company != self.company_id:
```
gbrito, line 14: This can fail if you have a recordset with different companies
```suggestion
        if self.filtered(lambda x: x.company_id != self.env.company)
```

or loop through all the recordset

gbrito (changes requested):

PR #701, opened 2024-09-27
[16] account_invoice_inter_company : Ensure supplier reference is set in case the source invoice is the supplier one

florian-dacosta:
@bguillot @sebastienbeau @bealdav 

Same as https://github.com/OCA/multi-company/pull/558
I will port it to v17/18 when/if it is merged.

### DIFF
--- a/account_invoice_inter_company/models/account_move.py
+++ b/account_invoice_inter_company/models/account_move.py
@@ -40,6 +40,12 @@ class AccountMove(models.Model):
         )
         return company or False
 
+    def _set_intercompany_supplier_invoice_ref(self):
+        self.ensure_one()
+        supplier_invoice = self.auto_invoice_id
+        if not supplier_invoice.ref:
+            supplier_invoice.write({"ref": self.name})
+
     def action_post(self):
         """Validated invoice generate cross invoice base on company rules"""
         res = super().action_post()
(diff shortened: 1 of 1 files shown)


Kev-Roche (approved):
LGTM

bguillot (approved):

pedrobaeza (commented):
I see something weird: or it shouldn't be `is_purchase_document` check, or what you should pass is the auto generated invoice for setting there the reference.

florian-dacosta:
> I see something weird: or it shouldn't be `is_purchase_document` check

It is `is_sale_document` not `is_purchase_document`, I don't get your remark.

The case is the following.
Validation of a supplier (purchase) invoice.
The intercompany customer (sale) invoice is then generated. and is flaggued `auto_generated`.
Then in this case, we want to write on the source (the supplier) invoice, the reference of the customer invoice.
So in order to check if we are in this specific case, we check that this is a "sale document" and that it is "auto-generated".

We have to do that afterward because when the supplier invoice is created, the customer one does not exists yet (so it does not have any number yet)

pedrobaeza:
But the flag `auto_generated` is set on the supplier invoice, not on the customer one, isn't it? If so, you are doing the `filtered` incorrectly.

florian-dacosta:
> But the flag `auto_generated` is set on the supplier invoice, not on the customer one, isn't it? If so, you are doing the `filtered` incorrectly.

No, in this case, the flag auto_generated is set on the customer invoice!

There are 2 cases with this module
1) Most common case : the customer invoice is created first and generate the supplier invoice. In this case, the supplier invoice is `auto_generated`
2) The contrary : the supplier invoice is created first and generate the customer invoice. In this less common case, the customer invoice is `auto_generated`.
This is the case we manage here

pedrobaeza:
That's totally weird I have to say. In fact, I have never understood why the code is generating a customer invoice from a vendor bill, but OK...

/ocabot merge patch

PR #708, opened 2024-10-21
[18.0][MIG] base_multi_company: Migration to 18.0

BertVGroenendael:


base_multi_company/models/multi_company_abstract.py
```diff
@@ -0,0 +1,109 @@
...
+    @api.model
+    def _search_display_name(self, operator="ilike", value=dict):
+        domain = super()._search_display_name(operator=operator, value=value)
+        new_domain = self._patch_company_domain(domain)
+        return new_domain
+
+    @api.model
+    def search_read(self, domain=None, fields=None, offset=0, limit=None, order=None):
+        new_domain = self._patch_company_domain(domain)
+        return super().search_read(new_domain, fields, offset, limit, order)
```
gbrito:
```suggestion
        return super().search_read(domain=new_domain, fields=fields, offset=offset, limit=limit, order=order)
```

gbrito (changes requested):

victoralmau:
Please cherry-pick https://github.com/OCA/multi-company/pull/719 to commit history

woutswennen:
Opening the register in PoS gave an error as it will give extra kwargs that where not expected 
640647f fixes that

yajo (changes requested):
There's a refactor in https://github.com/OCA/multi-company/pull/760 that deprecates a lot of stuff. I think this migration is the perfect moment to remove all those things. Would you please include it (and then an extra commit to remove all that's deprecated there?). I could PR your branch if you prefer; just tell me.

PieterPaulussen:
Superseeding this PR to include various fixes that have been added to V17 later on.
https://github.com/OCA/multi-company/pull/790

PR #712, opened 2024-10-24
Add partner_base_multicompany_default_queue

leemannd:
Glue module inbetween `partner_base_multicompany` and `queue_job`.

The propagation of the fields can be costly given the amount of fields to be propagated and the amount of companies. By default the base module will make a write for each field to be propagated on each companies. With 4 fields and 4 companies it will trigger 16 writes which makes unusefull slownesses in the UI.
With this module, we want to make the propagation of the fields asynchronous with the creation of queue_jobs.


partner_base_multicompany_default_queue/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 Camptocamp SA
```
simahawk, line 1: I would rename it to `*_multicompany_queue` as you might add other feats other than setting defaults.

partner_base_multicompany_default_queue/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+
+{
+    "name": "Partner Base Multi-Company Default Queue",
+    "summary": "Glue module between partner_base_multicompany and queue_job",
+    "version": "16.0.0.1.0",
+    "development_status": "Alpha",
+    "category": "Technical",
+    "website": "https://github.com/OCA/multi-company",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "maintainers": ["camptocamp"],
```
simahawk, line 12: Maintainers have to be persons

partner_base_multicompany_default_queue/models/base.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Base(models.AbstractModel):
+
+    _inherit = "base"
+
+    @api.model
+    def _job_prepare_context_before_enqueue_keys(self):
+        """Keys to keep in context of stored jobs
+        Empty by default for backward compatibility.
+        """
```
simahawk, line 15:
```suggestion
```

simahawk (commented):
any description on what this module is trying to solve?

partner_base_multicompany_default_queue/__manifest__.py
simahawk: Nerver mind, I see where it comes from.

partner_base_multicompany_default_queue/models/base.py
simahawk: also, this override should be moved to res.partner

partner_base_multicompany_default_queue/models/partner.py
```diff
@@ -0,0 +1,23 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+
+from odoo import models
+
+
+class Partner(models.Model):
+
+    _inherit = "res.partner"
+
+    def _propagate_property_fields(self):
```
simahawk, line 11: you miss a queue.job.function to be able to delay this

simahawk (changes requested):

PR #715, opened 2024-11-18
[15.0][FIX] base_multi_company: search with in operator

chienandalu:

cc @Tecnativa TT51779

please review @pedrobaeza @carlosdauden


pedrobaeza (approved):
Can you please add a regression test?

carlosdauden (approved):

victoralmau (approved):

chienandalu:
> Can you please add a regression test?

Done

PR #717, opened 2024-11-19
[16.0][FIX] base_multi_company: search with in operator

victoralmau:
FWP from 15.0: https://github.com/OCA/multi-company/pull/715

Please @chienandalu and @pedrobaeza can you review it?

@Tecnativa TT51779


chienandalu (commented):
I think you slipped an extra commit

victoralmau:
The extra commit needed to correct test errors in another module (not related to this change).

chienandalu (approved):
Ok :+1:

pedrobaeza (approved):
/ocabot merge patch

PR #719, opened 2024-11-20
[17.0][FIX] base_multi_company: search with in operator

victoralmau:
FWP from 16.0: https://github.com/OCA/multi-company/pull/717

Please @chienandalu and @pedrobaeza can you review it?

@Tecnativa TT51779


pedrobaeza:
Better to squash both commits, as the second one is the slight adaptation to 17.0

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #720, opened 2024-11-20
[16.0][FIX] product_multi_company: search in False

chienandalu:
fw of 

- #716 

product.product has a delageted inheritance from product.template so we are getting the fields logic but not the ORM logic. We need some of that logic in the search method to be able to get the right results with domains like [("company_id", "in", [1, False]) to include records which are shared between companies.

cc @Tecnativa TT51779

please review @pedrobaeza @victoralmau


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #732, opened 2024-11-29
[FIX] stock_intercompany_bidirectional: Fix incorrect move warehouse

GabbasovDinar:
Before this commit
==============

After moving from company A to company B, a route with a push rule for company B does not work because the stock move retains the warehouse reference from company A

After this commit
=============

Triggers push rule and creates a picking between locations in company B, the link to the warehouse for the stock move is updated


Bearnard21 (approved):
Functionality LGTM

geomer198 (approved):
LGTM!

hoangtrann (approved):
Code LGTM

PR #733, opened 2024-11-29
[13.0][FIX] purchase_sale_stock_inter_company: Allow reception with serial tracking and partial quantity (Dropshipping)

carlos-lopez-tecnativa:
Issue: When handling intercompany dropshipping with serialized products, an error is raised during reception if the quantity received is less than the original purchase order's quantity. The error message states: "Mismatch between move lines with the corresponding."

**Steps to Reproduce:**

**Settings:**

In General Settings (both companies):
- Enable Sales from Purchase Orders.
- Disable Sales Order Auto Validation.
- In Purchase Settings: Enable Dropshipping.
- In Inventory Settings: Enable Multi-Step Routes.

image

**Setup:**

In Company 1:
Configure a product with vendor information linked to Company 2.
In Company 2:
Configure the same product with vendor information linked to any other vendor (not a company contact).
Scenario:

In Company 1:

- Create a sales order with a line using the Dropshipping route and the configured product.
- Confirm the sales order.
- Navigate to Purchase Orders, find the related record, and confirm it.

Switch to Company 2:

- Locate the generated sales order and update the sales line to use the Dropshipping route.
- Set the quantity to 2 and confirm the sales order.
- Go to Purchase Orders, find the related record, and confirm it.
- Navigate to Receipts for the purchase order, enter the lot/serial numbers, and attempt to validate the receipt.

Result: An error is raised, preventing the validation of the receipt due to a mismatch between the move lines and their corresponding records.

SO in C1
http://oca-multi-company-13-0-c48386e2793f.runboat.odoo-community.org/web#id=25&action=376&model=sale.order&view_type=form&cids=1&menu_id=224

PO in C1
http://oca-multi-company-13-0-c48386e2793f.runboat.odoo-community.org/web#id=12&action=355&model=purchase.order&view_type=form&cids=&menu_id=206

SO in C2
http://oca-multi-company-13-0-c48386e2793f.runboat.odoo-community.org/web#id=26&action=376&model=sale.order&view_type=form&cids=2&menu_id=224

PO in C2
http://oca-multi-company-13-0-c48386e2793f.runboat.odoo-community.org/web#id=13&action=355&model=purchase.order&view_type=form&cids=&menu_id=206

image

TT51858
@Tecnativa @pedrobaeza could you please review this

### DIFF
--- a/purchase_sale_stock_inter_company/models/stock_move_line.py
+++ b/purchase_sale_stock_inter_company/models/stock_move_line.py
@@ -1,0 +1,99 @@
+from odoo import api, models
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        new_move_lines = super().create(vals_list)
+        for move_line in new_move_lines.filtered(lambda x: x.state == "done"):
+            po_moves = move_line._get_stock_moves_to_sync()
+            for po_move in po_moves:
+                po_move_line_vals = po_move._prepare_move_line_vals(
+                    quantity=move_line.qty_done
+                )
+                if move_line.lot_id:
+                    dest_lot = move_line._get_or_create_lot_intercompany(
+                        po_move.company_id
+                    )
+                    po_move_line_vals["lot_id"] = dest_lot.id
+                po_move_line_vals["qty_done"] = move_line.qty_done
+                self.sudo().create(po_move_line_vals)
+        return new_move_lines
+
+    def write(self, vals):
+        moves_to_sync = {}
+        fields_to_sync = self._get_fields_to_sync_intercompany()
+        if fields_to_sync.intersection(set(vals.keys())):
+            for move_line in self:
+                purchase_line_origin = (
+                    move_line.move_id.sale_line_id.sudo().auto_purchase_line_id
+                )
+                if move_line.state == "done" and purchase_line_origin:
+                    moves_to_sync.setdefault(move_line, move_line.lot_id.name or "")
+        res = super().write(vals)
+        for move_line, lot_name in moves_to_sync.items():
+            move_line._sync_intercompany_move(lot_name, vals)
+        return res
+
+    def _sync_intercompany_move(self, lot_name, vals):
+        self.ensure_one()
+        fields_to_sync = self._get_fields_to_sync_intercompany()
+        po_moves = self._get_stock_moves_to_sync()
+        for po_move in po_moves:
+            po_move_line = po_move.move_line_ids.filtered(
+                lambda x: not self.lot_id or x.lot_id.name == lot_name
+            )
+            if not po_move_line:
+                continue
+            vals_to_write = {}
+            for field in fields_to_sync:
+                if field not in vals:
+                    continue
+                field_value = self[field]
+                if field == "lot_id" and field_value:
+                    dest_lot = self._get_or_create_lot_intercompany(
+                        po_move_line.company_id
+                    )
+                    field_value = dest_lot.id
+                vals_to_write[field] = field_value
+            if vals_to_write:
+                po_move_line.write(vals_to_write)
+
+    def _get_stock_moves_to_sync(self):
+        """
+        Get the stock moves that need to be synced with the intercompany move.
+        """
+        self.ensure_one()
+        purchase_line_origin = self.move_id.sale_line_id.sudo().auto_purchase_line_id
+        po_moves = self.env["stock.move"]
+        if purchase_line_origin:
+            po_moves = purchase_line_origin.move_ids.filtered(
+                lambda m: m.picking_id
+                == self.move_id.picking_id.intercompany_picking_id
+                and m.product_id == self.product_id
+            )
+        return po_moves
+
+    @api.model
+    def _get_fields_to_sync_intercompany(self):
+        return {"qty_done", "lot_id"}
+
+    def _get_or_create_lot_intercompany(self, dest_company):
+        # search if the same lot exists in destination company
+        self.ensure_one()
+        ProductionLot = self.env["stock.production.lot"].sudo()
+        lot = self.lot_id
+        dest_lot = ProductionLot.search(
+            [
+                ("product_id", "=", lot.product_id.id),
+                ("name", "=", lot.name),
+                ("company_id", "=", dest_company.id),
+            ],
+            limit=1,
+        )
+        if not dest_lot:
+            # if it doesn't exist, create it by copying from original company
+            dest_lot = lot.sudo().copy({"company_id": dest_company.id})
+        return dest_lot
--- a/purchase_sale_stock_inter_company/models/stock_picking.py
+++ b/purchase_sale_stock_inter_company/models/stock_picking.py
@@ -3,14 +3,13 @@
 # Copyright 2023 Tecnativa - Carolina Fernandez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import _, fields, models
-from odoo.exceptions import UserError
+from odoo import fields, models
 
 
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
-    intercompany_picking_id = fields.Many2one(comodel_name="stock.picking")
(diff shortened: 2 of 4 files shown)


purchase_sale_stock_inter_company/models/stock_picking.py
```diff
@@ -19,26 +19,61 @@ def action_done(self):
             purchase = pick.sale_id.auto_purchase_order_id
             if not purchase:
                 continue
-            purchase.picking_ids.write({"intercompany_picking_id": pick.id})
-            if not pick.intercompany_picking_id and purchase.picking_ids[0]:
-                pick.write({"intercompany_picking_id": purchase.picking_ids[0]})
+            po_picking_pending = purchase.picking_ids.filtered(
+                lambda x: x.state not in ["done", "cancel"]
+            )
+            po_picking_pending.write({"intercompany_picking_id": pick.id})
```
pedrobaeza: Although it gets the same result, I consider dot notation cleaner:

```suggestion
            po_picking_pending.intercompany_picking_id = pick.id
```

and why writing this on each loop? It seems a bit weird this structure.

purchase_sale_stock_inter_company/models/stock_picking.py
```diff
@@ -19,26 +19,61 @@ def action_done(self):
...
+                    and x.state not in ["done", "cancel"]
                 )
-                if not len(move_lines) == len(po_move_lines):
-                    raise UserError(
-                        _(
-                            "Mismatch between move lines with the "
-                            "corresponding  PO %s for assigning "
-                            "quantities and lots from %s for product %s"
+                po_move_lines = po_moves_pending.mapped("move_line_ids")
+                if move.product_id.tracking == "serial":
```
pedrobaeza: Why doing a special case depending on the product tracking instead of just mimicking the `stock.move.line`?

purchase_sale_stock_inter_company/models/stock_picking.py
carlos-lopez-tecnativa: Sorry, but I don’t understand. I only adapted the current code, but I don’t know the reasons. I think the action_done method can expect many records, and it’s better to iterate in a for loop.

purchase_sale_stock_inter_company/models/stock_picking.py
carlos-lopez-tecnativa: > Why doing a special case depending on the product tracking instead of just mimicking the `stock.move.line`?

My error occurs only with serial products, so I isolated the error. In other cases, the check is fine, but for serial products, it’s not.

pedrobaeza: I don't think so, as lots may have the same problem if you have more than one lot. As said, you should mimic the `stock.move.line` in one place and the other.

carlos-lopez-tecnativa: Done, please update your review

pedrobaeza (commented):
AFAIK, this change is not changing existing `stock.move.line` to adapt things like the quantity, so it's not a real synchronization like it should be.

carlos-lopez-tecnativa:
> AFAIK, this change is not changing existing `stock.move.line` to adapt things like the quantity, so it's not a real synchronization like it should be.

@pedrobaeza I updated the code to synchronize the `stock.move.line`. Could you review it again? Thanks in advance.

purchase_sale_stock_inter_company/models/stock_move_line.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        new_move_lines = super().create(vals_list)
+        for move_line in new_move_lines:
+            purchase_line_origin = (
+                move_line.move_id.sale_line_id.sudo().auto_purchase_line_id
+            )
+            if move_line.state == "done" and purchase_line_origin:
+                po_moves = purchase_line_origin.move_ids.filtered(
```
pedrobaeza: Is it assured here that only one `stock.move` is selected?

purchase_sale_stock_inter_company/models/stock_move_line.py
```diff
@@ -0,0 +1,95 @@
...
+                    lambda m: m.picking_id
+                    == move_line.move_id.picking_id.intercompany_picking_id
+                )
+                if not po_moves:
+                    continue
+                po_move_line_vals = po_moves._prepare_move_line_vals()
```
pedrobaeza: Why not simply copying this sml values and changing the `lot`? Anyway, I think there's an argument for passing the quantity directly to the method.

purchase_sale_stock_inter_company/models/stock_move_line.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        new_move_lines = super().create(vals_list)
+        for move_line in new_move_lines:
+            purchase_line_origin = (
```
pedrobaeza: Is the direction OK from sales to purchase instead of the reverse?

purchase_sale_stock_inter_company/models/stock_move_line.py
pedrobaeza: And on write, you may need also to create some sml. Why not unifying the code?

purchase_sale_stock_inter_company/models/stock_move_line.py
carlos-lopez-tecnativa: You're right, now the moves are iterated and filtered by product. Thanks for the observation!

purchase_sale_stock_inter_company/models/stock_move_line.py
carlos-lopez-tecnativa: The argument `quantity` is used to populate the `product_uom_qty` field. In this case, I need to populate the `qty_done` field.
[Reference](https://github.com/odoo/odoo/blob/4cc086d330b73514fbc64a7fcf22d8a7a9f1b691/addons/stock/models/stock_move.py#L1105).

In the `write` method, the corresponding `stock.move.line` in the other company is always updated. There's no need to create a new record. Note that this code applies only to moves in the done state.

purchase_sale_stock_inter_company/models/stock_move_line.py
carlos-lopez-tecnativa: The direction is indeed from the sale order to the purchase order.

In this OCA module, the flow typically starts with a purchase order, which creates a new sale order. However, the delivery picking is always directed from the sale order to the purchase order.

In our case, for the customer, the flow begins with a sale order, continues to a purchase order (via Dropshipping), and then, from the purchase order, a new sale order is created according to the intercompany module. Hence, this direction is correct.

purchase_sale_stock_inter_company/models/stock_move_line.py
pedrobaeza: And shouldn't you populate also that field?

purchase_sale_stock_inter_company/models/stock_move_line.py
pedrobaeza: But the use case is not important here. Having MTO doesn't matter. The question is that you should receipt the purchase, and that synchronize the sales, not the contrary.

purchase_sale_stock_inter_company/models/stock_move_line.py
carlos-lopez-tecnativa: Done

purchase_sale_stock_inter_company/models/stock_move_line.py
carlos-lopez-tecnativa: In this case, the user must fill in the lot in the sale order of the company where the delivery is made. If I sync from the purchase to the sale, how can I determine which product lot it is? Sorry, but I don’t understand your point.

pedrobaeza: Yes, my fault, you are right. Delivery picking triggers the synchronization to receipt picking.

pedrobaeza (approved):

victoralmau (approved):
Code review OK

PR #737, opened 2024-12-10
[16.0] 34 purchase_sale_inter_company: Update purchase_order.py

patnal:
if there is product_packaging on PO, add this info to SO (expecting shared products between companies)


metaminux:
Hi @patnal 

I suppose the method **_prepare_sale_order_line_data** should be overridden in **purchase_sale_stock_inter_company** module instead, don't you think ?

PR #753, opened 2025-01-03
[IMP] sale_stock_warehouse_multicompany: security

jbaudoux:
Allow to validate a picking linked to a SO of another company without requiring enabling access to that company

cc @sebalix

### DIFF
--- a/sale_stock_warehouse_multicompany/models/stock_picking.py
+++ b/sale_stock_warehouse_multicompany/models/stock_picking.py
@@ -9,7 +9,9 @@ class StockPicking(models.Model):
 
     def _get_allowed_companies(self):
         return (
-            self.company_id | self.sudo().move_ids.sale_line_id.company_id
+            self.company_id
+            | self.sudo().move_ids.sale_line_id.company_id
+            | self.sudo().sale_id.company_id
         ) & self.env.user.company_ids
 
     def button_validate(self):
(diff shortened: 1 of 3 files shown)


sebalix (approved):

santostelmo (approved):

sale_stock_warehouse_multicompany/models/stock_picking.py
```diff
@@ -0,0 +1,20 @@
...
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _get_allowed_companies(self):
+        return (
+            self.company_id | self.sudo().move_ids.sale_line_id.company_id
+        ) & self.env.user.company_ids
```
santostelmo:
```suggestion
    def _get_allowed_companies(self):
        return (
            self.company_id
            | self.sudo().move_ids.sale_line_id.company_id
            | self.sudo().sale_id.company_id
        ) & self.env.user.company_ids
```

santostelmo (changes requested):

sale_stock_warehouse_multicompany/models/stock_picking.py
santostelmo: We also need to give allow access to `self.sudo().sale_id.company_id`

santostelmo: Or just append the new companies to the ones already set in the context. Better then smash the context

PR #760, opened 2025-01-29
[16.0][FIX] base_multi_company: support `in False` search operators

yajo:
Before this commit, when doing a read with a domain such as [the one used by upstream `base.res_partner_rule`][1], the read failed to find records without company.

This is now fixed in the base module, adding tested support for such operators. The only relevant part of the hooks that were provided to workaround the issue is extracted to a new `post_init_hook`. All other hooks are marked as deprecated.

**Instructions for functional tests:** This refactor is internal and should not be noticed functionally. The modules `product_multi_company` and `partner_multi_company` should work just as always. If so, then this is good.

[1]: https://github.com/odoo/odoo/blob/db072461cddced2a8f65a64fb6d2ddf0dd79b38e/odoo/addons/base/security/base_security.xml#L23

@moduon MT-8863 MT-8873

### DIFF
--- a/base_multi_company/README.rst
+++ b/base_multi_company/README.rst
@@ -84,6 +84,7 @@ Contributors
 * Rodrigo Ferreira <rodrigosferreira91@gmail.com>
 * Florian da Costa <florian.dacosta@akretion.com>
 * Denis Roussel <denis.roussel@acsone.eu>
+* Jairo Llopis (`Moduon <https://www.moduon.team/>`__)
 
 Maintainers
 ~~~~~~~~~~~
--- a/base_multi_company/hooks.py
+++ b/base_multi_company/hooks.py
@@ -1,6 +1,8 @@
 # Copyright 2015-2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
 # Copyright 2017 LasLabs Inc.
 # License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+import warnings
+
 from odoo import SUPERUSER_ID, api
 
 __all__ = [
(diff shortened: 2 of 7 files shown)


EmilioPascual (approved):

edlopen (approved):
Code review here. LGTM!

loida-vm (approved):
Functional review done.

I have verified that both in the Products module
imagen
 and in Contacts
imagen
it is allowed to select the companies we want for sharing.

LGTM!
Thanks @yajo

Shide (approved):

rafaelbn:
This module is already migrated to 17 and the is a PR for 18.

@victoralmau @pedrobaeza @chienandalu could you review please? We can fordward port to all versions.

Thank you! 😄 

Note Odoo 18 migration https://github.com/OCA/multi-company/pull/708

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge major

yajo:
Thanks everyone!

Now that this is merged, please review https://github.com/OCA/multi-company/pull/761 and https://github.com/OCA/multi-company/pull/762, which are the refactors for modules that implement this same behavior on partners and products respectively.

Just like here, the refactor is internal, and if modules just work as always, then it's OK.

victoralmau:
Can you add these changes also to v17 so I can bring them also to v18 in https://github.com/OCA/multi-company/pull/790?

yajo:
I'd love to, but right now I don't think I'll have the time to do so. Maybe maybe you can just fwport this to v18 directly?

victoralmau:
> I'd love to, but right now I don't think I'll have the time to do so. Maybe maybe you can just fwport this to v18 directly?

Ok, thanks for clarifying, I don't have time either. 
Whenever an improvement/fix is added, I think it is appropriate (should be mandatory in my opinion) to add those changes to higher versions, or else, we will find ourselves with bugs already resolved simply because no time has been spent to do so. I always do it, and personally I find it somewhat “_selfish_” to say that one does not have time, maybe not having time now ok, but doing it soon ok, not avoiding that someone will end up doing it (because at some point someone will stop doing it).

PR #761, opened 2025-01-29
[16.0][FIX] partner_multi_company: remove unnecessary rule rewrite hooks

yajo:
Now that `base_multi_company` supports `('company_id', 'in', some_ids + [False])` domains since https://github.com/OCA/multi-company/pull/760, it's time to remove the (un)install hooks that changed those rules.

The fix is important when migrating a database from previous versions where these hooks didn't exist. In those cases, users wouldn't be able to browse some partners, getting a weird `AccessError`.

FWIW, [upstream rules changed for performance][2], so the uninstall hook had to be changed nevertheless.

[2]: https://github.com/odoo/odoo/commit/a04df5cb1411a14752ba7a4d6b3c4b7b4f1d6d44#diff-1f2cafa0e26df218210181c72e599c279fd093c4215d1dc1da5ef8875195e6c2L23-R23

Draft until:
- [x] https://github.com/OCA/multi-company/pull/760

**Instructions for functional tests:** This refactor is internal and should not be noticed functionally. The module `partner_multi_company` should work just as always. If so, then this is good.

@moduon MT-8863

### DIFF
--- a/partner_multi_company/__init__.py
+++ b/partner_multi_company/__init__.py
@@ -1,3 +1,2 @@
 from . import models
 from .hooks import post_init_hook
-from .hooks import uninstall_hook
--- a/partner_multi_company/__manifest__.py
+++ b/partner_multi_company/__manifest__.py
@@ -14,5 +14,4 @@
     "data": ["views/res_partner_view.xml"],
     "installable": True,
     "post_init_hook": "post_init_hook",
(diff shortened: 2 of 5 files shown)


pedrobaeza (approved):

EmilioPascual (approved):

yajo:
This seems ready to merge @pedrobaeza

PR #755, opened 2025-01-15
[FIX]ir_config_parameter_multi_company: defensive for installation

flotho:
during the installation process, the table parameter could be queried by other modules, for exemple CRM. at this moment the filed is not available in the table causing https://github.com/OCA/multi-company/issues/754 This patch check the presence of the field as a defensive patch.

PR #763, opened 2025-01-30
[FIX] ir_config_parameter_multi_company: hook to avoid error with CRM

SaraDForgeFlow:
When installing the ir_config_parameter_multi_company module, an error occurs if the crm module is already installed. The issue arises because the module tries to access the company_id column in the ir_config_parameter table, but this column does not exist. Error observed: 

Captura desde 2025-01-30 10-14-26

To solve it, a hook is added to the ir_config_parameter_multi_company module to ensure that the company_id column exists before any code attempts to use it.

### DIFF
--- a/ir_config_parameter_multi_company/__init__.py
+++ b/ir_config_parameter_multi_company/__init__.py
@@ -1,1 +1,2 @@
 from . import models
+from .hooks import pre_init_hook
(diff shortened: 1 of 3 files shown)


eLBati:
@SaraDForgeFlow please check https://github.com/OCA/multi-company/pull/755

flotho:
Hi @eLBati @legalsylvain , isn't this PR better structured than mine for solving this issue ?

legalsylvain:
/ocabot merge patch

thanks !

PR #766, opened 2025-02-03
[16.0][port][account_invoice_inter_company] attach pdf

bguillot:
Hello, 

This is a port of https://github.com/OCA/multi-company/pull/283 to add the invoice pdf in both  companies

### DIFF
--- a/account_invoice_inter_company/models/account_move.py
+++ b/account_invoice_inter_company/models/account_move.py
@@ -1,7 +1,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-import logging
 import base64
+import logging
 
 from odoo import _, api, fields, models
 from odoo.exceptions import AccessError, UserError
@@ -56,7 +56,7 @@ class AccountMove(models.Model):
             # do not consider invoices that have already been auto-generated,
             # nor the invoices that were already validated in the past
             dest_company = src_invoice._find_company_from_invoice_partner()
-            if not dest_company or src_invoice.auto_generated:
+            if not dest_company:
                 continue
             # If one of the involved companies have the intercompany setting disabled, skip
             if (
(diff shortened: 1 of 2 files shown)


victoralmau (changes requested):
I think it would be more appropriate to make the `copy()` of the `ir.attachment` of the source invoice indicating the `res_id` of the destination invoice reducing and simplifying the code. What do you think?

bguillot:
Hello, I think the pdf is not always generated in the other company, and in case of several attachments, how to know which one to take ? 
It's safer to generate  the right attachment.

florian-dacosta (approved):
LGTM
I also think it is safer to generate the report.
AFAIK, at this stage, it is probable that the pdf does not exist yet on the customer invoice.

victoralmau (approved):

victoralmau:
Ping @pedrobaeza

account_invoice_inter_company/models/account_move.py
```diff
@@ -68,9 +69,12 @@ def action_post(self):
...
                 src_invoice = src_invoice.sudo()
-            src_invoice.with_company(dest_company.id).with_context(
-                skip_check_amount_difference=True
-            )._inter_company_create_invoice(dest_company)
+            if not src_invoice.auto_generated:
+                src_invoice.with_company(dest_company.id).with_context(
+                    skip_check_amount_difference=True
+                )._inter_company_create_invoice(dest_company)
+            if src_invoice.filtered(lambda i: i.is_sale_document()):
+                src_invoice._attach_original_pdf_report()
```
pedrobaeza: `src_invoice` is already unique.

```suggestion
            if src_invoice.is_sale_document():
                src_invoice._attach_original_pdf_report()
```

account_invoice_inter_company/models/account_move.py
```diff
@@ -79,6 +83,23 @@ def action_post(self):
             invoice.sudo()._set_intercompany_supplier_invoice_ref()
         return res
 
+    def _attach_original_pdf_report(self):
+        supplier_invoice = self.auto_invoice_id
+        if not supplier_invoice:
+            supplier_invoice = self.search([("auto_invoice_id", "=", self.id)], limit=1)
+        report = self.env.ref("account.account_invoices").with_company(self.company_id)
```
pedrobaeza, line 90:
```suggestion
        report = self.env.ref("account.account_invoices").with_company(supplier_invoice.company_id)
```

florian-dacosta: I think we really want the customer invoice's pdf here (to attach it on the supplier invoice).
Why would we generate the pdf of customer invoice (which is `self`) using the supplier invoice's company ?
Maybe the `with_company` is kind of useless since the current company should already be `self.company_id`, but I don't understand why we would want to use the supplier one.

pedrobaeza: OK, right. I have mixed things, but I think this would help if `self` is the supplier invoice, as it's the target of the action to do (attach there the customer invoice), and it's also advised to add docstrings. Let's not block this anyway.

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
And the commit messages are horrible too...

Please next time stick to guidelines:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

PR #767, opened 2025-02-04
[16.0][MIG] crm_tag_multi_company

MarinaAForgeFlow:
Supersedes PR #659 

[ADD] crm_tag_multi_company_sale: glue module
[ADD] crm_tag_multi_company_event_crm: glue module


JordiMForgeFlow:
@MarinaAForgeFlow could you squash Migration 16.0 commits? also it would be nice to put in the description of the PR that you are also adding a couple of new modules :)

AlexPForgeFlow (approved):
code review!

legalsylvain (approved):
code review / no test.

thanks for porting this module !

crm_tag_multi_company/views/crm_tag.xml
```diff
@@ -0,0 +1,31 @@
...
+<odoo>
+
+    <record model="ir.ui.view" id="sales_team_crm_tag_view_form">
+        <field name="name">crm.tag.form (in crm_tag_multi_company)</field>
+        <field name="model">crm.tag</field>
+        <field name="inherit_id" ref="sales_team.sales_team_crm_tag_view_form" />
+        <field name="arch" type="xml">
+            <field name="name" position="after">
+                <field name="company_id" groups="base.group_multi_company" />
+                <field name="company_id" invisible="1" />
```
legalsylvain, line 13: Is this second line mandatory ? 
it is not explained in the odoo documentation.

https://www.odoo.com/documentation/18.0/fr/developer/howtos/company.html#views

MarinaAForgeFlow: I understood it was needed after reading this point on the migration notes for v16.0 (https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0):

> On the same mood, you should check if the fields protected by a group interact in some way with business logic or UI. For example, a field being part of a domain of another field, or being used in copy/defaults. An special case is the field company_id, which is always used internally for domains on many2one fields referencing multi-company aware records. On these cases, you have to add the same field twice, one with the group, and another invisible without it.

legalsylvain: Thanks a lot !

PR #769, opened 2025-02-06
[17.0][IMP] account_invoice_inter_company: Add pdf invoices in both companies

victoralmau:
FWP from 16.0: https://github.com/OCA/multi-company/pull/766

Add pdf invoices in both companies

@Tecnativa TT54805

### DIFF
--- a/account_invoice_inter_company/models/account_move.py
+++ b/account_invoice_inter_company/models/account_move.py
@@ -1,5 +1,6 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import base64
 import logging
 
 from odoo import _, api, fields, models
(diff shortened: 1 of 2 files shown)


account_invoice_inter_company/models/account_move.py
```diff
@@ -48,18 +49,38 @@ def action_post(self):
...
 
+    def _attach_original_pdf_report(self):
+        supplier_invoice = self.auto_invoice_id
+        if not supplier_invoice:
+            supplier_invoice = self.search([("auto_invoice_id", "=", self.id)], limit=1)
+        report = self.env.ref("account.account_invoices").with_company(self.company_id)
+        pdf = report._render_qweb_pdf(report.report_name, [self.id])[0]
+        self.env["ir.attachment"].create(
+            {
+                "name": self.name + ".pdf",
```
pedrobaeza: I'm thinking that the name may contain illegal characters for a file name, like `/`, so you should mask it. It may be also advisable to switch the method for `self` being the target invoice where the PDF is attached, as the method name suggest.

pedrobaeza: and put `self.ensure_one()`.

victoralmau: > I'm thinking that the name may contain illegal characters for a file name, like `/`, so you should mask it. 

Checking the process of creating an attachment through a report or sending an email, I confirm that no characters are replaced (although normally the name defined in the `_get_report_base_filename()` method is used so it has been changed).

> It may be also advisable to switch the method for `self` being the target invoice where the PDF is attached, as the method name suggest.

I think changing the whole method to do the opposite can be even more confusing, IMO it is clear, from a sales invoice, you create the attachment and link it to the purchase invoice.

bguillot (approved):
Thanks for the port

florian-dacosta (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #771, opened 2025-02-11
[ADD] new module res_partner_category_multi_company.

legalsylvain:
Add a new field company_id on Contact tags model (res.partner.category), so it is possible to use differents tags per company, in a multi company context.

### Configure

* go to 'Contact' / 'Configuration' / 'Contact Tags'
* Set companies to tags.

res_partner_category_tree


legalsylvain:
Trivial module used in production since a year. 

/ocabot merge nobump

PR #772, opened 2025-02-12
[14.0][FIX][product_supplierinfo_intercompany] keep sellers sorted

Kev-Roche:
_prepare_sellers() method return the records in a specific order. 
We need to keep this order when returning the records.


sebastienbeau (approved):
LGTM (code review)

legalsylvain (commented):
I don't get why browse change the default order.

PR #775, opened 2025-02-12
[12.0][IMP] mail_multicompany: Improve company_id error

unaiberis:


quentinDupont:
Could you explain more precisely the aim of this PR please ? :) (and fix precommit and tests)

unaiberis:
Hi @quentinDupont. The aim of this PR is to fix the error we get when using multicompany because sometimes vals["model"] and vals["res_id"] don't exist and it gets the company id incorrectly from self.env.user.company_id.id, so it assigns the incorrect mail_server_id. With this PR we get the values vals["model"] and vals["res_id"]. From the body content

unaiberis:
Shouldn't the `create` method use `values_list` instead of `vals` in the code?  

```python
return super(MailMessage, self).create(vals)

unaiberis:
## Issue Explanation

The problem occurs when the `create` method is called from the `mail_thread` model, particularly when invoking the `message_post` function. 

image

The company it gets from `self.env.user.company_id.id` is incorrect, and that's why I added the way to get the `model` and `res_id` from the body.

PR #779, opened 2025-02-18
[12.0][IMP] mail_multicompany: take company using linked user

oihane:
When a contact is granted access to the portal, the user is created using OdooBot, so the emal takes the company from it even if the user is in other company.

With this change, this case is fixed.


quentinDupont (approved):

anajuaristi (approved):

anajuaristi:
@quentinDupont is it possible to merge this? 
thank you!!

PR #781, opened 2025-02-20
[16.0][FIX] account_invoice_inter_company: Adjust settings view to compatibilize with remove_odoo_enterprise

sergiobstoj:
Make adjustments on `res_config_settings` to make it compatible with `remove_odoo_enterprise`.

When having `account_invoice_inter_company` installed:
image

When having `account_invoice_inter_company` and `remove_odoo_enterprise`:
image

We miss **Inter Company OCA features**

### DIFF
--- a/account_invoice_inter_company/views/res_config_settings_view.xml
+++ b/account_invoice_inter_company/views/res_config_settings_view.xml
@@ -5,34 +5,37 @@
         <field name="inherit_id" ref="base_setup.res_config_settings_view_form" />
         <field name="arch" type="xml">
             <xpath expr="//div[@id='inter_company']" position="attributes">
-                <attribute name="title" />
+                <attribute name="invisible">True</attribute>
             </xpath>
-            <xpath expr="//div[@id='inter_company']/div[2]/span" position="before">
-                <div class="o_form_label">Inter Company OCA features
-                    <span
-                        class="fa fa-lg fa-building-o"
-                        title="Values set here are company-specific."
-                        aria-label="Values set here are company-specific."
-                        groups="base.group_multi_company"
-                        role="img"
-                    />
-                </div>
-            </xpath>
-            <xpath expr="//div[@id='inter_company']/div[2]" position="inside">
-                <div id="company_share_product">
-                    <field name="company_share_product" class="oe_inline" />
-                    <label
-                        string="Common Product Catalog"
-                        class="o_light_label"
-                        for="company_share_product"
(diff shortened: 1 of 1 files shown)


account_invoice_inter_company/views/res_config_settings_view.xml
```diff
@@ -5,30 +5,48 @@
...
-                <div class="o_form_label">Inter Company OCA features
-                    <span
-                        class="fa fa-lg fa-building-o"
-                        title="Values set here are company-specific."
-                        aria-label="Values set here are company-specific."
-                        groups="base.group_multi_company"
-                        role="img"
-                    />
-                </div>
+                <attribute name="attrs">{'invisible': True}</attribute>
```
pedrobaeza: Why not?

```suggestion
                <attribute name="invisible">True</attribute>
```

pedrobaeza (approved):

account_invoice_inter_company/views/res_config_settings_view.xml
sergiobstoj: Agree with you, thanks

aritzolea (approved):

pedrobaeza:
In the current runboat, the section is now shown (just the same as the second screenshot). Can you check?

sergiobstoj:
Ok, I see that the first section (Inter company OCA features) is visible to `base.group_multi_company`.

I believe we should make sure to apply same restriction to second section.

On it!

sergiobstoj:
Fixed and tested in runboat

pedrobaeza:
I'm wondering: why not having all of them under the same group?

sergiobstoj:
I tried to follow the same structure as it was before the PR, but it could be improve to make it simpler, if you're ok with it, as:

image

pedrobaeza:
Yes, I would remove the header `Invoicing` as well, and put "Common Product Catalog" after the main check (even hiding the rest if not marked).

pedrobaeza:
@sergiobstoj  `Common Product Catalog` option is still shown although "Generate Intercompany Invoices" is disabled. Please change that, and remember to ping involved people when the job is done, as if not, the may not be aware.

sergiobstoj:
@pedrobaeza sorry for the delay... I have applied all of your suggestions now. Can you check?

Thank you!

pedrobaeza (approved):
Thanks. Please fw-port it.

/ocabot merge minor

PR #710, opened 2024-10-23
[18.0][MIG] partner_multi_company: Migration to 18.0

BertVGroenendael:

PR #785, opened 2025-02-26
[17.0][IMP-FIX] partner_multi_company: no longer define display_name as store

### LINKED ISSUE #784: partner_multi_company v17 Missing contact detail with address on reports
THIS ISSUE IS IN V17 MODULE
## Module
The name of the module that has a bug.
Module Name : partner_multi_company

## Describe the bug
- After installing the module 'partner_multi_company', when we go to any purchase order and try to print purchase order report the vendor detail with address is missing.

- It is same on sale order and other reports.

## To Reproduce
**Affected versions**:

1. Install 'partner_multi_company', 'purchase' with just demodata
2. Check contact that has complete address details set 
3. Create purchase order or open existing with the same vendor
4. Print purchase order report
5. Check vendor detail on the report. There will be just name and phone number of vendor. It is missing the address lines, city, state and other address details.

**Expected behavior**
It should display the same address on the reports that we have in odoo standard(before installing the module)

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

pilarvargas-tecnativa:
In the partner_multi_company module, the display_name field had to be redefined as store and index because the module installation removed those field properties. As of v17 the field is no longer store so it is not necessary to redefine it as storable. In addition, this redefinition causes errors when printing reports because the field is not calculated when accessing and therefore only shows the data stored in the database, which is not updated until one of the fields of the dependency changes.

The issue in v10 why this redefinition was necessary: https://github.com/OCA/multi-company/issues/84

This PR resolves this issue: https://github.com/OCA/multi-company/issues/784

cc @tecnativa

@pedrobaeza @carlos-lopez-tecnativa please review

### DIFF
--- a/partner_multi_company/README.rst
+++ b/partner_multi_company/README.rst
@@ -84,6 +84,7 @@ Contributors
 
    -  Pedro M. Baeza <pedro.baeza@tecnativa.com>
    -  Vicent Cubells <vicent.cubells@tecnativa.com>
+   -  Pilar Vargas
 
 -  Kitti Upariphutthiphong <kittiu@ecosoft.co.th>
 
--- a/partner_multi_company/models/res_partner.py
+++ b/partner_multi_company/models/res_partner.py
@@ -2,22 +2,13 @@
 # Copyright 2015-2019 Pedro M. Baeza <pedro.baeza@tecnativa.com>
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html.html
 
-from odoo import Command, _, api, fields, models
+from odoo import Command, _, api, models
 from odoo.exceptions import ValidationError
 
 
 class ResPartner(models.Model):
     _inherit = ["multi.company.abstract", "res.partner"]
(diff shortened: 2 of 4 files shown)


captivea-darshan:
@pilarvargas-tecnativa , Thanks the code you have changed is fixing the issue. Can you please merge the code?

carlos-lopez-tecnativa (approved):
It makes sense because, in V17, the field used for searching is [complete_name](https://github.com/odoo/odoo/blob/c1bc0f8e0340a04bc921c436411c6ccd94c449f7/odoo/addons/base/models/res_partner.py#L179) instead of `display_name`, so there is no regression related to the issue reported in V10. Thanks.

After merging, please request to include it in the migration to 18: https://github.com/OCA/multi-company/pull/710

pedrobaeza (approved):
In fact, that code seems a bad patch for bugfixing the existing problem that has been too much time there.

/ocabot merge patch

PR #787, opened 2025-02-26
[16.0][IMP] product_tax_multicompany_default: Extend ignored_company_ids context functionality

sergiobstoj:
Give more functionality to `ignored_company_ids` context.

There could be cases where some companies could not apply to propagation, for example, if differs on country.

So here we add `ignored_company_ids` context into `set_multicompany_taxes()` in order to make the module inheritable to be able to exclude companies from tax propagation.

Taxes are already deleted before `self.write`, so replace `[6, 0, [ids]]` for `[[4, id] for id in ids]`


aritzolea:
Looks good but I miss a test with `ignored_company_ids` context

aritzolea:
> Looks good but I miss a test with `ignored_company_ids` context

Thanks

aritzolea (approved):

PR #564, opened 2024-01-03
[16.0][IMP] account_invoice_inter_company: move method _check_intercompany_product to account.move.line

JordiBForgeFlow:
This helps to improve inheritance.

PR #788, opened 2025-02-26
[16.0][IMP] account_invoice_inter_company: move method _check_intercompany_…

MarinaAForgeFlow:
…product to account.move.line

This change was already suggested in: https://github.com/OCA/multi-company/pull/564 but it got closed. Reopening it.

### DIFF
--- a/account_invoice_inter_company/__init__.py
+++ b/account_invoice_inter_company/__init__.py
@@ -1,5 +1,1 @@
-# Copyright 2013-2014 Odoo SA
-# Copyright 2015-2017 Chafique Delli <chafique.delli@akretion.com>
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-
 from . import models
--- a/account_invoice_inter_company/models/__init__.py
+++ b/account_invoice_inter_company/models/__init__.py
@@ -1,7 +1,3 @@
-# Copyright 2013-2014 Odoo SA
-# Copyright 2015-2017 Chafique Delli <chafique.delli@akretion.com>
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-
 from . import account_move
 from . import res_config_settings
 from . import res_company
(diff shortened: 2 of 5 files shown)


JordiMForgeFlow (approved):
code review 👍🏼

LoisRForgeFlow (approved):
Let's see if this time you are more lucky :crossed_fingers:

pedrobaeza (changes requested):
Thanks for doing this. The only arguable thing is to move out the `account.move.line` class into a separate file. It's correct and even desirable that a "one2many" model is inside its parent model file, but moreover, this not being a new module, it's convenient to preserve the previous developer decision to join/split them for not causing too much diff in the commit that makes difficult to recognize the purpose of it. If you want to move it out though due to several reasons (I don't recommend it), do it in a separate commit with tag `[REF]`.

MarinaAForgeFlow:
> Thanks for doing this. The only arguable thing is to move out the `account.move.line` class into a separate file. It's correct and even desirable that a "one2many" model is inside its parent model file, but moreover, this not being a new module, it's convenient to preserve the previous developer decision to join/split them for not causing too much diff in the commit that makes difficult to recognize the purpose of it. If you want to move it out though due to several reasons (I don't recommend it), do it in a separate commit with tag `[REF]`.

Thanks  @pedrobaeza  for the feedback. I see your point about keeping the diff minimal. However, I have decided to keep the change, but I have moved it to a separate commit with the [REF] tag, as suggested.

account_invoice_inter_company/models/account_move_line.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2013-2014 Odoo SA
+# Copyright 2015-2017 Chafique Delli <chafique.delli@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import logging
+
+from odoo import _, api, fields, models
+from odoo.exceptions import AccessError, UserError
+
+_logger = logging.getLogger(__name__)
```
pedrobaeza: This is not used AFAIK

pedrobaeza (commented):
Doing the move, it implies also that all the code translations (like `UserError` messages) are rewritten with the new code path (and this thanks to the bots we have in OCA doing extra commits. If not, the translations won't act), so again I beg you to not do such refactoring in this module to not increase the boilerplate and noise about it.

MarinaAForgeFlow:
Thanks @pedrobaeza , I was not aware of everything that implied moving the model to a new file. We have finally decided to undo the change.

pedrobaeza (approved):
Thank you very much for hearing me.

/ocabot merge patch

Issue #741, opened 2024-12-13
[16.0] purchase_sale_stock_inter_company : partial dropshipping deliveries is wrong


metaminux:
## Module
purchase_sale_stock_inter_company

## Describe the bug
When using inter company and dropshipping, if delivery is not complete (but with partial deliveries and backorders), quantities computed as delivered in the "customer" company are wrong.

## To Reproduce
**16.0**:

1. In a runboat, install **purchase_sale_stock_inter_company**
2. Configure My Company (Chicago) :
  1. Sale from purchase = True
  2. Warehouse = Chicago 1
  3. Sale user = Marc Demo
  4. Sale Orders Auto Validation = True
3. Configure My Company (San Francisco) :
  1. Sale from purchase = True
  2. Warehouse = Your Company
  3. Sale user = Marc Demo
  4. Sale Orders Auto Validation = True
4. Enable dropshipping
5. Configure product "FURN_6666" :
  1. Dropship route = True
  2. Add a first supplier price with supplier = "My Company (San Francisco)"
6. Switch to My company (Chicago)
7. Open customer "Deco Addict" and click on Sales/Quotations
8. Click New button to create a new quotation
9. Add a line for product "FURN_6666" with qty = 5
10. Confirm quotation.
11. Go to generated PO with My Company (San Francisco) as supplier and confirm it
12. Switch to My company (San Francisco)
13. Go to the new created inter company SO for customer My company (Chicago)
14. Go to the generated PO for the external suppllier (by default "Wood corner") and confirm it
15. Go to the dropship and set done quantity = 3, validate and create a backorder (everything is fine in both companies)
16. Finish the delivery (qty = 2)

**Expected behavior**
In both companies, we should have 2 dropship pickings for each sales orders in "done" state and quantities should be 3 for the first one and 2 for the second (the backorder).
Each sale order should have a computed delivered quanity of 5.

Here, quantities are OK in My Company (San Francisco) (the "supplier" company)
But quantities are not OK in My Company (Chicago) (the "customer" company) because quantities are reset to 0 on the first dropship picking when validating the backorder. So final computed delivered quantity is only 2 !

**Additional context**
The steps to reproduce are quite long, but I think the bug is related to the _action_done method of stock.picking

https://github.com/OCA/multi-company/blob/c5bb14cf526688a44f83f8f2805ad522ef569044/purchase_sale_stock_inter_company/models/stock_picking.py#L14-L57

I'll propose a PR with a test confirming this issue and a fix.

[...]

PR #789, opened 2025-03-02
[16.0][FIX][purchase_sale_stock_inter_company] synchronize inter-company picking qty

Kev-Roche:
In an intercompany purchase => sale context, if the sale picking have a backorder and both initial picking and backorder are validated, on the purchase part, the initial picking is reset to 0 quantity received. 
We need to avoid changes of the received quantities if the picking is done. 
First commit add a failing test, second commit fix it.


emiliesoutiras (approved):

carlos-lopez-tecnativa (approved):
Tested before and after this PR, and I confirm that it solves the problem. Thanks!

metaminux:
fixes #741

PR #801, opened 2025-03-13
[15.0][ADD] product_multi_company_stock: do not allow to remove a company from company_ids if there is stock or moves in that company

JoanSForgeFlow:
This module ensures that a company cannot be removed from company_ids if there are quants or stock moves associated with it.

Without this restriction, access errors may occur when checking Stock Valuation and other reports

Odoo already prevents changes to company_id when there are existing quants or moves. This PR extends the same logic to company_ids.


product_multi_company_stock/__manifest__.py
```diff
@@ -0,0 +1,13 @@
...
+{
+    "name": "Product multi-company Stock",
+    "summary": "Does not allow to remove company if there is stock or moves in that company",
+    "author": "ForgeFLow," "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/multi-company",
+    "category": "Product Management",
+    "version": "15.0.1.0.0",
+    "license": "AGPL-3",
+    "depends": ["product_multi_company", "stock"],
+}
```
AaronHForgeFlow, line 14: Maybe auto_install = True ?

AaronHForgeFlow (approved):
Code LGTM!

JordiMForgeFlow (approved):
code review 👍🏼

product_multi_company_stock/__manifest__.py
LoisRForgeFlow: agree!

JoanSForgeFlow: Comments attended. Thank you!

PR #804, opened 2025-03-15
[16.0][FIX][purchase_sale_stock_inter_company] multiple products leads to a backorder

Kev-Roche:
In case of an intercompany purchase > sale with multiple lines / products, the validation of the sale's picking leads to : 
- first line / product wil be fill on the purchase picking
- other lines / products will be in a new picking on the purchase side. 

This PR add a test and the fix.


emiliesoutiras (approved):

giarve (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #812, opened 2025-04-08
[14.0][FIX] purchase_sale_inter_company: support multistep deliveries

HekkiMelody:
Before this commit, if the company selling had setup 2- or 3-step deliveries on its warehouse, 2 or 3 receipts would be created on the PO side.

When synchronizing, we only need to consider the final outgoing picking, and not the intermediate pickings.


francesco-ooops (approved):
Functional ok!

GSLabIt (approved):
LGTM

PR #821, opened 2025-05-13
[16.0][MIG] product_supplierinfo_intercompany

giarve:
https://github.com/OCA/multi-company/pull/608

Having 2 important fixes as additional commits such that they can be ported to other versions

1) Creating a product and generating its supplierinfo is working now (def create)
2) the pricelist from a company uses the prices from its own company (no data leak for another company) and dynamic costs and list price update on the fly is now supported (no need to re-sync the list with scheduled action)


product_supplierinfo_intercompany/models/product_template.py
```diff
@@ -0,0 +1,78 @@
...
+            if rec.sale_ok and rec.purchase_ok:
+                rec._synchronise_supplier_info()
+        return res
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "sale_ok" in vals or "purchase_ok" in vals:
+            for rec in self:
+                rec._synchronise_supplier_info()
+        return res
```
giarve, line 85: For some reason when I create a product on runboat this is not working and the pricelist is not created

both sale_ok and purchase_ok are set by default

if I save then change the fields, it works.

probably when they are modified (write) is when they are set as vals but otherwise (on creation) not

giarve:
Ideally in the future we could have something like this to ensure we are not missing any changes in the pricelist

https://github.com/OCA/product-attribute/blob/0220b57a2393214b8f0f579931ade2ad5c24646c/product_list_price_from_pricelist/models/product_pricelist.py#L77

giarve:
https://github.com/OCA/multi-company/pull/714

this was also fixed here

PR #818, opened 2025-04-22
[IMP] mail_template_multi_company

SirPyTech:
This improvment fixes the following use case:
1. Install `mail_template_multi_company` and `account`
2. Create another company **company_2**
3. Edit the email template `account.email_template_edi_invoice` and set its company to the new company **company_2**
4. Go back to the original company
5. In an invoice, click on the "Send invoice"

Before this PR:
An `AccessError` is raised because the template `account.email_template_edi_invoice` is not accessible, the relevant part of the stack is:
<details><summary>Stack</summary>
<pre>
[...]
  File "/opt/odoo/auto/addons/account/models/account_move.py", line 2895, in action_invoice_sent
    lang = template._render_lang(self.ids)[self.id]
  File "/opt/odoo/auto/addons/mail/models/mail_render_mixin.py", line 420, in _render_lang
    rendered_langs = self._render_template(self.lang, self.model, res_ids)
  File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 1022, in __get__
    record._fetch_field(self)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3085, in _fetch_field
    self._read(fnames)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3205, in _read
    raise self.env['ir.rule']._make_access_error('read', forbidden)
odoo.exceptions.AccessError: Due to security restrictions, you are not allowed to access 'Email Templates' (mail.template) records.
</pre>
</details> 

After this PR:
The email composer is opened, empty.

Additional notes:
A similar error is raised also for `sale` module (and maybe others), and its templates used to send a sale order.

PR #823, opened 2025-05-20
[IMP] mail_template_multicompany: Fall back on substitute

SirPyTech:
When a template corresponding to a XMLID is not accessible, find one of its configured substitutes.

This is useful when other modules find a mail template by XMLID but that cannot be accessed: currently (thanks to https://github.com/OCA/multi-company/pull/818) no template is found.
This happens in https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/sale/models/sale.py#L840-L852 for instance.

With this change, another template can be linked to the template identified by the XMLID so that the new template is found instead.

### DIFF
--- a/mail_template_multi_company/README.rst
+++ b/mail_template_multi_company/README.rst
@@ -30,6 +30,13 @@ Mail Template Multi Company
 
 This module add multi-company management to mail templates.
 
+This module also adds a fallback mechanism for templates that are hardcoded using their XMLID.
+For example: the `sale` module will propose the template `sale.mail_template_sale_confirmation` when an order is confirmed.
+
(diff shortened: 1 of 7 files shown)


francesco-ooops (approved):
Functional ok! 

With this setup I'm able to create 1 invoice confirmation template per company and have it loaded in invoice mail wizard

HekkiMelody (approved):
Code review, LGTM!

francesco-ooops:
@OCA/intercompany-maintainers good for merge? :)

gurneyalex (approved):

PR #826, opened 2025-05-24
[14.0][IMP] account_invoice_inter_company: Make tests more resilients

sergio-teruel:
cc @Tecnativa 

Ensure that companies created hava EUR currency after load account chart template


rvalyi (approved):

HekkiMelody (approved):

LoisRForgeFlow (changes requested):
Tests are actually failing in CI

PR #832, opened 2025-05-27
[18.0][MIG] crm_stage_multi_company

glitchov:
Straight migration from 16.0


marielejeune (approved):
LGTM (code review)

JordiMForgeFlow:
@glitchov It does not seem to be working fine for me. Whenever I assign a company to a stage, if I try to enter the CRM app with another company it always gives an error.
image

glitchov:
Indeed. I've already noticed that and try to find an acceptable solution. The problem comes from the kanban view and read_group more precisely.

crm_stage_multi_company/models/crm_lead.py
```diff
@@ -0,0 +1,18 @@
...
+    _inherit = "crm.lead"
+
+    stage_id = fields.Many2one(check_company=True)
+
+    @api.model
+    def _read_group_stage_ids(self, stages, domain):
+        # Only return the stages the user has access to due to the new
+        # multi company rule
+        res = super()._read_group_stage_ids(stages, domain)
+        return stages.search([("id", "in", res.ids)])
```
qgroulard, line 18: Could you please elaborate on this problem and the solution you provide in the README ?

Just by reading the code this is not obvious why we need this part and why this is the best solution.

JordiMForgeFlow: seems to be working now, I guess by doing the `.search` you ensure that the stages are correctly filtered? Odoo base method uses directly `._search`

glitchov: Yes it's the goal. Odoo base uses ._search in sudo. I think .search on id or browse is the same isn't it ?

JordiMForgeFlow (approved):

JordiMForgeFlow:
@OCA/intercompany-maintainers could we get this one merged?

legalsylvain:
No review.

/ocabot merge nobump

PR #845, opened 2025-06-16
[16.0][IMP] purchase_sale_inter_company: do not require user, be able to use current user

giarve:
In most places dupe code equivalent to this is used.

`intercompany_user = (
            self.intercompany_sale_line_id.sudo().company_id.intercompany_sale_user_id
            or self.env.user
        )`

A function called _get_intercompany_user() could be implemented to pick the proper user (create_uid or current user, as it is by default when the required field is not set).


LoisRForgeFlow (commented):
Your suggestion to add a centralized method _get_intercompany_user makes sense, it might be worth doing it.

PR #605, opened 2024-03-27
[17.0][MIG] purchase_sale_stock_inter_company

cuongnmtm:
Depends on:

- [x] <https://github.com/OCA/multi-company/pull/595>
- [x] <https://github.com/OCA/multi-company/pull/604>

PR #510, opened 2023-09-19
[16.0][ADD] purchase_sale_lot_intercompany

Kev-Roche:
Migration of the PR:  https://github.com/OCA/multi-company/pull/481

PR #852, opened 2025-06-23
[17.0][MIG] purchase_sale_lot_intercompany : Migration to 17.0

metaminux:
Depends on :

- [ ] #605


metaminux:
Based on #510 which is not merged...

PR #853, opened 2025-06-26
[18.0][MIG] account_multicompany_easy_creation: Migration to 18.0

victoralmau:
Migration to 18.0

Please @pedrobaeza, @carlos-lopez-tecnativa and @carlosdauden can you review it?

@Tecnativa TT55497


pedrobaeza:
I get this error when trying to create a new company with Spanish CoA:

imagen


It can be due to the trick with load_data?

victoralmau:
> I get this error when trying to create a new company with Spanish CoA:
> 
> imagen
> 
> It can be due to the trick with load_data?

I have checked in detail the error and it is not explicitly related (although I have changed the “_trick_” so that it is only done in tests).

**What is going on?** 
In the company creation (when defining the country) or when loading the chart_template the country is set, and in the https://github.com/odoo/odoo/blob/d83801027c9a88b6e4db8166b6e2aad41ed760b9/odoo/addons/base/models/res_company.py#L227 method it tries to automatically install the `account_peppol` module (because one of its dependencies, `account_edi_ubl_cii` is auto_install), but it can't because of the `phonenumbers` dependency.

In a “_normal_” environment this would not happen, because the `account_peppol` module would not be available or because in ~OCB it is not auto-install~.

Clarified this, the “solution” (although I don't like it) is to add the `phonenumbers` dependency so that it can be installed.

victoralmau:
Finally, one more “_trick_” has been added to avoid what caused the auto-installation of the module(s).

pedrobaeza (approved):

account_multicompany_easy_creation/readme/USAGE.md
```diff
@@ -0,0 +1,4 @@
+To use this module, you need to:
+
+1.  Go to Settings \> Users and companies \> Quick Company Creation
```
carlos-lopez-tecnativa:
```suggestion
1.  Go to Settings \> Users and companies \> Company easy creation
```

account_multicompany_easy_creation/wizards/multicompany_easy_creation.xml
```diff
@@ -0,0 +1,124 @@
...
+                            </group>
+                            <group invisible="not update_default_taxes">
+                                <!-- The invisible default_sale_tax_options field should not be deleted, it is necessary for the correct functioning of the dynamic_selection widget. !-->
+                                <field name="default_sale_tax_options" invisible="1" />
+                                <field
+                                    name="default_sale_tax"
+                                    widget="dynamic_selection"
+                                    options="{'available_field': 'default_sale_tax_options'}"
+                                    invisible="not chart_template"
+                                />
```
carlos-lopez-tecnativa, line 51: I think this should filter only active taxes. By mistake, I selected an inactive tax, and the product appeared without any taxes.

image

image


image

image

account_multicompany_easy_creation/wizards/multicompany_easy_creation.xml
```diff
@@ -0,0 +1,124 @@
...
+                                <field name="force_sale_tax" />
+                            </group>
+                            <group invisible="not update_default_taxes">
+                                <!-- The invisible default_purchase_tax_options field should not be deleted, it is necessary for the correct functioning of the dynamic_selection widget. !-->
+                                <field
+                                    name="default_purchase_tax_options"
+                                    invisible="1"
+                                />
+                                <field
+                                    name="default_purchase_tax"
```
carlos-lopez-tecnativa, line 61: The same here

carlos-lopez-tecnativa (changes requested):
LGTM, just a minor change

account_multicompany_easy_creation/models/account_chart_template.py
```diff
@@ -0,0 +1,35 @@
...
+        log in https://github.com/odoo/odoo/blob/d83801027c9a88b6e4db8166b6e2aad41ed760b9/addons/account/models/chart_template.py#L147
+        ONLY in tests.
+        """
+        test_condition = config["test_enable"] and self.env.context.get(
+            "test_account_multicompany_easy_creation"
+        )
+        install_demo = False if test_condition else install_demo
+        return super()._load(
+            template_code, company, install_demo, force_create=force_create
+        )
```
carlos-lopez-tecnativa: Just a doubt: if you tag the test with `@tagged("post_install", "-at_install")`, doesn't that solve the warning? I understand that this warning is caused when the registry is not fully loaded.

carlos-lopez-tecnativa (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #844, opened 2025-06-16
[17.0][MIG] purchase_sale_stock_inter_company_mrp: Migration to 17.0

metaminux:

PR #572, opened 2024-01-22
[16.0][ADD] purchase_sale_stock_inter_company_mrp

JordiMForgeFlow:
Module to support inter-company purchase flow with kit BoMs.

It considers the following cases:

Product is Kit in Company A but not in Company B
Product is Kit in Company B but not in Company A
Product is Kit in Company A and in Company B (having the same or different BoMs)

PR #856, opened 2025-07-09
[17.0][MIG] purchase_sale_stock_inter_company: Migration to version 17.0

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/multi-company/pull/605
TT54931
@Tecnativa @pedrobaeza @cuongnmtm @pomazanbohdan @metaminux @JordiMForgeFlow could you please review this?


purchase_sale_stock_inter_company/models/stock_move_line.py
```diff
@@ -0,0 +1,99 @@
...
+            po_moves = purchase_line_origin.move_ids.filtered(
+                lambda m: m.picking_id
+                == self.move_id.picking_id.intercompany_picking_id
+                and m.product_id == self.product_id
+            )
+        return po_moves
+
+    @api.model
+    def _get_fields_to_sync_intercompany(self):
+        return {"qty_done", "lot_id"}
```
pedrobaeza: What about `package_id` and `dest_package_id`?

purchase_sale_stock_inter_company/models/stock_picking.py
```diff
@@ -2,160 +2,200 @@
...
-from odoo import api, fields, models
+from odoo import SUPERUSER_ID, _, api, fields, models
+from odoo.exceptions import UserError
 
 
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
     intercompany_picking_id = fields.Many2one(comodel_name="stock.picking", copy=False)
+    # to silent the warning
```
pedrobaeza: `to silence the warning`

pedrobaeza (approved):

purchase_sale_stock_inter_company/models/stock_move_line.py
carlos-lopez-tecnativa: I think this is out of scope for this module. These fields are `many2one`, so it's necessary to implement the logic to search or create the corresponding records in the other company. This is not supported in any version. Please confirm if I should add this support here or after the migration process.

purchase_sale_stock_inter_company/models/stock_picking.py
carlos-lopez-tecnativa: Done

purchase_sale_stock_inter_company/models/stock_move_line.py
pedrobaeza: I think the same happens with `lot_id`. Not sure if `stock.quant.package` can have `company_id` empty, so you can assign it directly. If not, just put the comment in the roadmap.

carlos-lopez-tecnativa: According to this code, the `company_id` is computed based on the `stock.quant`:
https://github.com/odoo/odoo/blob/e07ff7fab04a3536187b8ab0facbd17236754fae/addons/stock/models/stock_quant.py#L1489-L1494
and in turn, it is related to the `location`:
https://github.com/odoo/odoo/blob/e07ff7fab04a3536187b8ab0facbd17236754fae/addons/stock/models/stock_quant.py#L58
So, internal locations always have a company set (it can be empty, but normally it has a company).
I added this to the roadmap.

metaminux: @carlos-lopez-tecnativa FYI, It seems stock_intercompany_bidirectional in 16.0 is managing packages...

carlos-lopez-tecnativa: > @carlos-lopez-tecnativa FYI, It seems stock_intercompany_bidirectional in 16.0 is managing packages...

I tested this module, but I think it’s not working as expected.
I created a new picking and set the package:
image
When I validate the picking in C1, the package is lost (more specifically, it is transferred to the picking in C2), but in C1 I don't see the package anymore. I'm not sure if it's completely correct.

Have you tested this module?

metaminux: I must admit I didn't test it and didn't look at the code either... Just read the description while looking for a solution to manage lots in inter-company pickings.

purchase_sale_stock_inter_company/tests/test_inter_company_purchase_sale_stock.py
```diff
@@ -0,0 +1,639 @@
...
+            cls.quant_obj.create(
+                {
+                    "product_id": product.id,
+                    "location_id": cls.warehouse_a.lot_stock_id.id,
+                    "quantity": 1,
+                    "lot_id": lot.id,
+                }
+            )
+        return lot
+
```
metaminux:
```suggestion
```

This code is duplicated, see lines 30-45

metaminux (changes requested):

metaminux:
This is amazing work, thanks @carlos-lopez-tecnativa

I gathered every PR for versions 14.0, 15.0 and 16.0 and a lot are already there.

Only some questions and remarks :

- Did you test if bug related to stock module is not there anymore (see #555)
- Should we add test_sync_picking_different_product_multiple_lines from #583 (I must admit I don't understand precisely the use case, but it's related to stock so should probably be there)
- Fix from #637 seems to be missing (no test though, probably because it would add a dependency on stock_dropshipping...)
- Also changes from #572 should be included to stay compatible with purchase_sale_stock_inter_company_mrp

That's all !

carlos-lopez-tecnativa:
> Only some questions and remarks :
> 
> * Did you test if bug related to stock module is not there anymore (see [[14.0][FIX] purchase_sale_inter_company: fix demo data #555](https://github.com/OCA/multi-company/pull/555))

According to [this comment](https://github.com/OCA/multi-company/pull/555#issuecomment-1864720091), the PR does not affect the tests, so I cherry-picked the commit. Thanks!


> * Should we add test_sync_picking_different_product_multiple_lines from [[14.0] [FIX] purchase_sale_inter_company: wrong field #583](https://github.com/OCA/multi-company/pull/583) (I must admit I don't understand precisely the use case, but it's related to stock so should probably be there)

As for whether we should include `test_sync_picking_different_product_multiple_lines`. I don’t think this commit should be included here.

In version 14.0, `sale.order` synchronizes with pickings in `action_confirm`. Honestly, I don’t fully understand that logic, synchronization should be done when the picking is validated, not before.

So this feature (and related commits like [this one](https://github.com/OCA/multi-company/commit/9f6784946368e7937c4d43b2bd1b5dc0330b3d32)) was not cherry-picked.

If you understand this change and believe it must be included, please provide more context.

Additionally, the test [test_sync_picking_different_product_multiple_lines](https://github.com/OCA/multi-company/blob/1996cfff41bb16d7a5603e406afd10d4c9fb89ee/purchase_sale_inter_company/tests/test_inter_company_purchase_sale.py#L652-L673) in the referenced PR does not include any assertion. It only creates data, so from my point of view, this test is incomplete.


> * Fix from [[14.0][FIX] Picking sync should also work for dropshipping #637](https://github.com/OCA/multi-company/pull/637) seems to be missing (no test though, probably because it would add a dependency on stock_dropshipping...)

This was originally included, but I later adapted it during the “Finish migration to version 17.0” commit, since it’s no longer applicable.

In V14, the logic was handled in button_validate, but in V17, it now resides in [action_done](https://github.com/OCA/multi-company/pull/856/files#diff-b19615b2f288e2552837438a2153b05fad17576e1cf7bc9c0954746f062e3ba8R65), and in this case, it's handled via `location_dest_id.usage == "customer"`.

image


> * Also changes from [[16.0][ADD] purchase_sale_stock_inter_company_mrp #572](https://github.com/OCA/multi-company/pull/572) should be included to stay compatible with purchase_sale_stock_inter_company_mrp
> 

The function `_get_product_intercompany_qty_done_dict` is already included, as it is inherited in the `purchase_sale_stock_inter_company_mrp` module.
However, the `_set_intercompany_picking_qty` function was merged and its logic adapted into the `_action_done_intercompany_actions` method.

Could you please update PR https://github.com/OCA/multi-company/pull/844 to reference this PR  and verify if it's OK?
If you find any inconsistencies, please let me know.

purchase_sale_stock_inter_company/tests/test_inter_company_purchase_sale_stock.py
carlos-lopez-tecnativa: Done, thanks

metaminux:
> In V14, the logic was handled in button_validate, but in V17, it now resides in [action_done](https://github.com/OCA/multi-company/pull/856/files#diff-b19615b2f288e2552837438a2153b05fad17576e1cf7bc9c0954746f062e3ba8R65), and in this case, it's handled via `location_dest_id.usage == "customer"`.

OK, sorry, I searched for `record.partner_id.commercial_partner_id.ref_company_ids` and found it in `button_validate` but it's used to block the picking related to the purchase, so dropshipping shouldn't have any effect here.

However, I think that we have an issue here to detect pickings that should trigger synchronization.
For instance, I think returns from an intercompany PO should  be sync on the SO side... but this could be treated in a new issue after merging.

> > * Also changes from [[16.0][ADD] purchase_sale_stock_inter_company_mrp #572](https://github.com/OCA/multi-company/pull/572) should be included to stay compatible with purchase_sale_stock_inter_company_mrp
> 
> The function `_get_product_intercompany_qty_done_dict` is already included, as it is inherited in the `purchase_sale_stock_inter_company_mrp` module. However, the `_set_intercompany_picking_qty` function was merged and its logic adapted into the `_action_done_intercompany_actions` method.

You're right, I searched for `_set_intercompany_picking_qty` and missed the important part. Sorry !

> Could you please update PR #844 to reference this PR and verify if it's OK? If you find any inconsistencies, please let me know.

I'll do that and let you know, but I can already approve this PR.

Thanks again !

metaminux (approved):

PR #859, opened 2025-07-15
[14.0][IMP] product supplierinfo intercompany[_multi_company]: refactor, archiving

HekkiMelody:
These commits

- slightly refactor these two modules so that the method doesn't need to be overwritten
- implement deletion and recreation of supplierinfo when archiving and unarchiving product templates and product pricelists

More details in each commit

### DIFF
--- a/product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
+++ b/product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
@@ -23,6 +23,7 @@ class ProductIntercompanySupplierMixin(models.AbstractModel):
             "name": pricelist.company_id.partner_id.id,
             "company_id": False,
             "price": price,
+            "currency_id": pricelist.currency_id.id,
         }
         return res
 
--- a/product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
+++ b/product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
@@ -24,6 +24,7 @@ class ProductIntercompanySupplierMixin(models.AbstractModel):
             "company_id": False,
             "price": price,
             "currency_id": pricelist.currency_id.id,
+            "delay": pricelist.intercompany_supplier_lead_time,
         }
         return res
 
--- a/product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
+++ b/product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
@@ -43,15 +43,28 @@ class ProductIntercompanySupplierMixin(models.AbstractModel):
             ):
                 domain = record._get_intercompany_supplier_info_domain(pricelist)
                 supplierinfo = record.env["product.supplierinfo"].search(domain)
-                if (
-                    record._has_intercompany_price(pricelist)
-                    and record.sale_ok
-                    and record.purchase_ok
-                ):
-                    vals = record._prepare_intercompany_supplier_info(pricelist)
-                    if supplierinfo:
-                        supplierinfo.write(vals)
-                    else:
-                        supplierinfo.create(vals)
-                elif supplierinfo:
-                    supplierinfo.sudo().unlink()
+                record._synchronise_supplier_info_for_record(pricelist, supplierinfo)
+
+    def _synchronise_supplier_info_for_record(self, pricelist, supplierinfo):
+        self.ensure_one()
+        if self._condition_supplierinfo_create_or_update(pricelist, supplierinfo):
+            vals = self._prepare_intercompany_supplier_info(pricelist)
+            if supplierinfo:
+                supplierinfo.write(vals)
+            else:
+                supplierinfo.create(vals)
+        elif self._condition_supplierinfo_unlink(pricelist, supplierinfo):
+            supplierinfo.sudo().unlink()
+
+    def _condition_supplierinfo_create_or_update(self, pricelist, supplierinfo):
+        self.ensure_one()
+        return (
+            self._has_intercompany_price(pricelist)
+            and self.sale_ok
+            and self.purchase_ok
+            and self.active
+        )
+
+    def _condition_supplierinfo_unlink(self, pricelist, supplierinfo):
+        self.ensure_one()
+        return bool(supplierinfo) or (not self.active)
--- a/product_supplierinfo_intercompany/models/product_pricelist.py
+++ b/product_supplierinfo_intercompany/models/product_pricelist.py
@@ -43,3 +43,13 @@ class ProductPricelist(models.Model):
         self.sudo().with_context(automatic_intercompany_sync=True).mapped(
             "generated_supplierinfo_ids"
         ).unlink()
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "active" in vals:
+            to_sync = self.filtered("is_intercompany_supplier")
+            if vals["active"]:
+                to_sync._active_intercompany()
+            else:
+                to_sync._unactive_intercompany()
+        return res
--- a/product_supplierinfo_intercompany/models/product_pricelist.py
+++ b/product_supplierinfo_intercompany/models/product_pricelist.py
@@ -12,6 +12,10 @@ class ProductPricelist(models.Model):
         default=False, inverse="_inverse_intercompany_supplier"
     )
 
+    intercompany_supplier_lead_time = fields.Float(
+        default=0, help="Vendor pricelist lead time, in days."
+    )
+
     generated_supplierinfo_ids = fields.One2many(
         comodel_name="product.supplierinfo",
         inverse_name="intercompany_pricelist_id",
--- a/product_supplierinfo_intercompany/models/product_template.py
+++ b/product_supplierinfo_intercompany/models/product_template.py
@@ -65,9 +65,18 @@ class ProductTemplate(models.Model):
         return res
 
     def write(self, vals):
-        res = super(ProductTemplate, self).write(vals)
-        for rec in self:
-            rec.update_intercompany_prices()
+        res = super().write(vals)
+        if (
+            "sale_ok" in vals
+            or "purchase_ok" in vals
+            or "list_price" in vals
+            or "standard_price" in vals
+            or "company_id" in vals
+            or "company_ids" in vals
+            or "active" in vals
+        ):
+            for rec in self:
+                rec.update_intercompany_prices()
         return res
 
     def update_intercompany_prices(self):
--- a/product_supplierinfo_intercompany/models/product_template.py
+++ b/product_supplierinfo_intercompany/models/product_template.py
@@ -66,18 +66,26 @@ class ProductTemplate(models.Model):
 
     def write(self, vals):
         res = super().write(vals)
-        if (
-            "sale_ok" in vals
-            or "purchase_ok" in vals
-            or "list_price" in vals
-            or "standard_price" in vals
-            or "company_id" in vals
-            or "company_ids" in vals
-            or "active" in vals
-        ):
+        if any(field in vals for field in self._fields_for_intercompany_pricelist()):
             for rec in self:
                 rec.update_intercompany_prices()
         return res
+
+    @api.model
+    def _fields_for_intercompany_pricelist(self):
+        """
+        If any of these fields change,
+        the intercompany pricelist will be recomputed
+        """
+        return [
+            "sale_ok",
+            "purchase_ok",
+            "list_price",
+            "standard_price",
+            "company_id",
+            "company_ids",
+            "active",
+        ]
 
     def update_intercompany_prices(self):
         pricelist_ids = (
--- a/product_supplierinfo_intercompany/tests/test_supplier_intercompany.py
+++ b/product_supplierinfo_intercompany/tests/test_supplier_intercompany.py
@@ -84,6 +84,9 @@ class TestIntercompanySupplierCase(SavepointCase):
         self.assertEqual(len(supplierinfo), 1)
         self.assertEqual(len(supplierinfo.intercompany_pricelist_id), 1)
         self.assertEqual(supplierinfo.price, price)
+        self.assertEqual(
+            supplierinfo.currency_id, supplierinfo.intercompany_pricelist_id.currency_id
+        )
 
 
 class TestIntercompanySupplier(TestIntercompanySupplierCase):
@@ -347,3 +350,11 @@ class TestIntercompanySupplier(TestIntercompanySupplierCase):
         # Select using sudo (as some native odoo code do it)
         supplier = self.product_product_4b.sudo()._select_seller()
         self.assertEqual(supplier.name, partner)
+
+    def test_currency_consistency(self):
+        different_currency = self.env.ref("base.CHF")
+        self.assertNotEqual(self.pricelist_intercompany.currency_id, different_currency)
+        self.pricelist_intercompany.currency_id = different_currency
+        template = self.env["product.template"].create({"name": "New One"})
+        self._add_item(template, 30)
+        self._check_supplier_info_for(template, 30)
--- a/product_supplierinfo_intercompany/tests/test_supplier_intercompany.py
+++ b/product_supplierinfo_intercompany/tests/test_supplier_intercompany.py
@@ -149,6 +149,14 @@ class TestIntercompanySupplier(TestIntercompanySupplierCase):
         supplierinfo = self._get_supplier_info()
         self.assertEqual(len(supplierinfo), 0)
 
+    def test_archive_pricelist_intercompany(self):
+        self.pricelist_intercompany.active = False
+        supplierinfo = self._get_supplier_info()
+        self.assertEqual(len(supplierinfo), 0)
+        self.pricelist_intercompany.active = True
+        supplierinfo = self._get_supplier_info()
+        self.assertEqual(len(supplierinfo), 4)
+
     def test_intercompany_access_rule(self):
         # Check that supplier this supplier info are not visible
         # with the current user "demo"
@@ -167,6 +175,10 @@ class TestIntercompanySupplier(TestIntercompanySupplierCase):
     def test_add_template_item(self):
         template = self.env.ref("product.product_product_2_product_template")
         self._add_item(template, 30)
+        self._check_supplier_info_for(template, 30)
+        template.active = False
+        self._check_no_supplier_info_for(template)
+        template.active = True
         self._check_supplier_info_for(template, 30)
 
     def test_update_product_item(self):
--- a/product_supplierinfo_intercompany_multi_company/models/product_intercompany_supplier_mixin.py
+++ b/product_supplierinfo_intercompany_multi_company/models/product_intercompany_supplier_mixin.py
@@ -1,50 +1,24 @@
-from odoo import _, models
-from odoo.exceptions import Warning as UserError
+from odoo import models
 
 
 class ProductIntercompanySupplierMixin(models.AbstractModel):
     _inherit = "product.intercompany.supplier.mixin"
 
-    def _synchronise_supplier_info(self, pricelists=None):
-        # replaced to add company_ids condition
-        if not pricelists:
-            pricelists = self.env["product.pricelist"].search(
-                [("is_intercompany_supplier", "=", True)]
+    def _synchronise_supplier_info_for_record(self, pricelist, supplierinfo):
+        res = super()._synchronise_supplier_info_for_record(pricelist, supplierinfo)
+        if (
+            self._has_intercompany_price(pricelist)
+            and self.company_id == pricelist.company_id
+        ):
+            supplierinfo.sudo().unlink()
+        return res
+
+    def _condition_supplierinfo_create_or_update(self, pricelist, supplierinfo):
+        res = super()._condition_supplierinfo_create_or_update(pricelist, supplierinfo)
+        return res and (
+            not self.company_ids
+            or (
+                pricelist.company_id in self.company_ids
+                and pricelist.company_id != self.company_ids
             )
-        for pricelist in pricelists:
-            if not pricelist.is_intercompany_supplier:
-                raise UserError(
-                    _("The pricelist %s is not intercompany") % pricelist.name
-                )
-            # We pass the pricelist in the context in order to get the right
-            # sale price on record.price (compatible v8 to v12)
-            for record in self.sudo().with_context(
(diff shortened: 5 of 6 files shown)


giarve (approved):
for _condition_supplierinfo_create_or_update

can you also add these fields such that when they are changed the pricelist recomputes?

this is when it is based on cost field and similars

https://github.com/OCA/multi-company/pull/821/files#diff-19f4129027b1cecdf793d8eb674abc4a90b136c629ee2b4cae958ff1a5f8d2f9R75

HekkiMelody:
> for _condition_supplierinfo_create_or_update
> 
> can you also add these fields such that when they are changed the pricelist recomputes?
> 
> this is when it is based on cost field and similars
> 
> https://github.com/OCA/multi-company/pull/821/files#diff-19f4129027b1cecdf793d8eb674abc4a90b136c629ee2b4cae958ff1a5f8d2f9R75

Thank you for your feedback!

I cherry-picked your commit as well as a previous one about currency. The modules are quite different, and doing more is definitely outside the scope of this PR.

I realized after I had already started that the change you requested wasn't strictly necessary because in v14 pricelists are recomputed ALWAYS, there was no list of fields to check against, and therefore there was no need to add any fields to that list. Still, while I was at it, I added it. Let me know what you think.

giarve:
> > for _condition_supplierinfo_create_or_update
> > can you also add these fields such that when they are changed the pricelist recomputes?
> > this is when it is based on cost field and similars
> > https://github.com/OCA/multi-company/pull/821/files#diff-19f4129027b1cecdf793d8eb674abc4a90b136c629ee2b4cae958ff1a5f8d2f9R75
> 
> Thank you for your feedback!
> 
> I cherry-picked your commit as well as a previous one about currency. The modules are quite different, and doing more is definitely outside the scope of this PR.
> 
> I realized after I had already started that the change you requested wasn't strictly necessary because in v14 pricelists are recomputed ALWAYS, there was no list of fields to check against, and therefore there was no need to add any fields to that list. Still, while I was at it, I added it. Let me know what you think.

great. For me it is interesting because this module is having lots of changes in between versions but nobody is set as maintainer so not guiding the boat and it was just to make sure we are following the right approach so when you migrate to v16 for instance you get the same features and non breaking changes

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
```diff
@@ -26,6 +26,7 @@ def _prepare_intercompany_supplier_info(self, pricelist):
         }
         return res
 
+    @api.model
```
SirPyTech: **todo**: This method has a `for record in self` loop so it depends on the records in `self`, therefore it should not be `api.model`.

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
```diff
@@ -43,15 +44,31 @@ def _synchronise_supplier_info(self, pricelists=None):
...
-                    supplierinfo.sudo().unlink()
+                self._synchronise_supplier_info_for_record(
+                    record, pricelist, supplierinfo
+                )
+
+    @api.model
+    def _synchronise_supplier_info_for_record(self, record, pricelist, supplierinfo):
+        if self._condition_supplierinfo_create_or_update(
+            record, pricelist, supplierinfo
+        ):
```
SirPyTech: **question**: Here and below, why are we calling on `record` methods that are also defined for `self`?
For instance, I think:
```suggestion
    def _synchronise_supplier_info_single(self, pricelist, supplierinfo):
        self.ensure_one()
        if self._condition_supplierinfo_create_or_update(
            pricelist, supplierinfo
        ):
```
would be more intuitive, but maybe it's just me, what do you think?

product_supplierinfo_intercompany_multi_company/models/product_intercompany_supplier_mixin.py
```diff
@@ -1,50 +1,26 @@
...
-                    supplierinfo.sudo().unlink()
+    @api.model
+    def _synchronise_supplier_info_for_record(self, record, pricelist, supplierinfo):
+        res = super()._synchronise_supplier_info_for_record(
+            record, pricelist, supplierinfo
+        )
+        if (
+            record._has_intercompany_price(pricelist)
+            and record.company_id == pricelist.company_id
+        ):
```
SirPyTech, line 12: **suggestion**: We could add this condition to the `_condition_..._unlink` method instead: that would use inheritance and should give the same result, what do you think?

product_supplierinfo_intercompany/models/product_pricelist.py
```diff
@@ -43,3 +43,14 @@ def _unactive_intercompany(self):
         self.sudo().with_context(automatic_intercompany_sync=True).mapped(
             "generated_supplierinfo_ids"
         ).unlink()
+
+    def write(self, vals_list):
```
SirPyTech: **typo**: Should be `vals`

product_supplierinfo_intercompany/models/product_pricelist.py
```diff
@@ -43,3 +43,14 @@ def _unactive_intercompany(self):
...
+        res = super().write(vals_list)
+        if "active" in vals_list:
+            for rec in self:
+                if rec.is_intercompany_supplier:
+                    if rec.active:
+                        rec._active_intercompany()
+                    else:
+                        rec._unactive_intercompany()
```
SirPyTech: **suggestion**: Can we batch records to activate/unactivate and then call the corresponding method on the batch? That could speed up the process

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
```diff
@@ -17,13 +17,19 @@ def _get_intercompany_supplier_info_domain(self, pricelist):
 
     def _prepare_intercompany_supplier_info(self, pricelist):
         self.ensure_one()
-        price = self.uom_id._compute_price(self.price, self.uom_po_id)
+
+        # Set context to ensure we are taking the list price / cost / etc. fields
+        # from the company set in the pricelist as it is a company-dependent field!
+        product_with_context = self.with_company(pricelist.company_id.id)
```
SirPyTech: **question**: Could you please clarify which one is the company-dependent field that needs the right company to be set?

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
```diff
@@ -17,13 +17,19 @@ def _get_intercompany_supplier_info_domain(self, pricelist):
...
+
+        price = product_with_context.uom_id._compute_price(self.price, self.uom_po_id)
```
SirPyTech: **suggestion**: `product.pricelist.get_product_price` seems to be the corresponding method to `product.pricelist._get_product_price` used in `16.0`; `uom_id`'s method simply converts from one UoM to another.

product_supplierinfo_intercompany/models/product_pricelist.py
```diff
@@ -12,6 +12,10 @@ class ProductPricelist(models.Model):
         default=False, inverse="_inverse_intercompany_supplier"
     )
 
+    intercompany_supplier_lead_time = fields.Float(
```
SirPyTech, line 15: **chore**: I don't see how adding this field helps with `[FIX] Dynamic intercompany pricelist with cost / list price as item`, can we move it to another more appropriate commit?

product_supplierinfo_intercompany/models/product_template.py
```diff
@@ -65,9 +65,18 @@ def create(self, vals):
...
-            rec.update_intercompany_prices()
+        res = super().write(vals)
+        if (
+            "sale_ok" in vals
+            or "purchase_ok" in vals
+            or "list_price" in vals
+            or "standard_price" in vals
+            or "company_id" in vals
+            or "company_ids" in vals
+            or "active" in vals
```
SirPyTech: **suggestion**: Please move this list of fields to its own method so it can be overridden if other modules add fields that need to `update_intercompany_prices`.

Also, how did we choose these fields? Can they be retrieved in some more dynamic way, if that's not overkill?

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
```diff
@@ -17,13 +17,19 @@ def _get_intercompany_supplier_info_domain(self, pricelist):
```
SirPyTech, line 17: **chore**: Please follow https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message for the commit message, right now it is missing the module name: 
image

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
```diff
@@ -23,6 +23,7 @@ def _prepare_intercompany_supplier_info(self, pricelist):
             "name": pricelist.company_id.partner_id.id,
```
SirPyTech, line 23: **chore**: Please follow https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message for the commit message, right now it is missing the module name: image

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
SirPyTech: **suggestion**: The tests are already well structured for multi-company testing, is it possible to add a test for this failing use-case?

SirPyTech (changes requested):
Thanks for the PR!
I only reviewed the code and seems mostly good to me, big thanks for keeping a clean commit history 👏.
I have some notes/suggestions, let me know what you think.

I am not a fan of cherry-picking commits from other open PRs because it doubles the developer (and reviewer's) work to sync changes in multiple PRs, but I understand that on the other hand it increases the probability to have a commit merged in _some PR_; in the end, it's a trade-off 🤷

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
HekkiMelody: I thought it would be `price` but looking at the compute method, it doesn't seem to be the case. Maybe a difference with v16.

I've deleted the comment and reverted the change

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
HekkiMelody: I kept the computation as it was previously in v14

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
HekkiMelody: I've reverted the change since the company doesn't actually seem to affect the computation of the `price` field in v14

HekkiMelody:
Thank you for the review @SirPyTech ! ~~Unless I missed something~~, I have addressed all of your comments (EDIT: I missed one comment, I'm working on it), including adding two new commits to

- add intercompany lead time in a dedicated commit
- refactor the list of fields into a method for easier overriding as you suggested

Let me know what you think!

product_supplierinfo_intercompany/models/product_template.py
HekkiMelody: I'm not entirely sure, as I've simply taken these fields from the commit by @giarve which I've cherry-picked here

Original: https://github.com/OCA/multi-company/pull/821/commits/78a47255fb145659790db01df444dcb3a7ca783e#diff-19f4129027b1cecdf793d8eb674abc4a90b136c629ee2b4cae958ff1a5f8d2f9R75-R82

product_supplierinfo_intercompany_multi_company/models/product_intercompany_supplier_mixin.py
HekkiMelody: I've checked, but it's not quite the same. If I add it to `_condition_..._unlink` but `_condition_..._create_or_update` is true, it will not be executed. This way, it's executed regardless (meaning a supplierinfo record could be created and then immediately deleted, in some cases). I've kept it as is to preserve the existing logic since it's a refactor.

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
```diff
@@ -62,6 +62,7 @@ def _condition_supplierinfo_create_or_update(self, pricelist, supplierinfo):
             self._has_intercompany_price(pricelist)
             and self.sale_ok
             and self.purchase_ok
+            and self.active
```
SirPyTech, line 67: **thought**: A condition on `self.active` could also be added in `_condition_supplierinfo_unlink` since we want to delete the supplierinfo when the product is archived, what do you think?

product_supplierinfo_intercompany/tests/test_supplier_intercompany.py
```diff
@@ -347,3 +350,9 @@ def test_getting_supplier_with_sudo(self):
         # Select using sudo (as some native odoo code do it)
         supplier = self.product_product_4b.sudo()._select_seller()
         self.assertEqual(supplier.name, partner)
+
+    def test_currency_consistency(self):
+        self.pricelist_intercompany.currency_id = self.env.ref("base.USD")
```
SirPyTech: **suggestion**: This test passes even without the related change in the module, probably because this currency is the default one (see https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/product/models/product.py#L774)

Please use another currency, and maybe also `assert` it's not the default one.

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
SirPyTech: :+1:

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
SirPyTech: 👍

product_supplierinfo_intercompany/models/product_pricelist.py
SirPyTech: 👍

product_supplierinfo_intercompany/models/product_pricelist.py
SirPyTech: 👍

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
SirPyTech: 👍

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
SirPyTech: 👍

product_supplierinfo_intercompany/models/product_pricelist.py
SirPyTech: 👍

product_supplierinfo_intercompany/models/product_template.py
SirPyTech: 👍

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
SirPyTech: 👍

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
SirPyTech: 👍

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
SirPyTech: 👍

product_supplierinfo_intercompany_multi_company/models/product_intercompany_supplier_mixin.py
SirPyTech: Right, that's because of the **el**if, thanks for the clarification 👍

SirPyTech (changes requested):
Thanks for all the changes!
Just a note to make one of the added tests really useful and some food for thought, the rest looks good

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
HekkiMelody: To be honest, as it is structured now, with the `elif`, I don't really see much use for `_condition_supplierinfo_unlink`, because effectively, if `_condition_supplierinfo_create_or_update` is True, you want to keep the supplierinfo, otherwise you want to delete it. There aren't any cases I can think of where you would want to keep around a supplierinfo that you don't want to update. That would be... obsolete?

But I also didn't want to add yet another change to the structure of this module, this PR is already too large as it is.

So maybe I could remove `_condition_supplierinfo_unlink`?

SirPyTech: I understand that you either want to keep it or trash it, but keeping it does not mean that you have to update it.

I agree that as it is now, `_condition_supplierinfo_unlink` is not useful.
But since we really want to delete the `supplierinfo` of archived products (that's the purpose of the commit), I think the product being archived should be explicitly a part of the condition for the supplierinfo to be deleted (i.e. be part of `_condition_supplierinfo_unlink`).

Consider this snippet where the supplierinfo might have to be just kept, and not updated nor deleted:
```py
if supplierinfo_needs_update:
    supplierinfo.write(...)
elif supplierinfo_needs_unlink:
    supplierinfo.unlink()
```
if the supplierinfo does not need to be updated (because its values are fine already, for instance), then we would end up deleting it.

HekkiMelody: Added `or (not self.active)` to `_condition_supplierinfo_unlink` to ensure deletion of supplierinfo related to products that have been archived

product_supplierinfo_intercompany/tests/test_supplier_intercompany.py
SirPyTech: Now the test fails without the related change, thanks!

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
SirPyTech: And in the right commit of course, thanks!

product_supplierinfo_intercompany/models/product_intercompany_supplier_mixin.py
```diff
@@ -62,8 +62,9 @@ def _condition_supplierinfo_create_or_update(self, pricelist, supplierinfo):
...
         )
 
     def _condition_supplierinfo_unlink(self, pricelist, supplierinfo):
         self.ensure_one()
-        return bool(supplierinfo)
+        return bool(supplierinfo) or (not self.active)
```
SirPyTech, line 72: **thought**: This `bool(supplierinfo)` smells a bit in my opinion because it is saying that we want to delete the `supplierinfo` every time it is _truthy_, i.e. it exists.

This should always be `True`, but in the rare (if not impossible) case when it isn't True but `self` is not `active`, then we want to delete this not-existing `supplierinfo` 😵‍💫.

I understand I asked for this in https://github.com/OCA/multi-company/pull/859/changes/516b472ac6648dac1d767bf85f48d61716cd863a#r2606687413, and I am still quite sure the `active` condition should be here for the reasons explained in that thread, but I wouldn't know how to make it more readable; I think I'm missing something.
Maybe `(supplierinfo and not self.active) or supplierinfo`? But that would make the `self.active` condition redundant.

Anyway, `unlink`ing an empty recordset can be done and simply does nothing, so this isn't an issue.

SirPyTech (approved):

francesco-ooops (approved):
Functional ok!

@OCA/intercompany-maintainers can this be merged please? thanks!

francesco-ooops:
@OCA/intercompany-maintainers ping :)

ivs-cetmix:
Hi @HekkiMelody , thank you for your contribution! Shouldn't some non-cherry-picked commits be squashed before we merge this PR?

HekkiMelody:
Hi @ivs-cetmix thank you for your time.

I personally think having small commits each with a single logical change makes it easier to both review the PR and review the history of the module via git later on, so I would prefer to keep them if possible, since otherwise there would only a couple very very large commits with lots of unrelated changes.

However, I'm happy to squash my commits if it's necessary for the merge. Please let me know your view on it. Thanks!

ivs-cetmix:
> I personally think having small commits each with a single logical change makes it easier to both review the PR and review the history of the module via git later on, so I would prefer to keep them if possible, since otherwise there would only a couple very very large commits with lots of unrelated changes.
> 
> However, I'm happy to squash my commits if it's necessary for the merge. Please let me know your view on it. Thanks!

@HekkiMelody I totally agree with you as I would also prefer to have changes in separate commits. However this is not about my personal preferecnes, this is about following the OCA rules and recommendations which require commits to be squashed prior to merge. 
If you can squash some of them while also updating the commit message - this would be perfect. Not mandatory to squash each and every though, and of cause cherry-picked commits should remain as is.

HekkiMelody:
> @HekkiMelody I totally agree with you as I would also prefer to have changes in separate commits. However this is not about my personal preferecnes, this is about following the OCA rules and recommendations which require commits to be squashed prior to merge. If you can squash some of them while also updating the commit message - this would be perfect. Not mandatory to squash each and every though, and of cause cherry-picked commits should remain as is.

I have squashed however many commits I could (while keeping one commit per module) because I want this PR merged and not open a debate.

But I'd like to note that I can't find any rule or recommendation "which require commits to be squashed prior to merge."

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#7git

> Only make a single commit per logical change set.

It was also recently discussed in the contributor mailing list, for example here: https://odoo-community.org/groups/contributors-15/contributors-2164121?mode=thread&date_begin=&date_end=

And I think this squashing makes the PR actively worse, since the commits were curated and grouped by logical change.

ivs-cetmix:
@HekkiMelody thank you! Once again - I totally share your approach however there are rules we all need to follow. 
/ocabot merge patch

PR #861, opened 2025-07-18
[14.0][FIX] product_multi_company: variant company related to template company

HekkiMelody:
This PR fixes an issue with the `company_ids` field on `product.product` not being aligned with the same field on the related `product.template`

Issue found while using `purchase_sale_inter_company`: during the automated creation of the SO in company A from the PO in company B, the `company_ids` field on the product template was correct (showing both companies), but the same field on the variant was not (showing only company B), leading to an error due to inconsistent companies (sale order line in company A supposedly had a product from company B. check_company was failing).

Considered other solutions such as `api.depends_context("company")` which also seemed to work, but this seems to simplify while maintaining the same functionality.

### DIFF
--- a/product_multi_company/models/product.py
+++ b/product_multi_company/models/product.py
@@ -1,7 +1,7 @@
 # Copyright 2015-2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from odoo import api, fields, models
+from odoo import fields, models
 
 
 class ProductProduct(models.Model):
@@ -11,12 +11,5 @@ class ProductProduct(models.Model):
     company_ids = fields.Many2many(
         string="Companies",
         comodel_name="res.company",
-        relation="product_variant_companies_rel",
-        compute="_compute_company_ids",
-        store=True,
+        related="product_tmpl_id.company_ids",
     )
-
-    @api.depends("product_tmpl_id.company_ids")
-    def _compute_company_ids(self):
-        for rec in self:
-            rec.company_ids = [(6, 0, rec.product_tmpl_id.company_ids.ids)]


francesco-ooops (approved):
Functional ok!

product_multi_company/models/product.py
```diff
@@ -11,12 +11,5 @@ class ProductProduct(models.Model):
     company_ids = fields.Many2many(
         string="Companies",
         comodel_name="res.company",
-        relation="product_variant_companies_rel",
-        compute="_compute_company_ids",
-        store=True,
+        related="product_tmpl_id.company_ids",
```
SirPyTech, line 14: **issue**: Please keep `relation` and `store` attributes:
- `store` helps with performance, and allows to search by this field because only stored fields can be searched.
- `relation` lets Odoo see that the values for this field already exist in the DB and they are in that table, so it avoids recomputation of this field's values when the module is updated with this change.

SirPyTech, line 14: **suggestion**: In order to keep the same behavior as before, please also add `readonly=True` otherwise the user might change the companies of a variant not knowing that it will also change the companies of all the other variants.
I see that it is already `readonly` in the variant's form but I'd rather be explicit about it.

product_multi_company/models/product.py
```diff
@@ -11,12 +11,5 @@ class ProductProduct(models.Model):
...
     )
-
-    @api.depends("product_tmpl_id.company_ids")
-    def _compute_company_ids(self):
-        for rec in self:
-            rec.company_ids = [(6, 0, rec.product_tmpl_id.company_ids.ids)]
```
SirPyTech, line 22: **question**: I don't see how the `related`'s computation (https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/odoo/fields.py#L532) could be better than this one, do you have any insight on why the `related` works but this doesn't?

SirPyTech (changes requested):
Thanks for the PR!
I see there is quite the history on this part of code, both in https://github.com/OCA/multi-company/issues/585 and https://github.com/OCA/multi-company/pull/590, maybe @pedrobaeza that joined those conversations would be interested in this one too?

Is it possible to have a minimal use-case to reproduce the error without additional modules? `purchase_sale_inter_company` installs a lot of modules that might interfere.

Is it possible to add a test? That would prevent future regressions and more discussions like "it works on my machine".

PR #862, opened 2025-07-28
[17.0][IMP] purchase_sale_stock_inter_company: Prevent manual validation in…

carlos-lopez-tecnativa:
… purchase reception if it is a dropshipping

To avoid adding an explicit dependency, check the picking type code when it is a dropship.

TT54929
@Tecnativa @pedrobaeza @metaminux @JordiMForgeFlow could you please review this?

### DIFF
--- a/purchase_sale_stock_inter_company/models/stock_picking.py
+++ b/purchase_sale_stock_inter_company/models/stock_picking.py
@@ -23,10 +23,8 @@ class StockPicking(models.Model):
         """
         res = super()._compute_state()
         for picking in self.filtered(
-            lambda pick: pick.picking_type_code == "incoming"
+            lambda pick: pick._is_intercompany_reception()
             and pick.state not in ["done", "cancel"]
-            and pick.purchase_id
-            and pick.purchase_id.intercompany_sale_order_id
         ):
             # intercompany_picking_id is set when the picking is validated
             # meanwhile, the state should remain in 'waiting'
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

purchase_sale_stock_inter_company/models/stock_picking.py
```diff
@@ -48,7 +48,7 @@ def button_validate(self):
         ):
             dest_company = record.partner_id.commercial_partner_id.ref_company_ids
             if (
-                dest_company and record.picking_type_code == "incoming"
+                dest_company and record.picking_type_code in ["incoming", "dropship"]
```
metaminux: Hi @carlos-lopez-tecnativa 

I'm not really comfortable with the conditions used in this file from the beginning...

- in `button_validate`, we use `partner_id` and `picking_type_code`
- in `_action_done`, we use `location_dest_id.usage` and `sale_id.auto_purchase_order_id`
- in `_action_done_intercompany_actions`, we _"use both the SO-PO link and the intercompany_picking_id link"_

It seems to me that what we want to achieve here is that only "inter company deliveries" should be validated and the "incoming shipment counterpart" is synchronized (if enabled by configuration) and eventually blocked for manual validation.

But "deliveries" can be :
- one picking from several if using multi-step deliveries or advanced routes (with push rules, for example)
- returns from an incoming shipment
- dropship pickings where the SO is an inter company one

It seems to me these 3 cases are covered just fine by this PR for preventing the manual validation
(depending on the picking type for returns, and because partner for dropship pickings is the supplier) but I know returns aren't synchronized by _action_done (I must create an issue about that...)

I think blocking manual validation of incoming shipments is the way to get a bijection between inter company pickings but I also think we should map inter company stock moves to achieve synchronization properly.

Any thoughts ? @Tecnativa @pedrobaeza @JordiMForgeFlow

carlos-lopez-tecnativa: I must confess that I agree with you; I see the same logic in several places. So, now I have tried to unify this logic into two functions. Please let me know if it is cleaner now.

I decided to use the locations to cover dropshipping and deliveries in multiple steps. But as you mentioned, returns are not supported in any version of this module. Feel free to create the respective issue or PR.

pedrobaeza:
Isn't the `sale_line_id` / `purchase_line_id` links the way to determine things?

carlos-lopez-tecnativa:
> Isn't the `sale_line_id` / `purchase_line_id` links the way to determine things?

In this case, I think it is better to use fields at the picking level. The fields you mentioned are at the stock.move level, but I consider it unnecessary if this can be determined at the picking level. All moves of the same picking are intercompany.

pedrobaeza:
Yes, but I'm saying if there's no way to determine if a picking is "intercompany", like in return. It was just a thought.

purchase_sale_stock_inter_company/models/stock_picking.py
```diff
@@ -205,3 +197,25 @@ def _notify_picking_problem(self, purchase):
...
+        """
+        Check if the picking is an inter-company reception.
+        Consider the customer location(dropship case) to avoid
+        explicitly adding a dependency on the stock_dropshipping module.
+        :return: bool
+        """
+        return (
+            self.location_dest_id.usage in ["internal", "customer"]
+            and self.purchase_id.sudo().intercompany_sale_order_id
+        )
```
metaminux:
```suggestion
    def _is_intercompany_reception(self):
        """
        Check if the picking is an inter-company reception.
        :return: bool
        """
        return (
            self.location_id.usage == "supplier"
            and self.purchase_id.sudo().intercompany_sale_order_id
        )
```

Can't we check source location instead of destination for reception ? This would cover dropship ones too...

purchase_sale_stock_inter_company/models/stock_picking.py
metaminux: This feels way better to me :)

I'll create an issue and a PR for returns using _is_intercompany_delivery...

Thanks

metaminux (changes requested):

purchase_sale_stock_inter_company/models/stock_picking.py
carlos-lopez-tecnativa: Done

metaminux (approved):

PR #864, opened 2025-07-28
[FIX] purchase_sale_inter_company: Create sale from other company

SirPyTech:
This change fixes the following use case:
1. Install `product_multi_company` and `purchase_sale_inter_company`
2. Create 2 companies: **A** and **B**
4. Configure the companies so that a purchase from **A** triggers a sale from **B**
6. Create a product **Product** and set both companies in its `company_ids`
5. Activate only company **A**
7. Create a purchase in **A** with supplier **B**: it will trigger a sale in **B**
8. Confirm the purchase

**Expected behavior**:
A sale in **B** is created

**Actual behavior**:
Error:
> Companies not compatible
> \<Name of sale order line created in **B**> belongs to company **B** but **Product** belongs to another company

**Additional context**:
I cannot reproduce the issue in runboat, but the automated test (that does not need `product_multi_company`) makes sure that the root cause won't happen again.

The root cause is that the product's `company_ids` is read as only **A** while in a non-sudo environment and stored in cache, even if its real **sudo**ed value is **A**, **B**:
```
>>> product.company_ids
res.company(17,)  # A

>>> product.sudo().company_ids
res.company(17)  # Still A: using sudo does not help

>>> product.invalidate_cache()
None

>>> product.sudo().company_ids
res.company(1, 17)  # A, B

>>> product.company_ids
res.company(1, 17)  # A, B
```

### DIFF
--- a/purchase_sale_inter_company/models/purchase_order.py
+++ b/purchase_sale_inter_company/models/purchase_order.py
@@ -3,6 +3,8 @@
 # Copyright 2018-2019 Tecnativa - Carlos Dauden
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+from contextlib import contextmanager
+
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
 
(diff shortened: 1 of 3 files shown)


francesco-ooops (approved):
Functional ok!

HekkiMelody (approved):
Code review, LGTM

PR #867, opened 2025-08-18
[18.0][MIG] account_reconcile_model_multicompany_propagate

Xino61122:
Migration of the module from v16.0 to v18.0.
    
Module for a multi-company environment. This module allows for the
propagation of reconciliation models to be shared across all companies
in the database.

We can check the functionality in this video
https://www.loom.com/share/0c4f37fb6be84558a565c426cd059d29

MT-11040 @moduon


account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
```diff
@@ -0,0 +1,258 @@
...
+            "effect": {
+                "type": "rainbow_man",
+                "message": _(
+                    "Congratulations! Has been created/update %s reconciles models"
+                )
+                % len(companies_chart_template_equal),
+            }
+        }
+
+    def _get_account_map_propagate(self, companies):
```
EmilioPascual, line 130: Something happens with the accounts, they aren't propagated correctly.
image
You should update this method for get the right account

EmilioPascual (changes requested):
@Xino61122 thank you for the job. I found a bug with the accounts.

One more thing: the version is not usually included in commits, and the title of the pull request should be [18.0][MIG] account_reconcile_model_multicompany_propagante.

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
```diff
@@ -0,0 +1,256 @@
...
+            return defaultdict(dict)
+        accounts_map = defaultdict(dict)
+
+        for com_id in companies.ids:
+            accounts = (
+                self.env["account.account"]
+                .with_company(com_id)
+                .search(
+                    [
+                        ("company_ids", "in", companies.ids),
```
EmilioPascual:
```suggestion
```
Why search in all companies but limit to one using `with_company`

EmilioPascual (changes requested):
Good job @Xino61122, now the accounts are propagated.

account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py
Xino61122: If the selected company is not there, the search does not work correctly. 
However, it is true that the com_id could be passed directly in the query so that it does not perform an extensive search. I will change it now.
Thank you for checking it 😄.

EmilioPascual (approved):

ArantxaSudon (commented):
Good job @Xino61122 The reconciliation models are propagated to the other companies.

I don't approve it because I think there is still a problem with the propagation of accounts.

There are two options:

1.- The accounting account I use in the reconciliation model exists in the other companies --> It propagates perfectly.

2.- The accounting account I use exists in the company where I create the model but not in the others --> The model propagates but instead of the accounting account, it only puts the name I have given it in the source company but does not take any account because it does not exist in the chart of accounts, so the model does not work


imagen

I suggest that you indicate that the accounting account used does not exist in the other companies or that you create it with the same account number.

Xino61122:
Hello @ArantxaSudon.
Thank you for checking how the module works.
As you say, if the account does not exist in another company, and you propagate it from where the account does exist, it does not give any warning, and creates the record but empty.

Currently, it works the same as in v16.0, so it should not be a barrier to migration; we have also opened an issue so that we can improve it.

Thank you very much for detecting this functionality.

account_reconcile_model_multicompany_propagate/__manifest__.py
```diff
@@ -3,9 +3,8 @@
 
 {
     "name": "Account Reconcile Model Multicompany Propagate",
-    "summary": {
-        "Propagate account reconcile model in companies " "with same chart template"
-    },
+    "summary": "Propagate account reconcile model in companies "
+    "with same chart template",
```
chienandalu: Nicer IMO:
```suggestion
    "summary": (
        "Propagate account reconcile model in companies with same chart template"
    ),

File: account_reconcile_model_multicompany_propagate/models/account_reconcile_model.py [hunk 4]
```diff
@@ -139,7 +140,7 @@ def _get_account_map_propagate(self, companies):
                 .with_company(com_id)
                 .search(
                     [
-                        ("company_ids", "in", companies.ids),
+                        ("company_ids", "in", com_id),
```
chienandalu: I don't know about the logic, but this is going to fail if not in an enumerator object:
```suggestion
                        ("company_ids", "in", [com_id]),

File: account_reconcile_model_multicompany_propagate/views/account_reconcile_model.xml [hunk 5]
```diff
@@ -0,0 +1,56 @@
...
+                        confirm="This will copy this model to other companies where you have access, if they share the same chart of accounts. If they already have a model with the same name, it will be updated in place. Are you sure to proceed?"
+                    />
+                </header>
+                <div
+                    groups="account.group_account_user"
+                    class="alert alert-info mb-0"
+                    role="alert"
+                    invisible="same_name_other_companies == False"
+                >
+                    <field name="same_name_other_companies" invisible="1" />
```
chienandalu: Remeber that now you can get rid of some auxiliar fields in the views
```suggestion

**Review by chienandalu — CHANGES_REQUESTED** [review]
Just code review

**Comment by Xino61122** [comment]
Hello @chienandalu,

Thank you very much for your review. I have uploaded the changes as you suggested, and everything is still working, but with cleaner code.

I appreciate your contribution.

**Review by chienandalu — APPROVED** [review]
Just code review

File: account_reconcile_model_multicompany_propagate/__manifest__.py [hunk 6]
```diff
@@ -0,0 +1,24 @@
...
+    "name": "Account Reconcile Model Multicompany Propagate",
+    "summary": (
+        "Propagate account reconcile model in companies with same chart template"
+    ),
+    "version": "18.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Accounting/Accounting",
+    "website": "https://github.com/OCA/multi-company",
+    "author": "Moduon, Odoo Community Association (OCA)",
+    "maintainers": ["EmilioPascual"],
```
yajo: **issue:** change this please:


```suggestion
    "maintainers": ["rafaelbn", "EmilioPascual"],
```

yajo (changes requested):

account_reconcile_model_multicompany_propagate/__manifest__.py
Xino61122: It's do it, thank you.

yajo (approved):

rafaelbn (approved):
/ocabot merge patch

rafaelbn:
Hello, @yajo could please ping the maintainers of this repo? we are maintainter of this module so also we could fix the BUG in the bot in order to merge this one. Thank you!

yajo:
@OCA/intercompany-maintainers could you please merge this migration?

[We are maintainers](https://github.com/OCA/multi-company/blob/d4499ea12b2c4a62fd05535427e25ce71d3b9395/account_reconcile_model_multicompany_propagate/__manifest__.py#L12), but due to [a bot issue](https://github.com/OCA/oca-github-bot/issues/226), [we cannot merge](https://github.com/OCA/multi-company/pull/867#issuecomment-3250196947).

rafaelbn (approved):
Functionally tested 👍🏼 

/ocabot migration account_reconcile_model_multicompany_propagate
/ocabot merge patch

PR #791, opened 2025-03-05
[18.0] [MIG] product_multi_company: Migration to 18.0

PieterPaulussen:
Depends on:

- [ ] #790

PR #869, opened 2025-08-22
[18.0][MIG] product_multi_company: Migration to 18.0

ArnauCForgeFlow:
Standard Migration to version 18.0

Supersedes: https://github.com/OCA/multi-company/pull/791


product_multi_company/migrations/17.0.1.1.0/pre-migrate.py
LoisRForgeFlow, line 1: Move this to 18.0.1.0.0

product_multi_company/migrations/17.0.1.1.0/pre-migrate.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+PREVIOUS_DOMAIN = [
+    "|",
+    ("company_ids", "in", "COMPANY_IDS"),
+    ("company_ids", "=", False),
+]
+
+
+UPSTREAM_DOMAIN = " [('company_id', 'in', company_ids + [False])]"
```
LoisRForgeFlow: Also update this to match the upstream domain in 18.0 https://github.com/odoo/odoo/blob/d4886faf12ccaf63d5e899c20df2543d1ce046ab/addons/product/security/product_security.xml#L37

LoisRForgeFlow (changes requested):
We still want to keep the migration script, there could be cases in which you reach 18.0 without the rule updated.

pedrobaeza:
@LoisRForgeFlow which cases don't have the migration script run? Are you using enterprise upgrade system?

product_multi_company/migrations/17.0.1.1.0/pre-migrate.py
ArnauCForgeFlow: @LoisRForgeFlow done

LoisRForgeFlow:
> @LoisRForgeFlow which cases don't have the migration script run? Are you using enterprise upgrade system?

Indeed. But it can be the case also if you use OU from v17 with a non-up-to-date v17 oca/multi-company

LoisRForgeFlow (approved):
LGTM now :+1:

BhaveshHeliconia (approved):
Functional review LGTM!

PR #860, opened 2025-07-17
[18.0][MIG] purchase_sale_inter_company

HviorForgeFlow:
Standard migration

Includes: 

- [x] https://github.com/OCA/multi-company/pull/602

CC @ForgeFlow

PR #871, opened 2025-08-23
[17.0][FW][IMP] purchase_sale_inter_company: add more ctx to update locked sales

metaminux:
Port from 16.0 to 17.0:
- #602

### DIFF
--- a/purchase_sale_inter_company/models/purchase_order_line.py
+++ b/purchase_sale_inter_company/models/purchase_order_line.py
@@ -31,10 +31,14 @@ class PurchaseOrderLine(models.Model):
         """Sync lines between an confirmed unlocked purchase and a confirmed unlocked
         sale order"""
         lines = super().create(vals_list)
+        allowed_states = self._get_allowed_sale_order_states()
         for order in lines.order_id.filtered(
             lambda x: x.state == "purchase" and x.intercompany_sale_order_id
         ):
-            if order.intercompany_sale_order_id.sudo().state in {"cancel", "done"}:
+            if (
+                order.intercompany_sale_order_id.sudo().state,
+                order.intercompany_sale_order_id.sudo().locked,
+            ) not in allowed_states:
                 raise UserError(
                     _(
                         "You can't change this purchase order as the corresponding "
@@ -87,7 +91,10 @@ class PurchaseOrderLine(models.Model):
         ).sudo()
         if not sale_lines:
             return res
-        closed_sale_lines = sale_lines.filtered(lambda x: x.state != "sale")
+        closed_sale_lines = sale_lines.filtered(
+            lambda x: (x.state, x.order_id.locked)
+            not in self._get_allowed_sale_order_states()
+        )
         if closed_sale_lines:
             raise UserError(
                 _(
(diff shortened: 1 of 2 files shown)


metaminux:
I had to adapt `_get_allowed_sale_order_states` because state "done" doesn't exist anymore (see odoo/odoo@17bece3)

Those changes were taken from #860 by @HviorForgeFlow

purchase_sale_inter_company/models/purchase_order_line.py
```diff
@@ -31,10 +31,14 @@ def create(self, vals_list):
...
         sale order"""
         lines = super().create(vals_list)
+        allowed_states = self._get_allowed_sale_order_states()
         for order in lines.order_id.filtered(
             lambda x: x.state == "purchase" and x.intercompany_sale_order_id
         ):
-            if order.intercompany_sale_order_id.sudo().state in {"cancel", "done"}:
+            if (
+                order.intercompany_sale_order_id.sudo().state,
+                order.intercompany_sale_order_id.sudo().locked,
```
pedrobaeza, line 40: I don't think we should handle `locked` flag, but only the state. What do you think?

metaminux: Hi Pedro,

Not sure to understand... This PR is forwarding #602 to 17.0.
The purpose is to allow updating locked sales on specific flows (using ctx).

As "done" state has been replaced by "sale" state with `locked` flag set as `True`, I don't see how to allow bypassing without handling the `locked` state...

Do you mean only looking for `state` in `_get_allowed_sale_order_states` and checking `locked` flag separately ?

pedrobaeza: No, I mean not checking `locked` at all. That doesn't impact in this module IMO.

metaminux: Hello @pedrobaeza,

I looked further on the subject. It turns out, @QuocDuong1306 needed a way to create new lines on intercompany sales orders when using **purchase_sale_container_deposit_inter_company** 
But **purchase_sale_inter_company** is checking that `state` is not `done` before creating a sale order line.

In 17.0, `done` state has disappeared so locked and unlocked sales orders both have `state` = `sale`

I checked with a small test and nothing prevents creating a line on a locked sale order... (not sure if it was the case in 16.0)

So, it seems this PR is just not useful at all in 17.0. The only benefit we could get is to make it possible for other modules to extend states allowed in a sale order when synchronizing from a purchase...

If @QuocDuong1306 or @santostelmo are OK, I'll close that PR soon.

FYI, this is the small test I wrote :

```
    def test_create_line_in_locked_state(self):
        self.intercompany_sale_user_id.groups_id += self.env.ref(
            "sale.group_auto_done_setting"
        )
        self.intercompany_sale_user_id.groups_id += self.env.ref(
            "sales_team.group_sale_salesman"
        )
        with Form(self.env["sale.order"].with_user(self.intercompany_sale_user_id)) as sale_form:
            sale_form.partner_id = self.env.ref("base.res_partner_2")
            with sale_form.order_line.new() as line:
                line.product_id = self.product
                line.product_uom_qty = 1.0
        sale = sale_form.record
        self.assertEqual(sale.state, "draft")
        self.assertFalse(sale.locked)
        sale.action_confirm()
        self.assertEqual(sale.state, "sale")
        self.assertTrue(sale.locked)
        self.assertEqual(len(sale.order_line), 1)
        self.env["sale.order.line"].create(
            {
                "product_id": self.product.id,
                "product_uom_qty": 3.0,
                "order_id": sale.id,
            }
        )
        self.assertTrue(sale.locked)
        self.assertEqual(len(sale.order_line), 2)
```

pedrobaeza: Waiting then for their answer to close it.

santostelmo: On At first glance, it seems okay to close this PR.

PR #872, opened 2025-08-27
[18.0][IMP] account_invoice_inter_company: propagate date to new invoice

carlos-lopez-tecnativa:
When creating an invoice with a date earlier than the current date, the invoice generated on the C2 side does not preserve the same date, instead, it recalculates the date and uses the current date. This commit ensures that the created invoice inherits the same date as the original invoice.

FWP of https://github.com/OCA/multi-company/pull/815

TT55902

@Tecnativa @pedrobaeza @victoralmau could you please review this?

### DIFF
--- a/account_invoice_inter_company/models/account_move.py
+++ b/account_invoice_inter_company/models/account_move.py
@@ -252,6 +252,7 @@ class AccountMove(models.Model):
             "ref": self.name,
             "payment_reference": self.payment_reference,
             "invoice_date": self.invoice_date,
+            "date": self.date,
             "invoice_origin": _("%(company_name)s - Invoice: %(invoice_name)s")
             % {"company_name": self.company_id.name, "invoice_name": self.name},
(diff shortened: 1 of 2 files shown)


victoralmau (approved):

pedrobaeza (changes requested):
I'm now thinking that you may have the accounting on the other company already locked for the accounting date, so this should be checked and put the date according that.

carlos-lopez-tecnativa:
> I'm now thinking that you may have the accounting on the other company already locked for the accounting date, so this should be checked and put the date according that.

This case is already handled by Odoo in https://github.com/odoo/odoo/blob/02086a062ea538b3908337e3ad2f2c996c301229/addons/account/models/account_move.py#L4992-L4994 so I think there’s nothing else to do.

pedrobaeza (approved):
OK then.

/ocabot merge patch

PR #811, opened 2025-04-02
[18.0][MIG] product_tax_multicompany_default: Migration to 18.0

Qlasta:

PR #873, opened 2025-09-01
[18.0][MIG] product_tax_multicompany_default: Migration to 18.0

Xino61122:
This PR is a continuation of PR https://github.com/OCA/multi-company/pull/811.

This PR is for the migration of the product_tax_multicompany_default module, which is used to propagate taxes across multiple companies.

This commit makes changes so that it is not duplicated in the Sales Taxes form view, and removes the functionality when creating a new product in all companies, since Odoo 18 already does this.

https://www.loom.com/share/825a048b696f436cb77a38f6ad5b1ae5

@moduon MT-11041


EmilioPascual (changes requested):
@Xino61122 now the error doesn't happen but the behavior has been lost, when you create a product, taxes aren't propagated automatically.

IMHO deleting a feature doesn't fix the bug, you just lose the behavior.

Xino61122:
> @Xino61122 now the error doesn't happen but the behavior has been lost, when you create a product, taxes aren't propagated automatically.
> 
> IMHO deleting a feature doesn't fix the bug, you just lose the behavior.

Thank you very much for reviewing the code @EmilioPascual. 

Currently, when you create a product, it is propagated with the corresponding taxes to the other companies, so this functionality is not necessary and has therefore been removed.

When you want to propagate taxes to other companies, that's when you need the module to be able to propagate them to the other companies.

That's why I think the current creation function complicates the logic of the module.

Am I wrong about anything?

Shide (changes requested):
Lost functionality of propagating taxes when creating a product:

NOK: Propagate taxes when creating a product
OK: Propagate taxes when clicking on the button

https://www.loom.com/share/974968a060bf4048a0d76f053dc37f41?sid=1c8aa431-c201-4492-b337-2e11f26a7197

product_tax_multicompany_default/models/product.py
```diff
@@ -155,13 +155,6 @@ def set_multicompany_taxes(self):
             }
         )
 
-    @api.model_create_multi
-    def create(self, vals_list):
-        new_products = super().create(vals_list)
-        for product in new_products:
-            product.set_multicompany_taxes()
-        return new_products
-
```
chienandalu: @Xino61122 here's how you can fix you issues with setting the taxes on the `create`:

```suggestion
    @api.model_create_multi
    def create(self, vals_list):
        new_products = super().create(vals_list)
        for product in new_products:
            product.set_multicompany_taxes()
        new_products.invalidate_recordset(fnames=["taxes_id", "supplier_taxes_id"])
        return new_products

File: product_tax_multicompany_default/models/product.py [hunk 2]
```diff
@@ -0,0 +1,171 @@
...
+            # Skip single-company products
+            if one.company_id:
+                continue
+            # A unique constraint in account.tax makes it impossible to have
+            # duplicated tax names by company
+            customer_taxes = {
+                frozenset(tax.name for tax in one.taxes_id if tax.company_id == company)
+                for company in all_companies
+            }
+            if len(customer_taxes) > 1:
```
Shide: Functional test: 
If taxes are empty in Company 1 and filled in Company 2, button to propagate taxes in Company 1 (empty tax) doesn't appear.

Check this to ensure all compatible companies have the same taxes.

Shide (changes requested):

product_tax_multicompany_default/models/product.py
EmilioPascual: IMHO this behavior is an improvement to the module and is outside the scope of the migration. What do you think @Shide?

EmilioPascual (approved):

product_tax_multicompany_default/models/product.py
Shide: I think should be in this module. The goal is to ensure all companies with the same taxes has the same taxes, so if you leave this field to blank, should be blank on all companies.

chienandalu (approved):

product_tax_multicompany_default/models/product.py
```diff
@@ -24,37 +23,105 @@ class ProductTemplate(models.Model):
...
+    def is_divergent_company_taxes(self, one, tax_type):
+        """Returns true or false if there are differences in product taxes.
+
+        :param one: product to be checked
+        :param tax_type: can be 'taxes' or 'purchase'
+        """
+        dict_tax_type = {
+            "taxes": "taxes",
+            "purchase": "purchase",
+        }
```
Shide: This is not needed. In this function, names are the same.

product_tax_multicompany_default/models/product.py
```diff
@@ -24,37 +23,105 @@ class ProductTemplate(models.Model):
...
+        for cid in other_company_ids:
+            data = taxes_bd.get(cid)
+            if not data:
+                other_names.add(None)
                 continue
+            if isinstance(data, dict):
+                other_names.update(data.values())
+            else:
+                other_names.update(data)
+        if field_name == dict_tax_type.get("taxes"):
```
Shide: compare directly to the name of the field (same on purchase)

product_tax_multicompany_default/models/product.py
```diff
@@ -24,37 +23,105 @@ class ProductTemplate(models.Model):
...
-            if len(supplier_taxes) > 1:
-                one.divergent_company_taxes = True
+
+            one.divergent_company_taxes = self.is_divergent_company_taxes(
+                one, "purchase"
+            )
+            if one.divergent_company_taxes:
+                continue
+
+    def is_divergent_company_taxes(self, one, tax_type):
```
Shide: Make it private with _
Rename parameters as something more readable.

product_tax_multicompany_default/models/product.py
```diff
@@ -24,37 +23,105 @@ class ProductTemplate(models.Model):
...
+            current_names = set(
+                one.supplier_taxes_id.filtered(
+                    lambda t: t.company_id == current_company
+                ).mapped("name")
+            ) or {None}
+
+        return not current_names.issubset(other_names)
+
+    def _get_product_taxes(self, field, product_id):
+        """Returns taxes on purchase or sales taxes
```
Shide, line 116: Explan why this operation should be done with an sql.

I think this can be done with `self.ensure_one()`

product_tax_multicompany_default/models/product.py
```diff
@@ -24,37 +23,105 @@ class ProductTemplate(models.Model):
...
+                    ("id", "not in", self.env.context.get("ignored_company_ids", []))
+                ]
+            )
+            .ids
+        )
+        other_company_ids = set(all_companies)
+        other_company_ids.discard(current_company.id)
+
+        other_names = set()
+        for cid in other_company_ids:
```
Shide: Add comments to know what you are doing? When they are really needed.

Shide (changes requested):
Much better!
Some issues :)

product_tax_multicompany_default/models/product.py
Xino61122: I have changed the way I use it.

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+            "Does this product have divergent cross-company taxes? "
+            "(Only for multi-company products)"
+        ),
+    )
+
+    @api.depends("company_id", "taxes_id", "supplier_taxes_id")
+    def _compute_divergent_company_taxes(self):
+        """Know if this product has divergent taxes across companies."""
+        for one in self:
+            one.divergent_company_taxes = False
```
Shide: Apply update for all recordset

```suggestion
        """Know if this product has divergent taxes across companies."""
        self.divergent_company_taxes = False
        for one in self:
```

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+            # Skip single-company products
+            if one.company_id or len(one.env["res.company"].search([]).ids) <= 1:
```
Shide: Move `if len(one.env["res.company"].search([]).ids) <= 1` to the top of the function. This need to be checked only once.

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+                continue
+            # A unique constraint in account.tax makes it impossible to have
+            # duplicated tax names by company
+            one.divergent_company_taxes = self._is_divergent_company_taxes(one, "taxes")
```
Shide: Call it from product

```suggestion
            one.divergent_company_taxes = one._is_divergent_company_taxes("taxes")
```

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+            )
+            if one.divergent_company_taxes:
+                continue
+
+    def _is_divergent_company_taxes(self, current_product, tax_type):
+        """Returns true or false if there are differences in product taxes.
+
+        :param one: product to be checked
+        :param tax_type: can be 'taxes' or 'purchase'
+        """
```
Shide: If you are only to use one product, make it available direclty from product. Otherwise, use  @api.model

Update parameters help


```suggestion
    def _is_divergent_company_taxes(self, tax_type):
        """Returns true or false if there are differences in product taxes.

        :param tax_type: can be 'taxes' or 'purchase'
        """
        self.ensure_one()
```

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+        """Returns taxes on purchase or sales taxes
+
+        We need the product taxes for other companies, and we cannot take
+        them from the object itself, so we have to consult the database
+        where the product taxes passed by parameter are located for all
+        companies.
+
+        :param field: can be 'taxes' or 'purchase'
+        :param product_id: product to which the query should be applied
+        """
```
Shide: Make it available only for one product

```suggestion
    def _get_product_taxes(self, field):
        """Returns taxes on purchase or sales taxes

        We need the product taxes for other companies, and we cannot take
        them from the object itself, so we have to consult the database
        where the product taxes passed by parameter are located for all
        companies.

        :param field: can be 'taxes' or 'purchase'
        """
        self.ensure_one()
```

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def taxes_by_company(self, field, company, match_tax_ids=None):
+        taxes_ids = []
+        if match_tax_ids is None:
+            taxes_ids = company[field].ids
+        # If None: return default taxes
+        if not match_tax_ids:
+            return taxes_ids
+        type_tax_use = "sale" if field == "account_sale_tax_id" else "purchase"
+        AccountTax = self.env["account.tax"]
```
Shide: No camel case in python. Only in class names

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+        for tax in AccountTax.browse(match_tax_ids):
```
Shide: Use a `.filtered_domain` instead of using a search inside the loop (where has been browsed before)

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+        sql = f"""
+            SELECT DISTINCT a_tax.company_id, a_tax.name
+            FROM {table_name} ptr
+            LEFT JOIN account_tax a_tax ON tax_id = a_tax.id
+            WHERE ptr.prod_id = %s
+            """
+        self.env.cr.execute(sql, [product_id])
+        return self.env.cr.fetchall()
+
+    def taxes_by_company(self, field, company, match_tax_ids=None):
```
Shide: Make it private (_), add docstring

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+        self.env.cr.execute(customer_sql + ";", customer_sql_params)
+        # Delete supplier taxes
+        supplier_sql = "DELETE FROM product_supplier_taxes_rel WHERE prod_id IN %s"
+        supplier_sql_params = [tuple(self.ids)]
+        if excl_supplier_tax_ids:
+            supplier_sql += tax_where
+            supplier_sql_params.append(tuple(excl_supplier_tax_ids))
+        self.env.cr.execute(supplier_sql + ";", supplier_sql_params)
+
+    def set_multicompany_taxes(self):
```
Shide, line 200: Add docstring

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+        :param excl_supplier_tax_ids: Excluded supplier tax ids
+        """
+        tax_where = " AND tax_id NOT IN %s"
+        # Delete customer taxes
+        customer_sql = "DELETE FROM product_taxes_rel WHERE prod_id IN %s"
+        customer_sql_params = [tuple(self.ids)]
+        if excl_customer_tax_ids:
+            customer_sql += tax_where
+            customer_sql_params.append(tuple(excl_customer_tax_ids))
+        self.env.cr.execute(customer_sql + ";", customer_sql_params)
```
Shide:
```suggestion
        self.env.cr.execute(customer_sql, customer_sql_params)
```

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+        # Delete supplier taxes
+        supplier_sql = "DELETE FROM product_supplier_taxes_rel WHERE prod_id IN %s"
+        supplier_sql_params = [tuple(self.ids)]
+        if excl_supplier_tax_ids:
+            supplier_sql += tax_where
+            supplier_sql_params.append(tuple(excl_supplier_tax_ids))
+        self.env.cr.execute(supplier_sql + ";", supplier_sql_params)
```
Shide:
```suggestion
        self.env.cr.execute(supplier_sql, supplier_sql_params)
```

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+        user_company = self.env.company
+        customer_tax = self.taxes_id
+        customer_tax_ids = customer_tax.ids
+        if not customer_tax.filtered(lambda r: r.company_id == user_company):
+            customer_tax_ids = []
+        supplier_tax = self.supplier_taxes_id
+        supplier_tax_ids = supplier_tax.ids
+        if not supplier_tax.filtered(lambda r: r.company_id == user_company):
+            supplier_tax_ids = []
+        obj = self.sudo()
```
Shide: Try to not use sudo here.

Use it in `_taxes_by_company`

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,257 @@
...
+            list(customer_tax_ids)
+            if default_customer_tax_ids != customer_tax_ids
+            else None
+        )
+        match_suplier_tax_ids = (
+            list(supplier_tax_ids)
+            if default_supplier_tax_ids != supplier_tax_ids
+            else None
+        )
+        for company in obj.env["res.company"].search([("id", "!=", user_company.id)]):
```
Shide: I think the user can only set taxes for companies that are allowed to change. You don't need sudo here (in obj).

Also, check _delete_product_taxes to only remove from companies that are allowed to change.

Shide (changes requested):

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,258 @@
...
+        """Know if this product has divergent taxes across companies."""
+        # Skip single-company products
+        if self.company_id or len(self.env["res.company"].search([]).ids) <= 1:
+            return
+        self.divergent_company_taxes = False
+        for one in self:
+            # A unique constraint in account.tax makes it impossible to have
+            # duplicated tax names by company
+
+            one.divergent_company_taxes = one._is_divergent_company_taxes("taxes")
```
Shide: Simplify and change self on purchase `_is_divergent_company_taxes`. If first one is True, second one won't be evaluated.

```suggestion
            one.divergent_company_taxes = one._is_divergent_company_taxes("taxes") or one._is_divergent_company_taxes("purchase")
```

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,258 @@
...
+        all_companies = self.env["res.company"].search(
+            [
+                # Useful for tests, to avoid pollution
+                ("id", "not in", self.env.context.get("ignored_company_ids", [])),
+            ]
+        )
+        current_company = self.env.company
+
+        taxes_bd = dict(self._get_product_taxes(tax_type))
+        for k in list(taxes_bd.keys()):
```
Shide: Rename "k" to a meaningfull word. This applies to the other functions

```suggestion
        for company_id in list(taxes_bd.keys()):
```

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,258 @@
...
+    def _taxes_by_company(self, field, company, match_tax_ids=None):
+        taxes_ids = []
+        if match_tax_ids is None:
+            taxes_ids = company[field].ids
+        # If None: return default taxes
+        if not match_tax_ids:
+            return taxes_ids
+        type_tax_use = "sale" if field == "account_sale_tax_id" else "purchase"
+        account_tax = self.env["account.tax"].sudo().browse(match_tax_ids)
+        taxes_ids = account_tax.filtered_domain([("type_tax_use", "=", type_tax_use)])
```
Shide: Rename to a word that doesn't contain _ids. This name looks like the variable conatains a list of ids instead of records. Check out all code for this pattern

```suggestion
        all_taxes = self.env["account.tax"].sudo().browse(match_tax_ids)
        type_use_taxes = account_tax.filtered_domain([("type_tax_use", "=", type_tax_use)])
```

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,258 @@
...
+        if not taxes_ids:
+            return taxes_ids
+        taxes_name = list({n for n in taxes_ids.mapped("name") if n})
+        taxes_ids = account_tax.search(
+            [
+                ("type_tax_use", "=", type_tax_use),
+                ("company_id", "=", company.id),
+                ("name", "in", taxes_name),
+            ]
+        ).ids
```
Shide: type_tax_use has been already filtered
taxes_name will have the same data because comes from type_use_taxes. I'm missing something here?

```suggestion
        return type_use_taxes.filtered_domain([("company_id", "=", company.id)]).ids
```

Shide (changes requested):

product_tax_multicompany_default/models/product.py
Xino61122: We need to do a search for the taxes of other companies that are in the current record set.

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,261 @@
...
+    )
+
+    @api.depends("company_id", "taxes_id", "supplier_taxes_id")
+    def _compute_divergent_company_taxes(self):
+        """Know if this product has divergent taxes across companies."""
+        # Skip single-company products
+        if self.company_id or len(self.env["res.company"].search([]).ids) <= 1:
+            self.divergent_company_taxes = False
+            return
+        self.divergent_company_taxes = False
```
Shide:
```suggestion
        self.divergent_company_taxes = False
        if self.company_id or len(self.env["res.company"].search([]).ids) <= 1:
            return

```

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,261 @@
...
+        self.ensure_one()
+        all_companies = self.env["res.company"].search(
+            [
+                # Useful for tests, to avoid pollution
+                ("id", "not in", self.env.context.get("ignored_company_ids", [])),
+            ]
+        )
+        current_company = self.env.company
+
+        dict_product_taxes_bd = dict(self._origin._get_product_taxes(tax_type))
```
Shide: "dict" word at the end. First the object, then the type.
Or make it more readable like the suggestion.

```suggestion
        company_tax_bd_map = dict(self._origin._get_product_taxes(tax_type))
```

product_tax_multicompany_default/models/product.py
```diff
@@ -0,0 +1,261 @@
...
+        them from the object itself, so we have to consult the database
+        where the product taxes passed by parameter are located for all
+        companies.
+
+        :param field: can be 'taxes' or 'purchase'
+        """
+        if not self.id:
+            return []
+
+        self.ensure_one()
```
Shide: If someone pass multiple recordset to the function, i think `self.id` crashes.

```suggestion
        if not self.ids:
            return []

        self.ensure_one()
```

Shide (changes requested):

product_tax_multicompany_default/models/product.py
Shide: This has not been addresset on all the cases.

Shide (changes requested):
This text doesn't appear in .pot file:
```
"This button propagates sales and purchase taxes if they have changed and if "
"the companies have the same tax rate"
```

Please update .pot translation file too

Xino61122:
> This text doesn't appear in .pot file:
> 
> ```
> "This button propagates sales and purchase taxes if they have changed and if "
> "the companies have the same tax rate"
> ```
> 
> Please update .pot translation file too

Its do it, thank you to review it

Shide (approved):

PR #885, opened 2025-10-06
[19.0] [MIG] res_company_search_view: Migration to 19.0

rven:


weinni2000 (approved):
LGTM

res_company_search_view/models/res_company.py
```diff
@@ -0,0 +1,29 @@
...
+                    "res_partner",
+                    SQL(
+                        "%s = %s",
+                        SQL.identifier(partner_alias, "id"),
+                        SQL.identifier(alias, "partner_id"),
+                    ),
+                )
+            return SQL.identifier(partner_alias, "country_id")
+        else:
+            return super()._field_to_sql(alias, field_expr, query)
```
legalsylvain, line 29:
```suggestion
    country_id = fields.Many2one(store=True)
```

This should be enough (and more simple).

Don't you think ? 

thanks !

legalsylvain (changes requested):
a remark inline. Otherwise, LGTM.

thanks for porting this module, and fixing the error (that exists since many versions).

res_company_search_view/models/res_company.py
rven: I don't think so, in `odoo/addons/base/models/res_company.py` the country is defined as

```python
    country_id = fields.Many2one('res.country', compute='_compute_address', inverse='_inverse_country', string="Country")
```

This means that all these fields also need to be stored?
```python
    street = fields.Char(compute='_compute_address', inverse='_inverse_street')
    street2 = fields.Char(compute='_compute_address', inverse='_inverse_street2')
    zip = fields.Char(compute='_compute_address', inverse='_inverse_zip')
    city = fields.Char(compute='_compute_address', inverse='_inverse_city')
```

> 2026-04-13 06:34:30,476 57574 WARNING res_company_search_view py.warnings: /opt/odoo/oca-19.0/OCB/odoo/orm/registry.py:537: UserWarning: res.company: inconsistent 'compute_sudo' for computed fields street, street2, zip, city, state_id, country_id. Either set 'compute_sudo' to the same value on all those fields, or use distinct compute methods for sudoed and non-sudoed fields.

legalsylvain: The objective of this module is to add a search view, so other module can use it. (because odoo does'nt add any on res.company).
searching on country is not the main objective. But, well, I'm not against to add such feature, just i'd like to avoid huge part of code. 
Indeed, all others fields should be stored. But in fact, it makes sense. I mean, why not search on city ? 

WDYT ?

rven: Seems to work fine, I also tested this with the base_address_extended module installed and the enforce cities enabled.

legalsylvain (commented):
Thanks for the simplification.

I'm a little annoyed by the design this module is trivial : 
- it add a simple view
- it store fields. 
The coverage is natively 100%.

However, your PR come with +200 lines of test. (python / JS). I don't see it usefull, and I think it will make the module harder to maintain. I propose to remove all that tests. What do you think ?

thanks !

legalsylvain (approved):
thanks !

res_company_search_view/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,6 @@
+This module simply add a search view for the `res.company` model.
+
+It is used to be extended in other modules.
+
+Technically, it is an adaptation of the `res.partner` search view.
+(`base.view_res_partner_filter`)
```
legalsylvain, line 6: could you update the description, to mention that new fields are searchable ? 

thanks !

legalsylvain:
Thanls !
/ocabot merge patch

PR #887, opened 2025-10-09
[16.0][FIX] partner_multi_company: Keep partner companies aligned

SirPyTech:
**Steps**:
1. In an empty DB with demo data install `l10n_it`, `partner_multi_company`, `stock`

**Actual Behavior**:
```
2025-10-06 10:55:13,769 1 WARNING ? odoo.modules.loading: Module l10n_it demo data failed to install, installed without demo data 
  File "/opt/odoo/custom/src/odoo/odoo/modules/[loading.py](https://loading.py/)", line 91, in load_demo
    load_data(cr, idref, mode, kind='demo', package=package)
  File "/opt/odoo/custom/src/odoo/odoo/modules/[loading.py](https://loading.py/)", line 73, in load_data
    tools.convert_file(cr, package.name, filename, idref, mode, noupdate, kind)
  File "/opt/odoo/custom/src/odoo/odoo/tools/[convert.py](https://convert.py/)", line 771, in convert_file
    convert_xml_import(cr, module, fp, idref, mode, noupdate)
  File "/opt/odoo/custom/src/odoo/odoo/tools/[convert.py](https://convert.py/)", line 837, in convert_xml_import
    obj.parse(doc.getroot())
  File "/opt/odoo/custom/src/odoo/odoo/tools/[convert.py](https://convert.py/)", line 757, in parse
    self._tag_root(de)
  File "/opt/odoo/custom/src/odoo/odoo/tools/[convert.py](https://convert.py/)", line 717, in _tag_root
    raise ParseError(msg) from None  # Restart with "--log-handler odoo.tools.convert:DEBUG" for complete traceback
odoo.tools.convert.ParseError: while parsing /opt/odoo/auto/addons/l10n_it/demo/demo_company.xml:25
The partner must have at least all the companies associated with the user.
View error context:
'-no context-'
```

**Expected behavior**:
Demo data of `l10n_it` is loaded

**Additional context**:
This happens because `stock` creates a new company with admin user (see https://github.com/odoo/odoo/blob/5a991ffaf50d3a12d628bd70e6b2f2c9eb912d84/addons/stock/data/stock_demo.xml#L161), but the field `res.users.company_ids` is only updated at cache level (in https://github.com/odoo/odoo/blob/5a991ffaf50d3a12d628bd70e6b2f2c9eb912d84/odoo/fields.py#L4907) so `res.users.write` is not called so the new company is not added to the partner's companies as it happens when writing to `res.users.company_ids`.

The following `write` on `res.users.company_ids` done by `l10n_it`'s demo data (in https://github.com/odoo/odoo/blob/5a991ffaf50d3a12d628bd70e6b2f2c9eb912d84/addons/l10n_it/demo/demo_company.xml#L27) triggers the validation of the field so the constraint raises a `ValidationErrorì that is downgraded to a WARNING because it's just demo data.

The same happens on write to `res.company.user_ids` but that is way less common because the field is not even shown in the UI.

I believe this is an edge case missed by https://github.com/OCA/multi-company/pull/670 so people involved there might be interested @ArnauCForgeFlow @LoisRForgeFlow @pedrobaeza please let me know what you think.

### DIFF
--- a/partner_multi_company/models/res_company.py
+++ b/partner_multi_company/models/res_company.py
@@ -1,0 +1,24 @@
+# Copyright 2025 Simone Rubino - PyTech
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        companies = super().create(vals_list)
+        # The new company is added to the user
+        # but that does not trigger the user's write
+        # that would align the partner's companies,
+        # so we have to do it explicitly
+        for company in companies:
+            for user in company.user_ids:
+                # If the partner has no companies,
+                # it means it already accepts all the companies
+                # so there is no need to add the new one
+                if user.partner_id.company_ids:
+                    user.partner_id.company_ids += company
+        return companies
--- a/partner_multi_company/models/res_users.py
+++ b/partner_multi_company/models/res_users.py
@@ -9,10 +9,13 @@ class ResUsers(models.Model):
 
     @api.model_create_multi
     def create(self, vals):
-        res = super(ResUsers, self).create(vals)
-        if "company_ids" in vals:
-            res.partner_id.company_ids = vals["company_ids"]
-        return res
+        users = super(ResUsers, self).create(vals)
+        for user in users:
+            # The new user might have a company even if it was not in `vals`
+            # because of defaults for example.
+            if user.company_ids:
+                user.partner_id.company_ids += user.company_ids
+        return users
 
     def write(self, vals):
         res = super(ResUsers, self.with_context(from_res_users=True)).write(vals)
--- a/partner_multi_company/models/res_users.py
+++ b/partner_multi_company/models/res_users.py
@@ -18,7 +18,6 @@ class ResUsers(models.Model):
         return users
 
     def write(self, vals):
-        res = super(ResUsers, self.with_context(from_res_users=True)).write(vals)
         if "company_ids" in vals:
             for user in self.sudo():
                 new_company_ids = []
@@ -41,4 +40,4 @@ class ResUsers(models.Model):
             for user in self.sudo():
                 if user.partner_id.company_ids:
                     user.partner_id.company_ids = [(4, vals["company_id"])]
-        return res
+        return super(ResUsers, self.with_context(from_res_users=True)).write(vals)
--- a/partner_multi_company/tests/test_partner_multi_company.py
+++ b/partner_multi_company/tests/test_partner_multi_company.py
@@ -3,7 +3,7 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 from odoo.exceptions import AccessError
-from odoo.tests import common, tagged
+from odoo.tests import common, new_test_user, tagged
 
 
 @tagged("post_install", "-at_install")
@@ -181,3 +181,10 @@ class TestPartnerMultiCompany(common.TransactionCase):
         )
         self.assertIn(company, admin_user.company_ids)
         self.assertIn(company, admin_user.partner_id.company_ids)
+
+    def test_new_user_partner_company(self):
+        """When creating a new user, its partner has all the user's companies."""
+        new_user = new_test_user(self.env, login="test_new_user_partner_company")
+        companies = new_user.company_ids
+        self.assertTrue(companies)
+        self.assertIn(companies, new_user.partner_id.company_ids)
--- a/partner_multi_company/tests/test_partner_multi_company.py
+++ b/partner_multi_company/tests/test_partner_multi_company.py
@@ -166,3 +166,18 @@ class TestPartnerMultiCompany(common.TransactionCase):
             self.partner_company_none.with_user(self.user_company_1).read(["name"]),
             [{"id": self.partner_company_none.id, "name": "partner without company"}],
         )
+
+    def test_company_creation_with_users(self):
+        """When creating a new company with users,
+        the users and their partners are linked to the new company."""
+        admin_user = self.env.ref("base.user_admin")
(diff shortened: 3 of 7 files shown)


partner_multi_company/tests/test_partner_multi_company.py
```diff
@@ -166,3 +166,18 @@ def test_can_read_partner_without_company(self):
             self.partner_company_none.with_user(self.user_company_1).read(["name"]),
             [{"id": self.partner_company_none.id, "name": "partner without company"}],
         )
+
+    def test_company_creation_with_users(self):
+        admin_user = self.env.ref("base.user_admin")
+        company = self.env["res.company"].create(
+            [
```
pedrobaeza: Why are you adding a list here whilst it's only one dictionary?

SirPyTech: No specific reason, it's just because `create` expects a list as an argument, I can remove it if you prefer.

pedrobaeza: Not really, `create` admits both a dictionary or a list of dictionary.

partner_multi_company/models/res_company.py
```diff
@@ -0,0 +1,17 @@
...
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        companies = super().create(vals_list)
+        for vals, company in zip(vals_list, companies):
+            if "user_ids" in vals:
+                for user in company.user_ids:
+                    user.partner_id.company_ids += company
```
pedrobaeza: No need to over-complicate:

```suggestion
        for company in companies
            for user in company.user_ids:
                    user.partner_id.company_ids += company
```

partner_multi_company/models/res_company.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    @api.model_create_multi
+    def create(self, vals_list):
```
pedrobaeza, line 11: It's also a good habit to document why are you overriding the method:

```suggestion
    def create(self, vals_list):
        # Inject the created company into the allowed companies of users
        # already indicated in the company
```

partner_multi_company/tests/test_partner_multi_company.py
SirPyTech: Right, but that's just for backward compatibility so I usually try to avoid it:
```py
    @api.model_create_multi
    @api.returns('self', lambda value: value.id)
    def create(self, vals_list):
        """ create(vals_list) -> records

        Creates new records for the model.

        The new records are initialized using the values from the list of dicts
        ``vals_list``, and if necessary those from :meth:`~.default_get`.

        :param Union[list[dict], dict] vals_list:
            values for the model's fields, as a list of dictionaries::

                [{'field_name': field_value, ...}, ...]

            For backward compatibility, ``vals_list`` may be a dictionary.
            It is treated as a singleton list ``[vals]``, and a single record
            is returned.
[...]
        """
```
(from https://github.com/odoo/odoo/blob/5a991ffaf50d3a12d628bd70e6b2f2c9eb912d84/odoo/models.py#L3925)

I'd prefer to keep it, but as I said I can remove it if you prefer.

pedrobaeza: Yes, I prefer. It's wasting lines for no reason, and such "backward compatibility" (I wouldn't call it that way) is an essential part of the API, and it's not going to be removed. The polymorphism of the `create` method is on purpose.

partner_multi_company/models/res_company.py
SirPyTech: Right, done, please check

partner_multi_company/models/res_company.py
SirPyTech: :+1: I noticed this in the user's override but not in the company one, thanks! Please check

partner_multi_company/tests/test_partner_multi_company.py
SirPyTech: No problem, it's done :smile:

pedrobaeza (approved):

SirPyTech:
I'm still struggling to coordinate all the tests in the repo but reviews could help in the meanwhile

partner_multi_company/models/res_users.py
```diff
@@ -9,10 +9,13 @@ class ResUsers(models.Model):
 
     @api.model_create_multi
     def create(self, vals):
-        res = super(ResUsers, self).create(vals)
-        if "company_ids" in vals:
-            res.partner_id.company_ids = vals["company_ids"]
-        return res
+        users = super(ResUsers, self).create(vals)
```
chienandalu, line 12: We're stumbling upon some cases where the constraint is raising before being able to ensure the compatibility post actions. So maybe it's useless to perform that check before the following code...

This context would be catched in the constraint, of course :)
```suggestion
        users = super(ResUsers, self.with_context(check_company_id=True)).create(vals)
```

chienandalu (changes requested):

partner_multi_company/models/res_users.py
SirPyTech: I'm not a fan of this approach because I've seen how activating/deactivating features with context can end up in a _flags mess_, but I agree it can be a solution, thanks for proposing it!
I would leave it as a last resort when everything else fails.

As for this particular case I have found how to make the tests happy, and I added a test with the use-case that was failing previously, please check!

chienandalu (approved):
Thanks @SirPyTech

LoisRForgeFlow:
@SirPyTech Could you fwport this to v17 and v18?

SirPyTech:
> @SirPyTech Could you fwport this to v17 and v18?

Yes we have scheduled a task for it, I'll probably get to it this week or the next one.

PR #843, opened 2025-06-13
[18.0][MIG] purchase_sale_stock_inter_company: Migration to 18.0

metaminux:
This migration is based on migration from 16.0 to 17.0 by @cuongnmtm that is still an open PR : #605 

Depends on :
- [x] #757 
- [x] #837 

@Kev-Roche

PR #892, opened 2025-10-20
[18.0][MIG] purchase_sale_stock_inter_company: Migration to version 18.0

carlos-lopez-tecnativa:
- Update the test assertion in the warehouse with multiple steps, because the next picking is created when the first picking is validated.
- Support shared lots between companies

Supersedes https://github.com/OCA/multi-company/pull/843
TT58525


purchase_sale_stock_inter_company/tests/test_inter_company_purchase_sale_stock.py
```diff
@@ -0,0 +1,623 @@
...
+
+        so_picking.with_user(self.user_company_b).button_validate()
+        self.assertEqual(so_picking.state, "done")
+        po_internal_pick_id = po_picking.move_ids.move_dest_ids.picking_id
+        # the move in the receipt should have a "next move" due to "two_steps"
+        self.assertTrue(purchase.picking_ids.move_ids.move_dest_ids)
+        self.assertEqual(len(sale.picking_ids), 3)  # Pick + Backorder + Delivery
+        self.assertTrue(
+            all((po_picking, po_internal_pick_id, so_picking, so_internal_pick))
+        )
```
metaminux:
```suggestion
        po_internal_pick = po_picking.move_ids.move_dest_ids.picking_id
        # the move in the receipt should have a "next move" due to "two_steps"
        self.assertTrue(purchase.picking_ids.move_ids.move_dest_ids)
        self.assertEqual(len(sale.picking_ids), 3)  # Pick + Backorder + Delivery
        self.assertTrue(
            all((po_picking, po_internal_pick, so_picking, so_internal_pick))
        )
```
Only to keep same policy with names in this test, but other tests are using "_id" so not blocking

metaminux (approved):
LGTM

purchase_sale_stock_inter_company/tests/test_inter_company_purchase_sale_stock.py
carlos-lopez-tecnativa: Done, thanks

purchase_sale_stock_inter_company/demo/res_partner_demo.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <!-- Technically a bug in `stock` module's demo data,
+    Just changing these fields to make it easier to test the module
+    The route for location "Inter-warehouse transit" doesn't exist
+    causing an error when creating the intercompany SO with demo data -->
+    <record id="stock.res_partner_address_41" model="res.partner">
```
pedrobaeza: Is this still needed?

purchase_sale_stock_inter_company/models/purchase_order.py
```diff
@@ -0,0 +1,32 @@
...
+from odoo import models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    def _prepare_sale_order_data(
+        self, name, partner, dest_company, direct_delivery_address
+    ):
+        new_order = super()._prepare_sale_order_data(
```
pedrobaeza, line 16: Can you document what this override does (and well, all of them)?

purchase_sale_stock_inter_company/models/stock_move_line.py
```diff
@@ -0,0 +1,99 @@
...
+    @api.model
+    def _get_fields_to_sync_intercompany(self):
+        return {"quantity", "lot_id"}
+
+    def _get_or_create_lot_intercompany(self, dest_company):
+        # search if the same lot exists in destination company
+        self.ensure_one()
+        StockLot = self.env["stock.lot"].sudo()
+        lot = self.lot_id.sudo()
+        dest_lot = StockLot.search(
```
pedrobaeza: Don't perform the search if `not lot.company_id`. Return `lot`...

Or I would say more, that lot should directly be the same and just remove `company_id`.

purchase_sale_stock_inter_company/static/description/icon.png
pedrobaeza, line 1: Is this icon really helpful?

purchase_sale_stock_inter_company/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+# Copyright 2018-2019 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Inter Company Module for Purchase to Sale Order with warehouse",
+    "summary": "Intercompany PO/SO rules with warehouse",
+    "version": "18.0.1.0.0",
+    "category": "Purchase Management",
+    "website": "https://github.com/OCA/multi-company",
+    "author": "Odoo SA, Akretion, Tecnativa, Odoo Community Association (OCA)",
```
pedrobaeza: Why Odoo SA is here? I don't think in this specific module, we are using code from the old v9 AGPL module.

metaminux: Why not ?

This module is a split from purchase_sale_inter_company which was originally in Odoo core. So even if the module was refactored a lot, why not keeping it ?

pedrobaeza: That's not really true... The only one in core was `account_invoice_inter_company` (I participated in the discussion of the original one, that was a backport to 8: https://github.com/OCA/multi-company/tree/8.0/account_invoice_inter_company)

purchase_sale_stock_inter_company/demo/res_partner_demo.xml
carlos-lopez-tecnativa: I reviewed the demo data, and now this record has the correct locations, so I removed this code.

V18 without this code
image
V14
image

purchase_sale_stock_inter_company/models/purchase_order.py
carlos-lopez-tecnativa: Done

purchase_sale_stock_inter_company/models/stock_move_line.py
carlos-lopez-tecnativa: Done

purchase_sale_stock_inter_company/__manifest__.py
carlos-lopez-tecnativa: Done

purchase_sale_stock_inter_company/static/description/icon.png
carlos-lopez-tecnativa: I’m not sure which icon is better, any suggestions?

metaminux: Default icons for OCA modules are here : https://github.com/OCA/maintainer-tools/tree/master/template/module/static/description

carlos-lopez-tecnativa: > Default icons for OCA modules are here : https://github.com/OCA/maintainer-tools/tree/master/template/module/static/description

Thanks

carlos-lopez-tecnativa:
ping @pedrobaeza

purchase_sale_stock_inter_company/models/stock_move_line.py
```diff
@@ -0,0 +1,114 @@
...
+                ("product_id", "=", lot.product_id.id),
+                ("name", "=", lot.name),
+                ("company_id", "in", [False] + [dest_company.id]),
+            ],
+            limit=1,
+        )
+        if not dest_lot:
+            # if it doesn't exist, create it by copying from original company
+            dest_lot = lot.copy({"company_id": dest_company.id, "name": lot.name})
+        return dest_lot
```
pedrobaeza: It makes no sense IMO to look for other lot in the counterpart, or copy it.

```suggestion
        if lot.company_id:
            lot.company_id = False
        return lot
```

purchase_sale_stock_inter_company/models/purchase_order.py
```diff
@@ -12,6 +12,9 @@ class PurchaseOrder(models.Model):
     def _prepare_sale_order_data(
         self, name, partner, dest_company, direct_delivery_address
     ):
+        """
+        Add additional fields when the purchase_stock module is installed.
+        """
```
pedrobaeza: Docstring are not really this way, but:

```suggestion
        """Add additional fields when the purchase_stock module is installed.
        """
```

but anyway, please don't put docstring, but comment, as this is an override, and thus, it will replace the original docstring.

And please check the rest of the changes for fixing the docstring, or replacing it by a comment.

carlos-lopez-tecnativa: Done. Please let me know if it’s OK for you now.

purchase_sale_stock_inter_company/models/stock_move_line.py
pedrobaeza: What about this?

carlos-lopez-tecnativa: I’m not sure. You can have a product with a company, so in this case, the lot cannot be shared between companies. See https://github.com/odoo/odoo/blob/224aec2308bb7725bbba238037ba874c393a2e3b/addons/stock/models/stock_lot.py#L135
Additionally, what happens with databases where the lot already has a company set?

pedrobaeza: If the product has company, then you can't use that product in the other company as well!! And yes, if the lot has company, we remove it because it's a lot for an inter-company operation, so it's legit to take that path.

christian-ramos-tecnativa: IMO it makes sense to force the lot available to other companies and as @pedrobaeza pointed out if the product is not allowed in both companies the next error will be raised.
```
Invalid Operation
Incompatible companies on records:
- “S00028/Stock>Customers” belongs to company “C2 - My Test Company” and “Product” (product_id: 'product by lot C1') belongs to another company.
```

carlos-lopez-tecnativa: I’ve updated the code to ensure that the lot used does not have a company_id set, and I’ve also updated the test.

christian-ramos-tecnativa (approved):
Functional and code review. LGTM

pedrobaeza (approved):
/ocabot merge nobump

carlos-lopez-tecnativa:
@pedrobaeza please retry

PR #900, opened 2025-11-03
[16.0][FIX] partner_multi_company: Fixed wrong usage of create function

baguenth:
Using the wrong parameter in create function leads to, that for res.partner all create-functions called with parameter "vals_list" will break.

TypeError: Base.create() got an unexpected keyword argument 'vals_list'

### DIFF
--- a/partner_multi_company/models/res_users.py
+++ b/partner_multi_company/models/res_users.py
@@ -8,8 +8,8 @@ class ResUsers(models.Model):
     _inherit = "res.users"
 
     @api.model_create_multi
-    def create(self, vals):
-        users = super(ResUsers, self).create(vals)
+    def create(self, vals_list):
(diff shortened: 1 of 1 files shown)


amh-mw (changes requested):
I have never seen a `create` invocation using a keyword argument, _i.e._ `self.env[model].create(vals_list=vals_list)`; always a positional argument, _i.e._ `self.env[model].create(vals_list)`. ~~This isn't wrong _per se_, but add a unit test to cover the exception?~~

pedrobaeza:
@amh-mw this can happen in any third-party module, so I wouldn't request a unit test for this. Let's just accept the change, as it's correct.

amh-mw (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #903, opened 2025-11-11
[18.0][FIX] purchase_sale_inter_company: avoid multicompany issues when the sale order is created and confirmed

carlos-lopez-tecnativa:
Install the `purchase_sale_stock_inter_company` module

Configuration in the companies:
- Sync pickings
- Auto-validate the sale order

image

Enable Dropshipping and Multi-Step Routes in the inventory settings. In the Dropship route, set Company A.
In the Buy route, set Company B.

image

Create a product and, in Company A, enable only the `Dropshipping` route. Set the supplier info with the partner of `Company B`, but do not set another supplier info for Company B (because this product should not automatically generate a purchase in Company B).

image

Create a sale order with any customer, confirm it, and go to the generated purchase order. Try to confirm the purchase order, an error will be displayed:

“There is no matching vendor price to generate the purchase order for this product.”
image

This happens because the sale order in Company B is confirmed automatically, and the code executes with `sudo`, meaning **record rules are not applied**. As a result, the Dropship route is considered, although in this case it should be ignored.

After this commit, record rules are correctly applied, and routes are taken into account according to the record rules for the current company.

TT56365

### DIFF
--- a/purchase_sale_inter_company/models/purchase_order.py
+++ b/purchase_sale_inter_company/models/purchase_order.py
@@ -37,8 +37,8 @@ class PurchaseOrder(models.Model):
                 purchase_order.partner_id.commercial_partner_id.ref_company_ids
             )
             if dest_company and dest_company.so_from_po:
-                purchase_order.with_company(
-                    dest_company.id
+                purchase_order.with_context(
+                    allowed_company_ids=dest_company.ids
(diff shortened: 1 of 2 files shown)


christian-ramos-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #904, opened 2025-11-11
[18.0][IMP] purchase_sale_stock_inter_company: add support for using transit locations between companies

carlos-lopez-tecnativa:
Currently, intercompany transactions are handled as follows: 
**Sale order**: from Stock to Customer
**Purchase order**: from Vendors to Stock

When using Dropshipping, the moves are:
**Sale order**: from Stock to Customer
**Purchase order**: from Vendors to Customer

However, when using a product tracked by serial number, an error occurs because the same product cannot exist in the same location with the same lot, the lot is effectively entered twice into the customer location.

image

To avoid this constraint, the movement of locations is now performed as follows:

**Standard flow:**

**Sale order**: from Stock to Intercompany Location
**Purchase order**: from Intercompany Location to Stock

**With Dropshipping:**

**Sale order**: from Stock to Intercompany Location
**Purchase order**: from Intercompany Location to Customer

TT56365

### DIFF
--- a/purchase_sale_stock_inter_company/models/sale_order_line.py
+++ b/purchase_sale_stock_inter_company/models/sale_order_line.py
@@ -1,0 +1,23 @@
+# Copyright 2025 Tecnativa - Carlos Lopez
+from odoo import models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def _get_location_final(self):
+        partner = self.order_id.partner_id
+        commercial_partner = partner.commercial_partner_id
+        Company = self.env["res.company"].sudo()
+        dest_company = Company.search(
+            [("partner_id", "parent_of", commercial_partner.id)], limit=1
+        )
+        # Check if the partner belongs to another company and use its customer location
+        # instead of the partner's shipping_id location.
+        if (
+            dest_company
+            and dest_company != self.company_id
+            and partner != self.order_id.partner_shipping_id
+        ):
+            return partner.with_company(self.company_id).property_stock_customer
+        return super()._get_location_final()
--- a/purchase_sale_stock_inter_company/models/__init__.py
+++ b/purchase_sale_stock_inter_company/models/__init__.py
@@ -3,3 +3,4 @@ from . import res_company
 from . import res_config
 from . import stock_picking
 from . import stock_move_line
+from . import sale_order_line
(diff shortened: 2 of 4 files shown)


purchase_sale_stock_inter_company/models/sale_order_line.py
```diff
@@ -0,0 +1,23 @@
...
+            [("partner_id", "parent_of", commercial_partner.id)], limit=1
+        )
+        # Check if the partner belongs to another company and use its customer location
+        # instead of the partner's shipping_id location.
+        if (
+            dest_company
+            and dest_company != self.company_id
+            and partner != self.order_id.partner_shipping_id
+        ):
+            return partner.with_company(self.company_id).property_stock_customer
```
christian-ramos-tecnativa, line 22: If I understand correctly, the `property_stock_customer` field should contain the transit location, right? If that’s the case, in my opinion, this should be mentioned in the README, as the module does not update it by it self.

carlos-lopez-tecnativa: Yes, in the sale order of Company B we have:
`partner_id` = Partner of Company A
`partner_shipping_id` = Final Customer of the sale order in Company A

Therefore, the `property_stock_customer` field of `partner_id` should have the `Transit Location`.

However, this is actually the standard behavior, by default, when Odoo creates a new company, it automatically assigns the transit location to the company’s partner. So, the locations are already set that way by default. See https://github.com/odoo/odoo/blob/591584268cfcbad54ee675ea62d047c3900ed629/addons/stock/models/res_company.py#L201

In fact, to use the `customer` and `vendor` locations instead, you would need to change the partner configuration manually.

That’s why I’m not sure what exactly should be updated in the README if the locations already come with the transit location by default.

carlos-lopez-tecnativa: Now, if the `partner_id` currently has the `customer` location, the movements are performed as before this commit, from `Stock` to `Customer` Location in the sale order, and from `Vendors` to `Stock` in the purchase order.

This change does not affect the current configuration of the partner.

It is an alternative to using the standard locations, but if the locations have been modified in the database, it will still work correctly.

Therefore, the change is transparent to the user, it only depends on the partner’s location configuration.

christian-ramos-tecnativa: In that case, is ok the way it is right now

christian-ramos-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #905, opened 2025-11-12
[12.0][IMP] purchase_sale_inter_company: make move deletion overridable

RicardCForgeFlow:
After https://github.com/OCA/multi-company/pull/508

We can experience slow performance when confirming intercompany orders

## Proposal
Do not unlink the original stock move of the purchase order to improve action_confirm performance

### DIFF
--- a/purchase_sale_inter_company/models/sale_order.py
+++ b/purchase_sale_inter_company/models/sale_order.py
@@ -45,6 +45,10 @@ class SaleOrder(models.Model):
                 _('Error. The following lines do not match on'
                   ' the remote order: %s') % "\n".join(unequal_line))
 
+    @api.model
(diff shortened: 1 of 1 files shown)


purchase_sale_inter_company/models/sale_order.py
```diff
@@ -45,6 +45,10 @@ def assert_intercompany_prices_equal(self):
                 _('Error. The following lines do not match on'
                   ' the remote order: %s') % "\n".join(unequal_line))
 
+    @api.multi
```
MiquelRForgeFlow:
```suggestion
    @api.model
```

MiquelRForgeFlow:
Selection_5200

Issue #906, opened 2025-11-12
Issue when validation receptions


YassineTal:
Issue when validation receptions, If the reception contains product(s),tracked by lot, and with company(s) in company_ids , that the current user does not have access to.

## Module
base_multi_company

## Describe the bug
Issue when validation receptions, If the reception contains product(s),tracked by lot, and with company(s) in company_ids , that the current user does not have access to.

## To Reproduce
**Affected versions**:
18.0 might be other versions aswell
1. Product with company_ids company A and company B
2. Product is tracked by lot
3. Current user has only access to company A
4. Create purchase order and validate
5. When validation the reception related to the created purchase order the following message appears

<img width="1910" height="1004" alt="Image" src="https://github.com/user-attachments/assets/8f3f8683-ae1b-4679-8770-06cb0a6456fa" />

**Expected behavior**
Reception should validate with no issue

PR #907, opened 2025-11-12
Fix for receptions validation issue for products tracked by lot

YassineTal:
This is a fix for the issue reported here:
https://github.com/OCA/multi-company/issues/906


pedrobaeza:
This sounds weird, as each user should have read access to companies.

shochstoeger:
We are encountering the same error. It appears to be triggered by the following record rule:
image

LauraCForgeFlow:
I was having the same issue, this PR solved it for me too. 
**Edit:** Actually this solution was causing some problems for me too so I developed another solution in #916.

PR #910, opened 2025-11-27
[18.0][FIX] base_multi_company: 'company_id' use all common companies

LauraCForgeFlow:
Before this commit, we could have wrong access errors on records. As an example, if we had a user with enabled companies A and C (being A the main company) and a record with enabled companies B and C, we would get and access error. 

That's because the `company_id` for the record was set to B (since it did not match with the user's main company, A, it took the first of the enabled companies for the record). But it should be set to C, because it's the first common company between the user and the record, and that would allow us to access the record without any error. Also added new use cases in the testing to check this behavior. Without the code changes in the commit, the test fails.

### DIFF
--- a/base_multi_company/models/multi_company_abstract.py
+++ b/base_multi_company/models/multi_company_abstract.py
@@ -22,23 +22,30 @@ class MultiCompanyAbstract(models.AbstractModel):
     )
 
     @api.depends("company_ids")
-    @api.depends_context("company", "_check_company_source_id")
+    @api.depends_context("companies", "company", "_check_company_source_id")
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow (approved):
Code LGTM

DavidJForgeFlow (approved):

ArnauCForgeFlow (approved):
LGTM!

LauraCForgeFlow:
Hello @pedrobaeza, could we get this merged? Thanks in advance!

pedrobaeza (approved):
/ocabot merge patch

PR #916, opened 2025-12-09
[18.0][FIX] base_multi_company: clean cache _compute_company_id

LauraCForgeFlow:
This error solves the issue https://github.com/OCA/multi-company/issues/906, since trying to fetch cached records (of field `company_ids`) caused some errors. That does not happen anymore with this PR. The other fix for this issue proposed in #907 was causing some problems too and was not entirely correct, so that's why this solution was developed.

Since always invalidating the cache for the `company_ids` field can be problematic (specially in a compute method) we only invalidate it when in fact we get the access error, and with that we fix our issue. In all other cases, we still keep the original behavior of the method, not breaking the compute logic.


LauraCForgeFlow:
Closing in favor of #932.

PR #918, opened 2025-12-12
[14.0][IMP] product_supplier_inter-multi_company: multi-company supplierinfo

HekkiMelody:
Add `company_ids` to supplierinfo in order to make them visible to some but not all companies. This solves an issue with supplierinfo for products shared only among some companies.

When creating intercompany supplierinfo automatically, sync the product's `company_ids` with the `supplierinfo`.

Steps to reproduce:

- product with companies A and B and not C
- generate an intercompany supplierinfo
- With only company C, try loading the supplierinfo list view -> The supplierinfo is visible to company C, even though the product is not.

To Do:
-  [x] migration to recompute all intercompany supplierinfo
-  [x] unit test

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,7 +1,6 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.38
+_commit: v1.41
 _src_path: gh:oca/oca-addons-repo-template
-ci: GitHub
 convert_readme_fragments_to_markdown: false
 enable_checklog_odoo: false
 generate_requirements_txt: true
@@ -18,6 +17,7 @@ org_name: Odoo Community Association (OCA)
 org_slug: OCA
 rebel_module_groups:
 - product_tax_multicompany_default
+- stock_intercompany
 repo_description: Addons for the management of multi company instances as well as
 repo_name: Multi company modules
 repo_slug: multi-company
--- a/product_supplierinfo_intercompany_multi_company/__manifest__.py
+++ b/product_supplierinfo_intercompany_multi_company/__manifest__.py
@@ -2,7 +2,7 @@
     "name": "Product SupplierInfo Intercompany Multi Company",
     "summary": """
     Compatibility of product_multi_company and product_supplierinfo_intercompany""",
-    "version": "14.0.1.1.1",
+    "version": "14.0.1.2.0",
     "category": "Generic Modules/Others",
     "license": "AGPL-3",
     "author": "Ilyas, Ooops404, Odoo Community Association (OCA)",
@@ -11,6 +11,11 @@
         "product_supplierinfo_intercompany",
         "product_multi_company",
     ],
+    "data": [
+        "views/product_supplierinfo_views.xml",
+    ],
     "installable": True,
     "auto_install": True,
+    "post_init_hook": "post_init_hook",
+    "uninstall_hook": "uninstall_hook",
 }
--- a/product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
+++ b/product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
@@ -1,0 +1,11 @@
+from openupgradelib import openupgrade
+
+# pylint: disable=odoo-addons-relative-import
+from odoo.addons.product_supplierinfo_intercompany_multi_company.hooks import (
+    post_init_hook,
+)
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    post_init_hook(env.cr, env.registry)
--- a/product_supplierinfo_intercompany_multi_company/models/product_supplierinfo.py
+++ b/product_supplierinfo_intercompany_multi_company/models/product_supplierinfo.py
@@ -1,0 +1,20 @@
+from odoo import api, fields, models
+
+
+class ProductSupplierinfo(models.Model):
+    _inherit = ["multi.company.abstract", "product.supplierinfo"]
+    _name = "product.supplierinfo"
+    _description = "Supplier Pricelist (Multi-Company)"
+
+    company_ids = fields.Many2many(
+        string="Companies",
+        comodel_name="res.company",
+        compute="_compute_company_ids",
+        store=True,
+    )
+
+    @api.depends("product_id.company_ids", "product_tmpl_id.company_ids")
+    def _compute_company_ids(self):
+        for rec in self.with_context(automatic_intercompany_sync=True):
+            product = rec.product_id or rec.product_tmpl_id
+            rec.company_ids = product.company_ids
--- a/product_supplierinfo_intercompany_multi_company/tests/test_supplier_multicompany.py
+++ b/product_supplierinfo_intercompany_multi_company/tests/test_supplier_multicompany.py
@@ -1,0 +1,80 @@
+from odoo.addons.product_supplierinfo_intercompany.tests.test_supplier_intercompany import (
+    TestIntercompanySupplierCase,
+)
+
+
+class TestSupplierMultiCompany(TestIntercompanySupplierCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.other_company = cls.env["res.company"].create({"name": "Other company"})
+
+    def test_supplier_all_companies(self):
+        """
+        Product and supplierinfo belongs to all companies
+        """
+        price = 5
+        self.product_template_4.company_ids = False
+        self._add_item(self.product_template_4, price)
+
+        prod_with_other_company = self.product_template_4.with_company(
+            self.other_company
+        )
+        self._check_supplier_info_for(self.product_template_4, price)
+        self._check_supplier_info_for(prod_with_other_company, price)
+
+        supplierinfo = self._get_supplier_info(self.product_template_4)
+        self.assertFalse(supplierinfo.company_ids)
+
+    def test_supplier_some_companies(self):
+        """
+        Product and supplierinfo belong only to two companies,
+        not available to the third
+        """
+        price = 5
+        companies = self.sale_company + self.purchase_company
+        self.product_template_4.company_ids = companies
+        self._add_item(self.product_template_4, price)
+
+        # cannot use utility function _get_supplier_info
+        # because company_id is hardcoded False
+        # search from sale_company
+        user = self.env.ref("base.user_admin")
+        self.env = self.env(user=user)
+        supplierinfo_model = self.env["product.supplierinfo"]
+        user.write(
+            {
+                "company_id": self.sale_company.id,
+                "company_ids": [(6, 0, self.sale_company.ids)],
+            }
+        )
+        domain = [
+            ("name", "=", self.partner.id),
+            ("intercompany_pricelist_id", "=", self.pricelist_intercompany.id),
+            ("product_tmpl_id", "=", self.product_template_4.id),
+            ("product_id", "=", False),
+        ]
+        supplierinfo = supplierinfo_model.search(domain)
+        self.assertEqual(supplierinfo.company_ids, companies)
+
+        # run search again with other company
+        user.write(
+            {
+                "company_id": self.other_company.id,
+                "company_ids": [(6, 0, self.other_company.ids)],
+            }
+        )
+        supplierinfo.invalidate_cache()
+        supplierinfo = supplierinfo_model.search(domain)
+        self.assertFalse(supplierinfo)
+
+    def test_uninstall(self):
+        from ..hooks import uninstall_hook
+
+        uninstall_hook(self.env.cr, None)
+        rule = self.env.ref("product.product_supplierinfo_comp_rule")
+        domain = (
+            " ['|', ('company_id', '=', user.company_id.id), "
+            "('company_id', '=', False)]"
+        )
+        self.assertEqual(rule.domain_force, domain)
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -18,6 +18,7 @@ jobs:
         with:
           python-version: "3.11"
           cache: 'pip'
+          cache-dependency-path: '.pre-commit-config.yaml'
       - name: Get python version
         run: echo "PY=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
       - uses: actions/cache@v4
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,6 +38,11 @@ repos:
         entry: found a en.po file
         language: fail
         files: '[a-zA-Z0-9_]*/i18n/en\.po$'
+      - id: obsolete dotfiles
+        name: obsolete dotfiles
+        entry: found obsolete files; remove them
+        files: '^(\.travis\.yml|\.t2d\.yml|CONTRIBUTING\.md)$'
+        language: fail
   - repo: https://github.com/oca/maintainer-tools
     rev: f9b919b9868143135a9c9cb03021089cabba8223
     hooks:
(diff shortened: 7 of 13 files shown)


HekkiMelody:
The two commits updating rebel modules and disabling auto-install are due to an incompatibility between `multi-company/product_supplierinfo_intercompany_multi_company` and `product_attribute/product_supplierinfo_group` (which is a dependency of `multi-company/product_supplierinfo_group_intercompany`)

This line https://github.com/OCA/product-attribute/blob/14.0/product_supplierinfo_group/models/product_supplierinfo.py#L29 bypasses the logic behind `base_multi_company` and therefore some access rules no longer apply as expected (the fields `company_id` and `company_ids` of supplierinfo are no longer linked as one would expect)

giarve (approved):

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def _sync_company_ids_on_supplierinfo(env):
+    supplierinfo = (
+        env["product.supplierinfo"]
+        .with_context(automatic_intercompany_sync=True, active_test=False)
+        .search([("intercompany_pricelist_id", "!=", False)])
+    )
+    for si in supplierinfo:
+        si.company_ids = si.product_id.company_ids or si.product_tmpl_id.company_ids
```
SirPyTech: **issue**: Migrating the module creates different data than installing it:

- when the module is migrated, `company_ids` is populated according to variant/template's `company_ids`
- when the module is installed, `company_ids` is empty

Please add a similar script in a `post_init` hook too.
Note that there are some helpers defined in `base_multi_company`, you can see an example of how they are used in https://github.com/OCA/multi-company/blob/a016cb337c14c5e0316ae29a6d52e0cd71fa031f/product_multi_company/hooks.py#L14-L26.
These helpers also update the multi-company security rule that appears to be missing here.

product_supplierinfo_intercompany_multi_company/models/product_supplierinfo.py
```diff
@@ -0,0 +1,10 @@
+# © 2019 Akretion (http://www.akretion.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class ProductSupplierinfo(models.Model):
+    _inherit = ["multi.company.abstract", "product.supplierinfo"]
```
SirPyTech, line 5: **question**: Why not set the supplierinfo's `company_ids` to always stay synced with its product's `company_ids`?

That would be an approach similar to the field `company_ids` of the product variant: https://github.com/OCA/multi-company/blob/a016cb337c14c5e0316ae29a6d52e0cd71fa031f/product_multi_company/models/product.py#L11-L22
And we could drop the new method `_prepare_intercompany_supplier_info`.

SirPyTech, line 5: **suggestion**: This makes sense even without the inter-company logic and could be moved to its own module `product_supplierinfo_multi_company`, what do you think?

With a new module, we could drop the migration script because everything would be in the new module's hooks.

product_supplierinfo_intercompany_multi_company/tests/test_supplier_multicompany.py
```diff
@@ -0,0 +1,72 @@
+# Copyright 2019 Akretion (http://www.akretion.com).
```
SirPyTech: **question**: Is this on purpose or just a copy/paste leftover? Please check the other added file's Copyright too.

.copier-answers.yml
```diff
@@ -18,6 +17,8 @@ org_name: Odoo Community Association (OCA)
 org_slug: OCA
 rebel_module_groups:
 - product_tax_multicompany_default
+- stock_intercompany
+- product_supplierinfo_intercompany_multi_company
```
SirPyTech: **question**: From https://github.com/OCA/multi-company/pull/918#issuecomment-3655876720 I understand that `product_supplierinfo_group_intercompany` should be a new rebel module, why are other modules added here instead?

.copier-answers.yml
```diff
@@ -1,7 +1,6 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.38
+_commit: v1.40
```
SirPyTech: **praise**: Thanks for updating the template!

product_supplierinfo_intercompany_multi_company/__manifest__.py
```diff
@@ -15,5 +15,5 @@
         "views/product_supplierinfo_views.xml",
     ],
     "installable": True,
-    "auto_install": True,
+    "auto_install": False,
```
SirPyTech: **chore**: This being a compatibility module means that it should be installed when its dependencies are installed, otherwise things don't work as expected.
If this is no more the case, please update its documentation (summary and README).

SirPyTech (changes requested):
Thanks for the PR!
**question**: from the PR description:

> Steps to reproduce:
> 
> - product with companies A and B and not C
> - generate an intercompany supplierinfo
> - With only company C, try loading the supplierinfo list view -> The supplierinfo is visible to company C, even though the product is not.

If we manually created the supplierinfo instead of generating it, the same issue would happen.
Maybe the solution is just a new module `product_supplierinfo_multi_company` similar to the other `<model>_multi_company` modules in the repo?

.copier-answers.yml
HekkiMelody: `stock_intercompany` was already de-facto added to rebel modules manually here https://github.com/OCA/multi-company/pull/507/changes/261ca0a0d78aca32094dcb1f15b84d4d767ed636 (by, ehm, me, apparently) without going through copier.

You can see that the diff in `.github/workflows/test.yml` is actually consistent with `stock_intercompany` already being a rebel module, so this is actually bringing `.copier-answers.yml` up to date with the actual existing configuration.

product_supplierinfo_intercompany_multi_company/__manifest__.py
HekkiMelody: I honestly don't remember why I removed it, it's been too long. But I agree, I'll restore the `auto_install`.

product_supplierinfo_intercompany_multi_company/models/product_supplierinfo.py
HekkiMelody: Unfortunately I tested it and I couldn't completely drop `_prepare_intercompany_supplier_info`, but I agree that it's better this way

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
HekkiMelody: Added as suggested in a new commit!

product_supplierinfo_intercompany_multi_company/models/product_supplierinfo.py
HekkiMelody: While I agree in theory that a standalone module would be good, it would also necessarily mean reorganizing the modules and the dependency tree. I think it would be too much, and may introduce unwanted features, on a version that is already quite old.

So I've chosen to keep the changes here in this bridge module.

Please let me know what you think

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
HekkiMelody: Maybe it makes sense to put it in the same commit that makes supplierinfo multi-company in the first place, actually...

product_supplierinfo_intercompany_multi_company/models/product_supplierinfo.py
SirPyTech: Thanks for changing this!
I see, you have to remove the `company_id` that is set in `super` otherwise it does not get computed: it makes sense.
That is probably the only change that would not go in the hypothetical `product_supplierinfo_multi_company`.

SirPyTech: I think it would be more clean to have a separate `product_supplierinfo_multi_company`, and it would also allow to have such behavior when `inter-company` modules are not installed.

But I don't think it's blocking for merge because it can splitted out of the module later, and I agree that could complicate the dependency management.

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
SirPyTech: :+1: thanks

product_supplierinfo_intercompany_multi_company/tests/test_supplier_multicompany.py
SirPyTech: :+1:

.copier-answers.yml
SirPyTech: Right, I didn't look at `.github/workflows/test.yml` close enough because I trusted it was aligned with `copier`'s configuration.

But why is `product_supplierinfo_intercompany_multi_company` added as rebel instead of `product_supplierinfo_group_intercompany`?
Moreover: https://github.com/OCA/multi-company/pull/918#issuecomment-3655876720 says the problem is in the multi-company access rule:

> This line https://github.com/OCA/product-attribute/blob/14.0/product_supplierinfo_group/models/product_supplierinfo.py#L29 bypasses the logic behind `base_multi_company` and therefore some access rules no longer apply as expected (the fields `company_id` and `company_ids` of supplierinfo are no longer linked as one would expect)

but now we are updating that rule: could you please check if the new rebel module is still needed?

product_supplierinfo_intercompany_multi_company/__manifest__.py
SirPyTech: :+1:

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def _supplierinfo_multicompany_rule(env):
+    post_init_hook(env.cr, env)
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    _sync_company_ids_on_supplierinfo(env)
+    _supplierinfo_multicompany_rule(env)
```
SirPyTech: **issue**: The `post_init_hook` executes a query to populate the `company_ids` based on what is in `company_id`: https://github.com/OCA/multi-company/blob/a016cb337c14c5e0316ae29a6d52e0cd71fa031f/base_multi_company/hooks.py#L45-L61
that might add more companies in the ones set by our new `_compute_company_ids`.

If we want our computation to prevail then it should be called after the hook or even better, just update the security rule (calling directly `set_security_rule`).
The same applies for the post init hook in `hooks.py`.

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,24 @@
...
+)
+
+
+def _sync_company_ids_on_supplierinfo(env):
+    supplierinfo = (
+        env["product.supplierinfo"]
+        .with_context(automatic_intercompany_sync=True, active_test=False)
+        .search([("intercompany_pricelist_id", "!=", False)])
+    )
+    supplierinfo._compute_company_ids()
```
SirPyTech: **suggestion**: The computation should happen on all the existing `product.supplierinfo` because that's what will happen after the module is installed.
Moreover, since we are adding a new computed stored field I think Odoo already has computed all of its values, could you please check if this is actually needed?

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,24 @@
...
+
+from odoo.addons.product_supplierinfo_intercompany_multi_company.hooks import (
+    post_init_hook,
+)
+
+
+def _sync_company_ids_on_supplierinfo(env):
+    supplierinfo = (
+        env["product.supplierinfo"]
+        .with_context(automatic_intercompany_sync=True, active_test=False)
```
SirPyTech: **polish**: The `automatic_intercompany_sync` context is redundant because is already assigned in `_compute_company_ids`.
```suggestion
        .with_context(active_test=False)
```

SirPyTech (changes requested):
Thanks for updating the implementation, I understand why you aren't adding a new module `product_supplierinfo_multi_company` and that's not blocking but I have some concerns/questions about the new code.

.copier-answers.yml
HekkiMelody: You were right, it doesn't seem to be necessary to change to add either module to rebel modules

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
HekkiMelody: Yes, the computation will happen automatically on a fresh installation. I removed this part of the migration

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
SirPyTech: :+1: thanks!

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
SirPyTech: :+1: thanks!

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
SirPyTech: :+1:

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,11 @@
...
+
+# pylint: disable=odoo-addons-relative-import
+from odoo.addons.product_supplierinfo_intercompany_multi_company.hooks import (
+    post_init_hook,
+)
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    post_init_hook(env.cr, env)
```
SirPyTech: **note**: `post_init_hook`'s 2nd parameter is `registry`, not an `Environment` but it should not be an issue since it is not used.

SirPyTech (approved):
Only code review: looks good.
Thanks!

product_supplierinfo_intercompany_multi_company/migrations/14.0.1.2.0/post-migrate.py
HekkiMelody: I mistakenly followed this example https://github.com/OCA/product-attribute/blob/14.0/product_state/migrations/13.0.1.0.0/post-migration.py#L11

But I found a better example: https://github.com/OCA/stock-logistics-warehouse/blob/14.0/account_move_line_stock_info/migrations/14.0.2.1.0/pre-migration.py#L16 

So I'll change it even though I tested it and it works, the parameters is not used as you say.

francesco-ooops (approved):
Functional ok!

@OCA/intercompany-maintainers this has been thoroughly reviewed and covered by tests, can it be merged? thanks!

ivs-cetmix:
Hey @HekkiMelody , thank you for your contribution! 
/ocabot merge minor

PR #925, opened 2026-01-08
[18.0][IMP] purchase_sale_stock_inter_company: Add an option to automatically validate the purchase order

carlos-lopez-tecnativa:
When the purchase order is generated from a sale order with a product using the MTO or Dropship route, the purchase order can be automatically confirmed to start the intercompany process.

TT56365
@Tecnativa @pedrobaeza @christian-ramos-tecnativa @metaminux could you please review this?

### DIFF
--- a/purchase_sale_stock_inter_company/models/res_company.py
+++ b/purchase_sale_stock_inter_company/models/res_company.py
@@ -35,3 +35,9 @@ class ResCompany(models.Model):
         string="Block manual validation of picking in the destination company",
     )
     notify_user_id = fields.Many2one("res.users", "User to Notify")
+    purchase_auto_validation = fields.Boolean(
+        string="Inter-company POs Auto Validation",
+        help="When a purchase order is created from a sales order "
+        "using the MTO or Dropship rule for this company, "
+        "it will be automatically validated.",
+    )
--- a/purchase_sale_stock_inter_company/models/res_config.py
+++ b/purchase_sale_stock_inter_company/models/res_config.py
@@ -38,3 +38,10 @@ class InterCompanyRulesConfig(models.TransientModel):
         help="User to notify incase of sync picking failure.",
         readonly=False,
     )
+    purchase_auto_validation = fields.Boolean(
+        related="company_id.purchase_auto_validation",
+        help="When a purchase order is created from a sales order "
+        "using the MTO or Dropship rule for this company, "
+        "it will be automatically validated.",
+        readonly=False,
+    )
--- a/purchase_sale_stock_inter_company/models/stock_rule.py
+++ b/purchase_sale_stock_inter_company/models/stock_rule.py
@@ -1,0 +1,27 @@
+# Copyright 2026 Tecnativa - Carlos Lopez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
+
+    @api.model
+    def _run_buy(self, procurements):
+        last_po = self.env["purchase.order"].search(
+            [],
+            order="id desc",
+            limit=1,
+        )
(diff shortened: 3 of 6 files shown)


purchase_sale_stock_inter_company/models/res_company.py
```diff
@@ -35,3 +35,9 @@ class ResCompany(models.Model):
         string="Block manual validation of picking in the destination company",
     )
     notify_user_id = fields.Many2one("res.users", "User to Notify")
+    purchase_auto_validation = fields.Boolean(
+        string="Purchase Orders Auto Validation",
+        help="When a purchase order is created from a sale order "
```
pedrobaeza:
```suggestion
        help="When a purchase order is created from a sales order "
```

purchase_sale_stock_inter_company/models/res_company.py
```diff
@@ -35,3 +35,9 @@ class ResCompany(models.Model):
         string="Block manual validation of picking in the destination company",
     )
     notify_user_id = fields.Many2one("res.users", "User to Notify")
+    purchase_auto_validation = fields.Boolean(
+        string="Purchase Orders Auto Validation",
```
pedrobaeza:
```suggestion
        string="Inter-company POs Auto Validation",
```

purchase_sale_stock_inter_company/models/res_config.py
```diff
@@ -38,3 +38,11 @@ class InterCompanyRulesConfig(models.TransientModel):
         help="User to notify incase of sync picking failure.",
         readonly=False,
     )
+    purchase_auto_validation = fields.Boolean(
+        related="company_id.purchase_auto_validation",
+        string="Purchase Orders Auto Validation",
```
pedrobaeza: No need of re-defining string for related fields. It's taken from the related field AFAIK.

purchase_sale_stock_inter_company/models/stock_rule.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @api.model
+    def _run_buy(self, procurements):
+        last_po = self.env["purchase.order"].search(
+            [],
+            order="id desc",
+            limit=1,
+        )
+        res = super()._run_buy(procurements)
+        new_po = self.env["purchase.order"].search([("id", ">", last_po.id)])
```
pedrobaeza, line 18: Instead of this, isn't possible to know the POs created by the rule itself? This can be misleading.

carlos-lopez-tecnativa: The `_run_buy` method does not return the generated purchase orders.

On the other hand, I can intercept `action_confirm` on the sale order and reuse the [_get_purchase_orders](https://github.com/odoo/odoo/blob/68e86f364a0594a42ab8e6c92698e3980e42cd19/addons/sale_purchase_stock/models/sale_order.py#L14C9-L14C29) function to determine whether the sale order generated a purchase order, but this implies adding a dependency on `sale_purchase_stock`. What do you think about this option?

christian-ramos-tecnativa:
IMO, this improvement is not related to the purpose of the module. I like the change and what it’s trying to solve, but it might be better suited for its own module, with the ability to configure whether those purchases are confirmed on the fly or at a fixed time via a cron job.
Also because this could be used in non intercompany scenarios

pedrobaeza:
Isn't the inter-company PO the one validated? If not, then I agree with Christian.

carlos-lopez-tecnativa:
> IMO, this improvement is not related to the purpose of the module. I like the change and what it’s trying to solve, but it might be better suited for its own module, with the ability to configure whether those purchases are confirmed on the fly or at a fixed time via a cron job. Also because this could be used in non intercompany scenarios

You’re right, sorry for mixing intercompany flow here.
I’ve opened a new PR that focuses only on this feature:
https://github.com/OCA/sale-workflow/pull/4101

PR #792, opened 2025-03-06
[17.0][MIG] pos_category_multicompany: Migration to 17.0

BhaveshHeliconia:

PR #927, opened 2026-01-12
[17.0][FIX] account_multicompany_easy_creation: Fix AccessError with POS installed

BhaveshHeliconia:
- When point_of_sale is installed, loading a chart template triggers the cleanup of default POS payment methods (point_of_sale.models.chart_template._load). If the user running the Easy Creation wizard (e.g., an Account Manager) does not have POS permissions (specifically point_of_sale.group_pos_manager), this fails with an AccessError.

- This commit adds sudo() to the try_loading call in the wizard. Company creation and COA loading are administrative tasks that should succeed regardless of the user's specific access rights to other installed modules.

### DIFF
--- a/account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
+++ b/account_multicompany_easy_creation/wizards/multicompany_easy_creation.py
@@ -230,7 +230,9 @@ class AccountMulticompanyEasyCreationWiz(models.TransientModel):
         new_company = self.new_company_id.with_context(
             allowed_company_ids=allowed_company_ids
         )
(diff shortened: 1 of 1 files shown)


BhaveshHeliconia:
@legalsylvain : Here is the test case fix related to this PR (https://github.com/OCA/multi-company/pull/792).

BhaveshHeliconia:
Hey @OCA/intercompany-maintainers would be great if someone could have a look at this fix.

pedrobaeza:
But this is something to be fixed at Odoo side, not here. Check the same for 18 at https://github.com/odoo/odoo/pull/240587

PR #930, opened 2026-01-16
[19.0][MIG] res_company_category: Migration to 19.0

kbentaleb:
Depends on #885 (res_company_search_view)


mikecolangelo:
Tested this on 19.0 on a clean DB — installs fine and the tests pass. Hit one crash though:

Creating a company while a `default_category_id` is in the context — which the **Companies** smart button on a category sets — raises:

```
ValueError: Wrong value for res.partner.category_id: <id>
```

`res.company.create()` builds the related `res.partner` inheriting the context, and `category_id` here collides with `res.partner.category_id` (the m2m tags), so the company-category id ends up on the partner's tag field.

I pushed a fix + regression test on top of your branch, feel free to cherry-pick:
https://github.com/mikecolangelo/multi-company/commit/b79d210bd4377706763e97871913de1b842d5e81

It overrides `create` on `res.company` to apply the default to the company and drop the key from the context before the partner is created.

Issue #929, opened 2026-01-16
Incorrect access error on writing products


LauraCForgeFlow:
Some users can get an incorrect access error on editing / writing product templates (and maybe with other models too). We have detected it for version 18.0, not sure if it happens for other versions.

## Module
product_multi_company (although probably related to base_multi_company too).

## To Reproduce
Steps to reproduce the behavior, I reproduced it on runboat.

* Create an environment with 3 or more companies. The admin user can access to all of them (in this case we have 4 companies in total), while another demo user can only access to the first 2 companies.

<img width="1634" height="862" alt="Image" src="https://github.com/user-attachments/assets/3bba701d-2bf6-4673-acd6-15bccb65fa29" />

<img width="1634" height="888" alt="Image" src="https://github.com/user-attachments/assets/070b1999-bf68-4cce-865c-f117a8c507f5" />

* With the admin user, select a product template, enter in its form view and assign it to the first 3 companies.

<img width="1844" height="919" alt="Image" src="https://github.com/user-attachments/assets/9adcb412-3b7b-41e2-9475-d6f7bdb817be" />

* With the demo user, open that product's form. In the companies field, you'll see only the first 2 companies because this user does not have access to the third one. This is expected.

<img width="1905" height="880" alt="Image" src="https://github.com/user-attachments/assets/cfa23d98-fee5-4428-8936-7e3beafa4810" />

* Now try to edit any field in the product. You will get a companies access error, which you should not be getting. The user has access to the product, so they should be able to write it without getting an error.

<img width="1916" height="893" alt="Image" src="https://github.com/user-attachments/assets/06baa63f-8ede-4d9f-8aeb-fd3f3d786ddf" />

**Expected behavior**
The user should be able to write the field in the product template normally, without getting the access error.

**Additional context**
I believe this is an issue related to cache. The `company_ids` field gets read first in a sudo environment (therefore returning the 3 companies). Later, when it is read in a non-sudo environment, it uses the cached value (the 3 companies, but out user only has access to 2 of them). And that's why the access error is thrown, because it tries to read information from the 3 companies but one of them is not accessible.

[...]

PR #932, opened 2026-01-19
[18.0][FIX] base_multi_company: company_ids UID context dependent

LauraCForgeFlow:
Fixes the problem reported in the issue #929, using a similar approach as Odoo to solve a similar problem (this is Odoo's original fix PR --> odoo/odoo#217752).

**Note:** This fix can cause issues with the `company_ids` field tracking and the generated logs. We have proposed a PR to Odoo in order to fix this problem as well --> odoo/odoo#244529.

### DIFF
--- a/base_multi_company/models/multi_company_abstract.py
+++ b/base_multi_company/models/multi_company_abstract.py
@@ -19,6 +19,8 @@ class MultiCompanyAbstract(models.AbstractModel):
     company_ids = fields.Many2many(
         string="Companies",
         comodel_name="res.company",
+        # avoid cache pollution in sudo / non-sudo uses of the field
+        depends_context=("uid",),
     )
 
     @api.depends("company_ids")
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (commented):
@LauraCForgeFlow LGTM, thanks for all the insights!

A small comment related to those insights, could you also add that explanation in the commit message? even adding a small code comment inline pointing to the issue or similar can be a good idea. The reasoning here is that, in some time it will be easier to understand why it has been done by anyone looking at the code if the explanation can be found inline or via git blame and finding your commit, the PR description is usually not the quickest reference.

LauraCForgeFlow:
@LoisRForgeFlow Thanks for the review! I've improved the commit message and added some comments in the code.

LoisRForgeFlow (approved):
Thanks! our "future selves" will like this :)

JoanSForgeFlow (approved):

PR #934, opened 2026-02-03
[FIX] ir_ui_view_multi_company: avoid invalid AND domains with website views

samirGuesmi:
This PR fixes an incompatibility between this module's override of **ir.ui.view._get_inheriting_views_domain()** and the **website** module in Odoo 18.

What happens

1. This module adds a company filter using: return expression.AND([domain, company_domain])
2. When the website module is installed, it also overrides _get_inheriting_views_domain() and does the following:

```
# when rendering for the website we have to include inactive views
# we will prefer inactive website-specific views over active generic ones
if current_website:
    domain = [leaf for leaf in domain if 'active' not in leaf]
return expression.AND([website_views_domain, domain])
```

The intention in website is to remove the ('active','=',True) constraint so inactive website-specific views can override active generic ones.

And then it crashes :bomb: 
When this module wraps the original domain inside expression.AND(), the resulting domain uses prefix operators (&, |).

The website code then removes the ('active','=',True) leaf without rebuilding the boolean structure, which can leave invalid domains such as:

`['&', '|', ('company_ids','=',False), ('company_ids','in',[1])]`

This domain is syntactically invalid and causes errors during module installation (e.g. when loading website_data.xml).


samirGuesmi:
@sbejaoui

yvaucher (approved):

sbejaoui (approved):

PR #936, opened 2026-02-04
[18.0][FIX] account_invoice_inter_company: Handle invoice validation in bulk from the wizard

carlos-lopez-tecnativa:
FW-Port of https://github.com/OCA/multi-company/pull/935
When the invoice is validated using the wizard
https://github.com/odoo/odoo/blob/d129570727f050b4b8d26db6951bd5dbf768d09f/addons/account/wizard/account_validate_account_move.py#L26

it calls the `_post` method instead of `action_post`, so the intercompany invoices are not created.

TT60710
@Tecnativa @pedrobaeza @victoralmau could you please review this?

image

### DIFF
--- a/account_invoice_inter_company/models/account_move.py
+++ b/account_invoice_inter_company/models/account_move.py
@@ -46,9 +46,9 @@ class AccountMove(models.Model):
         if not supplier_invoice.ref:
             supplier_invoice.write({"ref": self.name})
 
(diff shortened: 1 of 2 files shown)


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #954, opened 2026-02-20
[18.0][MIG] sale_purchase_inter_company : Migration to 18.0

metaminux:


christian-ramos-tecnativa:
IMO It would be better to add this logic into the existing module `purchase_sale_inter_company`, as it seems like many of the fields used are the same. It will require updating the README about this new feature and reuse/adapt the code but should be better in the long run

metaminux:
> IMO It would be better to add this logic into the existing module `purchase_sale_inter_company`, as it seems like many of the fields used are the same. It will require updating the README about this new feature and reuse/adapt the code but should be better in the long run

Hello @christian-ramos-tecnativa 

For sure this module is very similar to `purchase_sale_inter_company` as it has been developed as the "perfect" mirror of the first one.
But I'm not sure everybody using `purchase_sale_inter_company` is interested by this one...
The use case is really specific as it implies one company is "force selling" to the other one.

PR #955, opened 2026-02-20
[18.0][FIX] product_tax_multicompany_default - set taxes on all companies

enriquemartin:
When a user who does not have access to all existing companies sets taxes for a product, it only propagates the tax to the companies to which it has access, leaving the value empty for those to which it does not have access (the `_delete_product_taxes()` method executes a `DELETE` and skips security rules).

Therefore, it is necessary to use `sudo()` when iterating through the companies to which the tax is to be propagated.

### DIFF
--- a/product_tax_multicompany_default/models/product.py
+++ b/product_tax_multicompany_default/models/product.py
@@ -230,7 +230,9 @@ class ProductTemplate(models.Model):
             if default_supplier_tax_ids != supplier_tax_ids
             else None
         )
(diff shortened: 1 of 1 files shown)


metaminux:
Egun on Enrique,

You could rebase on 18.0 to get rid of red checks...

chienandalu:
can you check @Shide ?

Shide (approved):

chienandalu (approved):

EmilioPascual (approved):

BhaveshHeliconia (approved):
LGTM!

PR #958, opened 2026-02-24
[18.0][ADD] rma_inter_company: New module

victoralmau:
New module

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT57882


juancarlosonate-tecnativa (commented):
Following the flow and validating the RMA picking in Company A, the following error appears: 

Image

rma_inter_company/models/rma.py
```diff
@@ -0,0 +1,210 @@
...
+        return super()._get_location_final()
+
+    def _get_intercompany_rma_company(self):
+        """This method returns the company in which the intercompany RMA will be
+        created. It is possible to use it and modify the value by referring to
+        another field (for example, sale_line_id.auto_purchase_line_id.intercompany_sale_line_id.company_id
+        if purchase_sale_stock_inter_company is used).
+        """  # noqa: E501
+        self.ensure_one()
+        return self.company_id.intercompany_rma_company_id
```
carlos-lopez-tecnativa: When I do not set this field and enable the `Generate Intercompany RMA` field, I get this error.
Image

Image

rma_inter_company/models/rma.py
```diff
@@ -0,0 +1,210 @@
...
+        user = self.company_id.intercompany_rma_user_id
+        warehouse = (
+            self.env["stock.warehouse"]
+            .sudo()
+            .search([("company_id", "=", company.id)], limit=1)
+        )
+        partner = company.sudo().partner_id
+        vals = {
+            "company_id": company.id,
+            "partner_id": partner.id,
```
carlos-lopez-tecnativa: Sorry, but I understand that the partner must be the partner of the origin company, not the destination company. I created the RMA in San Francisco with customer X. This should generate a new RMA in the Chicago company, but the partner should be San Francisco, not Chicago. Or did I misunderstand the flow?

Image

carlos-lopez-tecnativa (changes requested):
As mentioned internally, with the current settings, where in C1 I configure that this RMA must generate a new RMA in C2, these settings do not allow generating an RMA in C3, for example. In other words, from one company I can generate an RMA to only a single destination company, but if I want to generate an RMA from C1 to C2 or from C1 to C3, it is not possible. I think this intercompany setting is limited.

rma_inter_company/models/rma.py
victoralmau: Finally, there is another approach.

rma_inter_company/models/rma.py
victoralmau: Finally, there is another approach: the data is the same as in the RMA that generated it.

rma_inter_company/models/stock_picking.py
```diff
@@ -0,0 +1,106 @@
...
+            )
+            if confirmed_rma_delivery_pickings:
+                # Tip for replace
+                confirmed_rma_delivery_pickings.action_assign()
+            pending_rma_delivery_pickings = rma_delivery_pickings.filtered(
+                lambda x: x.state == "assigned"
+            )
+            if pending_rma_delivery_pickings:
+                pending_rma_delivery_pickings.button_validate()
+        return res
```
carlos-lopez-tecnativa: Could you please split this function into another function to facilitate inheritance if needed?

carlos-lopez-tecnativa (changes requested):
Tested, and it looks fine. However, when I create an RMA from a sales order in C1, complete the whole flow, and go back to the sales order, a new extra line is created. Why?
Image

victoralmau:
> Tested, and it looks fine. However, when I create an RMA from a sales order in C1, complete the whole flow, and go back to the sales order, a new extra line is created. Why? Image

This module is currently being tested in a real-world customer environment, and this behavior does not occur; it is likely related to another module.

I think the best approach would be to merge this feature, and if there are any outstanding compatibility issues or other problems arise, it would be better to address them in a separate pull request after merging this one.

victoralmau:
Ping @juancarlosonate-tecnativa and @carlos-lopez-tecnativa Could you check it again?

It's currently working fine for one of our clients.

rma_inter_company/models/stock_picking.py
```diff
@@ -0,0 +1,123 @@
...
+                lambda x: x.state == "confirmed"
+            )
+            if confirmed_rma_delivery_pickings:
+                # Tip for replace
+                confirmed_rma_delivery_pickings.action_assign()
+            pending_rma_delivery_pickings = rma_delivery_pickings.filtered(
+                lambda x: x.state == "assigned"
+            )
+            if pending_rma_delivery_pickings:
+                pending_rma_delivery_pickings.button_validate()
```
carlos-lopez-tecnativa, line 87: If I change the lot/serial number during the receipt in C1, I would expect that change to be propagated to the corresponding receipt in C2. Please check this function; I imagine you need to implement similar logic here:
https://github.com/OCA/multi-company/blob/a3c0985586a9ad08aa291b3ba2856bcd134249df/purchase_sale_stock_inter_company/models/stock_picking.py#L95-L171


**Receipt of the RMA in C1** 
In this company, I validate the picking, but I change the lot/serial number because the customer specifies which lot is being returned.
Image

**Reception in RMA in C2**
Currently, the lot/serial number change made in C1 is not propagated to the corresponding receipt in C2.
Image

carlos-lopez-tecnativa (changes requested):
I tested this scenario:

- Product tracked by serial number.
- Sold 10 units.
- Completed the intercompany flow normally.
- Went to the sales order in C1 and created an RMA for 5 units.
- Completed the RMA flow in C1.
- Went back to the sales order in C1 and tried to create another RMA.

At this point, **I expected to see only the remaining 5 units available for return**. However, **I see two lines with quantities of 5 and 15, which seems incorrect.**

Could you please review this case?

Image

Image

victoralmau:
Everything related to lots also applies to non-intercompany RMA processes; the solution is to use the `rma_lot` and `rma_sale_lot` modules. This way, pickings (moves) will be restricted, and generating 5 RMAs will generate 5 `stock.move` records, each restricted to lot A, B, C, D, E. Therefore, they can only be validated with that specific lot and no other.

This module should **NOT** cover the workflow you describe because the `rma_lot` and `rma_sale_lot` modules already cover it.

carlos-lopez-tecnativa:
> Everything related to lots also applies to non-intercompany RMA processes; the solution is to use the `rma_lot` and `rma_sale_lot` modules. This way, pickings (moves) will be restricted, and generating 5 RMAs will generate 5 `stock.move` records, each restricted to lot A, B, C, D, E. Therefore, they can only be validated with that specific lot and no other.
> 
> This module should **NOT** cover the workflow you describe because the `rma_lot` and `rma_sale_lot` modules already cover it.

I tested this locally by installing these modules. When I create the RMA, a separate line is created for each lot, which is fine. However, after processing the RMA refund, the returned lots become available for a new RMA again.

Is this the expected behavior? I attached a video for reference. Please let me know if this is correct.

Additionally, why isn't rma_sale_lot an explicit dependency of this module? I think we could add it as a dependency, unless there is a specific reason not to. Could you clarify this?

https://github.com/user-attachments/assets/600221e1-8b8d-4870-8e62-ff867ef08e32

victoralmau:
> > Everything related to lots also applies to non-intercompany RMA processes; the solution is to use the `rma_lot` and `rma_sale_lot` modules. This way, pickings (moves) will be restricted, and generating 5 RMAs will generate 5 `stock.move` records, each restricted to lot A, B, C, D, E. Therefore, they can only be validated with that specific lot and no other.
> > This module should **NOT** cover the workflow you describe because the `rma_lot` and `rma_sale_lot` modules already cover it.
> 
> I tested this locally by installing these modules. When I create the RMA, a separate line is created for each lot, which is fine. However, after processing the RMA refund, the returned lots become available for a new RMA again.
> 
> Is this the expected behavior? I attached a video for reference. Please let me know if this is correct.
> 
> Additionally, why isn't rma_sale_lot an explicit dependency of this module? I think we could add it as a dependency, unless there is a specific reason not to. Could you clarify this?
> rma_inter_company.mp4

Remember what was done at https://github.com/OCA/rma/pull/597

It is important to invoice an item first if you want to issue a refund later, because the receiving move will be linked to the sales order at that time, thereby reducing the quantity delivered. 

Perhaps you haven’t installed `rma_sale_lot`?

The `rma_lot` and `rma_sale_lot` modules should **NOT** be dependencies of this module because not everyone wants to use that functionality (lots), which also has other dependencies (`stock_picking_return_lot`); if someone needs that RMA functionality already present in that repository, they can use it.

victoralmau:
In my opinion, this is ready for merge because it has been tested for some time in a production environment.

carlos-lopez-tecnativa (commented):
Does `action_refund_without_invoice` get synchronized with the other company or not?

In this example, I marked it to be refunded in C1, but the change was not propagated to C2. Is this the expected behavior, or should this be adapted to support the intercompany flow?
C1
Image
C2
Image
Image

juancarlosonate-tecnativa (approved):
LGTM functional and code review. Tested multiple flows and everything looks correct.

rma_inter_company/tests/common.py
```diff
@@ -0,0 +1,112 @@
...
+        )
+        cls.warehouse_b = cls.env["stock.warehouse"].search(
+            [("company_id", "=", cls.company_b.id)]
+        )
+        customer_loc, _supplier_loc = cls.warehouse_a._get_partner_locations()
+        cls.customer_loc = customer_loc
+        cls.warehouse_a.rma_out_replace_route_id = cls.warehouse_a.rma_out_route_id
+        cls.warehouse_b = cls.env["stock.warehouse"].search(
+            [("company_id", "=", cls.company_b.id)]
+        )
```
carlos-lopez-tecnativa, line 26: Please remove the duplicate code.


```suggestion
```

carlos-lopez-tecnativa (approved):
LGTM. If you can update the code to remove the duplicate code, that would be great.

pedrobaeza:
As there's a dedicated repo for RMA, this module should go there instead of going to a general purpose one.

victoralmau:
> As there's a dedicated repo for RMA, this module should go there instead of going to a general purpose one.

Seriously? Now? That could have been mentioned from the start (the PR was created in February). 

The "rule" regarding which repository each module should go into is pretty vague, but... OK, we'll move everything to RMA and update the relevant configuration (to indicate that the PR the customer needs is a different one, and that the module belongs to a different repository).

pedrobaeza:
Sorry for the disturbance. I didn't notice before.

victoralmau:
Moved to https://github.com/OCA/rma/pull/618

carlos-lopez-tecnativa:
What happened to the two additional commits from the other modules? Are they not necessary?

image

victoralmau:
> What happened to the two additional commits from the other modules? Are they not necessary?

They were only necessary when installing all the modules from this repository; in any case, I'm going to create a PR to add them because I think it's useful to have them to avoid side effects (especially in the future) and to analyze something that had already been “fixed.”

PR #959, opened 2026-02-24
[17.0][FIX] partner_multi_company: Add company_ids to partner banks

christian-ramos-tecnativa:
When a partner bank account is created is taking the hidden `company_id`, so the next situation happens:

- Partner visible for companies 2 and 3
- User from company 2 create a bank account (Only visible for company 2)
- User from company 3 is not able to see the bank account and try to create it again so the next error raises:
image

@Tecnativa TT61155
@pedrobaeza @pilarvargas-tecnativa


partner_multi_company/hooks.py
```diff
@@ -18,6 +18,12 @@ def post_init_hook(env):
             ),
         }
     )
+    rule_partner_bank = env.ref("base.res_partner_bank_rule")
+    rule_partner_bank.write(
+        {
+            "domain_force": ("[('company_ids', 'in', company_ids + [False])]"),
```
pedrobaeza: I think this should be new record rules, as they are additive, so we don't hard-replace them.

christian-ramos-tecnativa: When the record rule has no group is restrictive, if added as an extra rule will be computed as AND. So it wouldn't work as expected

pedrobaeza: Oh, true, this one is global...

pedrobaeza (approved):

christian-ramos-tecnativa:
Added new company check to avoid errors i.e when creating mandates

partner_multi_company/models/res_partner_bank.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2026 Tecnativa - Christian Ramos
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3).
+
+from odoo import fields, models
+from odoo.models import BaseModel
+
+
+def to_company_ids(companies):
```
pedrobaeza: Do it private method.

partner_multi_company/models/res_partner_bank.py
```diff
@@ -0,0 +1,48 @@
...
+    return [companies]
+
+
+def check_companies_domain_parent_of(self, companies):
+    """A `_check_company_domain` function that lets a record be used if
+    any company in record.company_ids is a parent of any of the given companies.
+    """
+    if isinstance(companies, str):
+        return [("company_ids", "parent_of", companies)]
+
```
pedrobaeza: Better to not put empty lines inside the method for following current style, and also preferred in general.

christian-ramos-tecnativa:
Done

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

PR #963, opened 2026-02-26
[18.0][FIX] account_multicompany_easy_creation - correct propagation of company_dependent fields

enriquemartin:
I don't know if I'm missing something, but the propagation of `company_dependent` fields is not working correctly.

The current code does not propagate any values, it just goes through the accounts and tax positions and sets the same code or name.

This fix aims to propagate the value of all `company_dependent` fields that point to `account.account` or `account.fiscal.position` following the same logic as in previous versions (when the `ir.property` model existed).


enriquemartin:
@victoralmau could you take a quick look, please? Thanks!

victoralmau (approved):
The change seems to make coherent.

pedrobaeza (approved):
/ocabot merge patch

PR #967, opened 2026-03-03
[17.0][FIX] base_multi_company: Add uid to depends context

christian-ramos-tecnativa:
Steps to reproduce the error:

1. Set a contact with the companies 1 and 2
2. Enter in the contact with Demo, with just one of the companies available
3. Change any field and try to save
4. The next error will be raised because is trying to render the display_name for both companies but Demo only has access to one of them
image

Adding this depends solve the issue, based in the company_ids from account.account 18.0
https://github.com/odoo/odoo/blob/18.0/addons/account/models/account_account.py#L107
This change is not needed in 18.0.

@Tecnativa TT61309
@pedrobaeza @victoralmau

### DIFF
--- a/base_multi_company/models/multi_company_abstract.py
+++ b/base_multi_company/models/multi_company_abstract.py
@@ -19,6 +19,9 @@ class MultiCompanyAbstract(models.AbstractModel):
     company_ids = fields.Many2many(
         string="Companies",
         comodel_name="res.company",
+        depends_context=(
+            "uid",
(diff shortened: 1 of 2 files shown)


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

I understand that this is not needed in 18, ¿right?

christian-ramos-tecnativa:
Exactly @pedrobaeza 

> I understand that this is not needed in 18, ¿right?

PR #809, opened 2025-03-31
[16.0][IMP] account_invoice_inter_company: support display type and retrieve invoice line name in auto invoice

bguillot:
Hello, 

This PR adds :

- Support other display type than product for inter company invoice creation.

- Retreive the invoice line name in the auto invoice

PR #968, opened 2026-03-05
[18.0][IMP] account_invoice_inter_company: support display type and retrieve invoice line name in auto invoice

florian-dacosta:
Keep line description + section and note in the intercompany invoice.

Forward port of https://github.com/OCA/multi-company/pull/809

### DIFF
--- a/account_invoice_inter_company/tests/test_inter_company_invoice.py
+++ b/account_invoice_inter_company/tests/test_inter_company_invoice.py
@@ -285,7 +285,7 @@ class TestAccountInvoiceInterCompanyBase(TransactionCase):
             line_form.quantity = 1
             line_form.product_uom_id = cls.env.ref("uom.product_uom_hour")
             line_form.account_id = cls.a_sale_company_a
-            line_form.name = "Service Multi Company"
+            line_form.name = "Service Multi Company test"
             line_form.price_unit = 450.0
         cls.invoice_company_a = cls.invoice_company_a.save()
         cls.invoice_line_a = cls.invoice_company_a.invoice_line_ids[0]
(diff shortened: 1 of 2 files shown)


florian-dacosta:
@bguillot can you review please ?
@bealdav

rvalyi (approved):

account_invoice_inter_company/tests/test_inter_company_invoice.py
```diff
@@ -382,6 +382,7 @@ def test03_confirm_invoice_and_cancel(self):
             invoice_line.product_id,
             self.invoice_company_a.invoice_line_ids[0].product_id,
         )
+        self.assertEqual(invoice_line.name, self.invoice_line_a.name)
```
pedrobaeza: Please hard-code the expected string, not to put "code" in the assert.

pedrobaeza (changes requested):
Please change the commit message to not include the Odoo version (and more if the version is not correct...).

florian-dacosta:
> Please change the commit message to not include the Odoo version (and more if the version is not correct...).

This was a cherry-pick from v16, hence the version number.
Anyway, I fixed the commit message

account_invoice_inter_company/tests/test_inter_company_invoice.py
pedrobaeza: This is still not resolved.


```suggestion
        self.assertEqual(invoice_line.name, "Service Multi Company test")
```

pedrobaeza (commented):
Being a fw-port, you should mention it in the main comment for doing cross-checks, and this also proves you shouldn't put the Odoo version in the commit message.

florian-dacosta:
> Being a fw-port, you should mention it in the main comment for doing cross-checks, and this also proves you shouldn't put the Odoo version in the commit message.

I did edit to add the original pull request

pedrobaeza (approved):
/ocabot merge patch

PR #979, opened 2026-04-01
[18.0][FIX] partner_multi_company: company_ids propagation to children

Killance:
Currently, when we modify the company_ids field in a parent contact, the modification is applied on children at the next write().

### DIFF
--- a/partner_multi_company/models/res_partner.py
+++ b/partner_multi_company/models/res_partner.py
@@ -32,6 +32,11 @@ class ResPartner(models.Model):
         commercial_fields += ["company_ids"]
         return commercial_fields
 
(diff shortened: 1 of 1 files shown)


BenjaHe (approved):
I check it on the runbot and have the same problem.
The sync of the company_ids on the child_ids are one write later than the parent.
This change fix this strange behavior

carlos-lopez-tecnativa:
A similar solution was proposed in https://github.com/OCA/multi-company/pull/1038, coming from V17.0 with tests. Could you review this please?

pedrobaeza:
@Killance do you agree to switch to the other, which contains tests?

Killance:
@pedrobaeza Yes it's ok

PR #946, opened 2026-02-16
[19.0][MIG] purchase_sale_stock_inter_company_mrp

JasminSForgeFlow:
Standard Migration

@ForgeFlow

Depends on:

- [ ] https://github.com/OCA/multi-company/pull/945

PR #987, opened 2026-04-17
[18.0][MIG] purchase_sale_stock_inter_company_mrp: Migration to 18.0

metaminux:
Standard migration from 16.0.

Supersedes #849 

Strongly inspired by #946 (@JasminSForgeFlow tell me if you want to be mentioned explicitly in CONTRIBUTORS)

cc @Kev-Roche @sebastienbeau


metaminux:
**Fix kit quantity synchronization in intercompany pickings**

The parent module synchronizes quantities between intercompany pickings by iterating over sale moves and zipping sale move lines with purchase move lines one-to-one. This works for standard products but breaks in kit scenarios where the sale and purchase sides have different move structures:

- **Purchase kit** (sale delivers kit product → purchase receives components): the zip stops after the first pair, leaving remaining component move lines unprocessed. Worse, the parent's `move_line_diff` logic deletes the extra component move lines before the zip even runs.
- **Sale kit** (sale delivers components → purchase receives kit product): `_compute_kit_quantities_done` was receiving only the current move instead of all component moves of the kit, always returning `0.0`.
- **Sale+purchase kit** (both sides use different kit decompositions): combination of both issues above.

**Fix:**

- Add a `_is_intercompany_purchase_kit_move(move, dest_picking)` hook in `purchase_sale_stock_inter_company` (returns `False` by default) so that the main sync loop can skip moves that require kit-specific handling.
- Override `_action_done_intercompany_actions` in `purchase_sale_stock_inter_company_mrp` to pre-process purchase kit moves before calling `super()`: moves are grouped by `sale_line_id`, kit quantities are computed or exploded as needed, and all destination component move lines are written in a single pass.
- Simplify `_get_product_intercompany_qty_done_dict` in the MRP module to handle only the sale kit case, fetching all kit component moves from `self.move_ids` rather than relying solely on the current zip pair.

Now that tests are all green, can someone review this PR ? @Kev-Roche @carlos-lopez-tecnativa @JordiMForgeFlow

PR #1012, opened 2026-06-10
[18.0][IMP] purchase_sale_inter_company: Allow canceling the purchase order…

carlos-lopez-tecnativa:
… when the sales order is automatically confirmed

This module adds an option to automatically confirm the sales order generated from a purchase order. When this option is enabled, attempting to cancel the purchase order triggers a validation error. However, this module also supports automatically canceling the related sales order when the purchase order is canceled. Therefore, this restriction is contradictory and has been removed.

Additionally, the `disable_cancel_warning` context key is injected to cancel the sale order directly instead of displaying the confirmation wizard.

TT62901
@Tecnativa @pedrobaeza @carlosdauden @christian-ramos-tecnativa @victoralmau could you please review this?

### DIFF
--- a/purchase_sale_inter_company/models/purchase_order.py
+++ b/purchase_sale_inter_company/models/purchase_order.py
@@ -205,11 +205,8 @@ class PurchaseOrder(models.Model):
             .search([("auto_purchase_order_id", "in", self.ids)])
         )
         for so in sale_orders:
-            if so.state not in ["draft", "sent", "cancel"]:
-                raise UserError(
-                    self.env._("You can't cancel an order that is %s", so.state)
-                )
-        for so in sale_orders:
-            so.action_cancel()
+            # inject the disable_cancel_warning context key
+            # to cancel directly instead of displaying the wizard.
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Is this working also on non interactive flows?

carlosdauden (approved):

carlos-lopez-tecnativa:
> Is this working also on non interactive flows?

Sorry, but I don't completely understand what you mean by "non-interactive flows".

If you cancel it from code, then yes: the Purchase Order method is `button_cancel`. It does not have an `_action_cancel` and `action_cancel` method or anything similar, as Sales Orders do.

Before this commit, this does not cancel the Sales Order directly because it calls `action_cancel`(), which tries to display a wizard. For that reason, I added the `disable_cancel_warning` context, as explained in the PR description.

If I misunderstood your comment, could you please provide more details about what you are referring to?

pedrobaeza (approved):
That's exactly what I meant: don't be interrupted by a wizard, so then it's OK.

/ocabot merge patch

PR #1009, opened 2026-06-03
[19.0][MIG] base_multi_company

EmilioPascual:
Supersed of https://github.com/OCA/multi-company/pull/889

I have re-run the migration from the current version 18. The old pull request was quite old and was causing numerous inconsistencies with the latest changes.
I have tried to preserve the commit authorship as far as possible.

@Vandan-Pandeji-SerpentCS, @astirpe, @weinni2000, @Murtaza-OSI, @chienandalu, @yajo please review

@moduon MT-14423

PR #917, opened 2025-12-11
[19.0][MIG] product_multi_company: Migration to 19.0

bizzappdev:
Dependency PR for module `base_multi_company`: https://github.com/OCA/multi-company/pull/889

PR #1014, opened 2026-06-19
[19.0][MIG] product_multi_company

EmilioPascual:
Supersed https://github.com/OCA/multi-company/pull/917 whilst preserving the original commits. There has been no response from the author following several requests.

Depends on:
- [x] https://github.com/OCA/multi-company/pull/1009

@fcvalgar @Gelojr @chienandalu @bizzappdev @weinni2000 please review, thank you.

MT-14467 @moduon


chienandalu (changes requested):
There's something wrong with the base

Image

EmilioPascual:
> There's something wrong with the base
> 
> Image

Fixed! Thanks for the warning.

fcvalgar (changes requested):
Hi @EmilioPascual, thanks for the migration work on this module. I found an important functional case that I think should be reviewed before approval.

When a product is initially shared across all companies, with no company restriction, and it has already been used in sales orders from different companies, restricting that product afterwards to only one company can cause errors when accessing historical documents from the other companies.

From a functional point of view, I think the ideal behavior would be that this restriction can still be applied without blocking access to past documents. Historical sales orders, invoices, pickings or other existing records should remain readable and usable, even if the product is no longer available for new operations in that company.

As an alternative minimum safeguard, the system could show a clear warning and prevent the change, for example indicating that the product is already used in documents from other companies and therefore cannot be restricted in a way that would exclude them. However, I think this should be considered a fallback option, because ideally the company restriction should affect product availability for new usage without breaking existing historical records.

Test 1: OK - Shared product without company restriction is visible from multiple companies.
Test 2: OK - Product restricted to one company is only visible from that company.
Test 3: OK - Product restricted to two selected companies is visible only from those companies.
Test 4: OK - Product allowed for the active company can be used in a sales order.
Test 5: OK - Product restricted to another company is not available in a new sales order.
Test 6: OK - Updating the allowed companies changes the product visibility as expected.
Test 7: OK - Product template and variants follow the same multi-company restriction.
Test 8: OK - Existing demo product can be restricted to a specific company.
Test 9: OK - Removing all company restrictions makes the product shared again.
Test 10: OK - Product list respects the user company access rules.
Test 11: Not OK - A product initially shared and already used in sales orders from multiple companies causes an error after being restricted to only one company and then accessing historical documents from an excluded company.

Also, there seems to be a small typo in the PR title: `product_multi_coimpany` should probably be `product_multi_company`.

Could you please review this case? Once adjusted, I will be happy to validate the PR again.

Image

Image

EmilioPascual:
> Test 11: Not OK - A product initially shared and already used in sales orders from multiple companies causes an error after being restricted to only one company and then accessing historical documents from an excluded company.

I think the current approach is appropriate, as if a product has been restricted to one company, it is normal for other companies not to be able to view the documents containing that product.

That said, I do wonder whether it makes sense to revoke a company’s access to a product that it has already used.

fcvalgar (approved):
Hi @EmilioPascual, thanks for the explanation.

I agree with your point. If the product is explicitly restricted to one company, it makes sense that other companies should no longer be able to access documents containing that product. The doubtful part is indeed more about whether it makes functional sense to revoke access to a product that has already been used by another company, but I agree this is aligned with the current approach of the module.

LGTM, thanks for the work.

chienandalu (approved):

rafaelbn:
Please @OCA/intercompany-maintainers could you /ocabot migration product_multi_company

We are running in production this module and dependencies

Gelojr (approved):
Great work on this migration, everything looks good from a functional point of view.
The following tests have been performed:

1. Shared product across all companies: A product without assigned companies was created and correctly remained visible from all companies. OK
2. Product restricted to one company: A product assigned only to one company was visible from that company and not available from another one. OK
3. Product shared between two companies: A product assigned to two companies was correctly visible and editable from both of them. OK
4. Product creation with an unauthorized company: A user tried to create a product assigned to a company they do not have access to, and Odoo correctly raised the expected multi-company access error. OK
5. Assigned companies update: A product initially assigned to one company was updated to include another company, and it became available for the newly added company. OK
6. Product made global again: All assigned companies were removed from a product, and it correctly became visible from all companies. OK
7. Product variants behavior: It is possible to assign companies from a product variant. However, adding or removing a company on a single variant automatically updates the company assignment for all variants of the same product. OK. **Note: Please confirm whether this behavior is the expected functional design of the module.**
8. Product visible in two companies but not in a third one: A product assigned to two companies was visible from both of them and not visible from a third company not included in the selection. OK

BhaveshHeliconia (approved):
Functional review LGTM!

test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-addon-base_multi_company @ git+https://github.com/OCA/multi-company.git@refs/pull/889/head#subdirectory=base_multi_company
```
LoisRForgeFlow: @EmilioPascual can you remove the "DON'T MERGE commit"?

LoisRForgeFlow (changes requested):

weinni2000 (approved):
LGTM

test-requirements.txt
EmilioPascual: Done! Thank you for letting me know.

LoisRForgeFlow (approved):
Thanks! let's move forward 🚀 

/ocabot merge nobump

PR #1020, opened 2026-06-25
[ADD] web_switch_company_favorite + copier update

petrus-v:
This module enhances Odoo's multi-company switcher to quickly switch
between different company combinations using filters.

Switching using a custom filter:

https://github.com/user-attachments/assets/b7206c55-950f-4a14-a994-93e69bf766eb

Create a new custom filter

https://github.com/user-attachments/assets/2792b558-bca7-477c-9b25-96581d741024


legalsylvain:
nice video preview. do you know current alternative ? https://github.com/OCA/web/tree/16.0/web_responsive_company#usage (not the same, but improve UX).

petrus-v:
> nice video preview. do you know current alternative ? https://github.com/OCA/web/tree/16.0/web_responsive_company#usage (not the same, but improve UX).

I didn't know, thanks for the hint, scope seems a bit different.

First in v19.0 (I don't know about v18),  the search is standard if you have more than 10 companies (a bit surprised to me this happen in javascript side

This do not depends on web responsive.

The main focus here is to be able to use ir.filter to pre select a set of companies that you have the chance to change before apply.

web_switch_company_favorite/static/src/xml/switch_company_menu.xml
```diff
@@ -0,0 +1,153 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+<!-- Copyright 2026 ACSONE SA/NV
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+     @author Pierre Verkest <pierre@verkest.fr> -->
+
+<t t-inherit="web.SwitchCompanyMenu" t-inherit-mode="extension" owl="1">
```
AnizR: I think that the attribute `owl` isn't necessary anymore

```suggestion
<t t-inherit="web.SwitchCompanyMenu" t-inherit-mode="extension">
```

AnizR (approved):
Great feature!

Code LGTM, I have one small comment.

web_switch_company_favorite/static/src/xml/switch_company_menu.xml
AnizR: Sorry but there are multiple occurences of `owl="1"` in your .xml :grimacing: 

For reference, here is why it isn't necessary anymore:https://github.com/odoo/odoo/pull/130467

petrus-v: Sorry I click in hurry !

Thanks for your attentive review

PR #1021, opened 2026-06-29
[ADD] product_pricelist_multi_company

tobiaszehntner:
AI (Claude 4.6) used for conceptualisation and writing of the files.


product_pricelist_multi_company/security/product_pricelist_security.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- Copyright 2026 ACSONE SA/NV -->
+<!-- License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="product.product_pricelist_comp_rule" model="ir.rule">
+        <field name="domain_force">
+            ['|', '|', ('company_id', 'parent_of', company_ids), ('company_id', '=', False), ('company_ids', 'in', company_ids)]
```
acsonefho, line 16: The best is maybe to create a new `ir.rule`.
As the original rule is `noupdate="1"` this update won't be applied.
And if another module (or a user) update it, you drop custom modifications.

As rules are merged with `AND`, you probably have to adapt the logic to make it compatible with existing rules.

product_pricelist_multi_company/security/product_pricelist_security.xml
```diff
@@ -0,0 +1,15 @@
...
+        </field>
+    </record>
+    <record id="product.product_pricelist_item_comp_rule" model="ir.rule">
+        <field name="domain_force">
+            ['|', '|', ('company_id', 'parent_of', company_ids), ('company_id', '=', False), ('company_ids', 'in', company_ids)]
```
acsonefho, line 23: Same here

product_pricelist_multi_company_website_sale/security/product_pricelist_security.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- Copyright 2026 ACSONE SA/NV -->
+<!-- License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="website_sale.product_pricelist_comp_rule" model="ir.rule">
+        <field name="domain_force">
+            ['|', '|', ('company_id', 'in', [False, website.company_id.id]), ('company_id', 'in', company_ids), ('company_ids', 'in', company_ids)]
```
acsonefho: Same here

product_pricelist_multi_company_website_sale/hooks.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2026 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+_WEBSITE_SALE_DOMAIN = (
+    "['|', ('company_id', 'in', [False, website.company_id.id]),"
```
acsonefho: Good idea but if another module (or the user) change it, you can not restore it with the previous value

acsonefho (changes requested):

tobiaszehntner:
@acsonefho I went with the strategy that Odoo uses in website_sale: archive existing rules so the new ones can be taken into account. An uninstall hook unarchives the existing rules.

PR #911, opened 2025-11-27
[18.0][MIG] product_account_multicompany_default: Migration to 18.

u0f:
Migration to version 18:
- Adapt code to the v18 multi-company
- Create missing info on tests
- README improve
- Inherit change on [product category view](https://github.com/OCA/multi-company/commit/f6508aed1f114d6c232ad97a3d9876656a3a505e#diff-8a130999e62532b052be25a438a4f14e5a38e4fe5d484cae9f91d2928d397debR6) due to a conflict with a odoo inherit in the same field.

https://www.loom.com/share/0e676588f7cc45109a23dc49f158de11

MT-12710 @moduon @yajo

PR #1026, opened 2026-07-02
[19.0][ADD] product_company_default: scope new products to the active company

mikecolangelo:
## New module: `product_company_default`

`product_multi_company` leaves new products global by default (empty `company_ids` = visible to every company). For a strict per-company setup that is the wrong default: a merchant creating a product should get it scoped to their own company without having to touch (or even see) the multi-company field. This is the product-side counterpart of `partner_company_default` (OCA/partner-contact).

Implementation notes:

- `company_ids` is set directly in `create()` instead of putting a default on `company_id`: `company_id` is a computed/inverse field on `multi.company.abstract`, and letting its inverse run during create triggers an early write that the company record rule rejects (see also OCA/partner-contact#2401 for the same pitfall on the partner side).
- An explicit company choice is respected — interpreted from the actual commands, not plain truthiness, so an emptying command such as `Command.set([])` only produces a deliberately global product for users holding `base.group_multi_company`; for everyone else (the field is hidden from them; such a value is only reachable through RPC/imports) the default scoping applies.

## Status

**Draft**: depends on `product_multi_company`, whose 19.0 migration is not merged yet (#911-family). Will be marked ready once the dependency lands. 4 tests, pre-commit green; validated live in a multi-company deployment sandbox together with the rest of the isolation stack.


AungKoKoLin1997:
@mikecolangelo Your proposed module is already exist in this repo.
https://github.com/OCA/product-attribute/tree/19.0/product_company_default

mikecolangelo:
Oh, i got it, thanks and sorry @AungKoKoLin1997

PR #1027, opened 2026-07-06
[18.0][FIX] purchase_sale_stock_inter_company: skip cancelled moves when syncing pickings

sergio-teruel:
When validating an intercompany delivery, the sync logic iterated over all
stock moves, including cancelled ones.

This caused a validation error after replacing products in a picking: the
cancelled original move still tried to find a pending corresponding PO move,
but its purchase-side move had already been cancelled.

Skip cancelled moves during the intercompany picking sync, so only active
delivery moves propagate quantities/lots to the destination receipt.

cc @Tecnativa TT63574

ping @carlos-lopez-tecnativa @carlosdauden @victoralmau

### DIFF
--- a/purchase_sale_stock_inter_company/models/stock_picking.py
+++ b/purchase_sale_stock_inter_company/models/stock_picking.py
@@ -97,7 +97,7 @@ class StockPicking(models.Model):
             dest_picking = self.intercompany_picking_id.with_user(
                 intercompany_user
             ).with_company(dest_company)
-            for move in self.move_ids:
+            for move in self.move_ids.filtered(lambda m: m.state != "cancel"):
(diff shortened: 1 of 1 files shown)


carlos-lopez-tecnativa (commented):
The change makes sense to me, but how can a user cancel an individual stock.move without canceling the entire picking?

carlos-lopez-tecnativa (approved):
I answered my own question. This could happen when using a module that allows changing products, or another custom module. Anyway, LGTM.

sergio-teruel:
Thanks @carlos-lopez-tecnativa

carlosdauden (approved):

PR #971, opened 2026-03-13
[16.0] [FIX] purchase_sale_inter_company: take into account sale order discounts

MarinaAForgeFlow:
Supersedes https://github.com/OCA/multi-company/pull/662

Currently, when a sale order line has a discount, this discount is not taken into account when updating the purchase order line price. Instead the sale order line unit price is passed to the purchase line unit price, causing an error in the valuation.

PR #1034, opened 2026-07-22
[FW][FIX] purchase_sale_inter_company: take into account sale order discounts

MarinaAForgeFlow:
Currently, when a sale order line has a discount, this discount is not taken into account when updating the purchase order line price. Instead the sale order line unit price is passed to the purchase line unit price, causing an error in the valuation.

Unlike the 17.0/19.0 ports, 18.0 already refactored the price sync into _sync_price_unit_to_intercompany_purchase_line (which also handles cross-company tax include/exclude). That method already serves as the inheritable override point, so the discount is folded into it rather than introducing a separate method. The tax-crossing branches use raw_total_*_currency, which already account for the discount, so only the base case needed the change.

### DIFF
--- a/purchase_sale_inter_company/models/sale_order_line.py
+++ b/purchase_sale_inter_company/models/sale_order_line.py
@@ -22,10 +22,11 @@ class SaleOrderLine(models.Model):
         of the sale order line to the price unit
         of the purchase order line in case of intercompany sale.
         It takes into account the tax inclusion/exclusion
-        of both the sale order line and the purchase order line.
+        of both the sale order line and the purchase order line,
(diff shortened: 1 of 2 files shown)


JordiMForgeFlow (approved):

AaronHForgeFlow (approved):
👍

MiquelRForgeFlow (approved):

LoisRForgeFlow:
For reference, origina PR https://github.com/OCA/multi-company/pull/971

LoisRForgeFlow (approved):
/ocabot merge patch

PR #1035, opened 2026-07-27
[19.0][FIX] base_multi_company: initialize company on create

chienandalu:
Avoid triggering the inverse write of company_id after creating a record. The intermediate record has no company_ids yet, which can make company-dependent record rules reject the write for users creating products.

cc @moduon MT-15413

please take a look @EmilioPascual @yajo

### DIFF
--- a/base_multi_company/models/multi_company_abstract.py
+++ b/base_multi_company/models/multi_company_abstract.py
@@ -93,6 +93,14 @@ class MultiCompanyAbstract(models.AbstractModel):
     def create(self, vals_list):
         """Discard changes in company_id field if company_ids has been given."""
         for vals in vals_list:
+            # Avoid triggering the inverse write of company_id after creating a record.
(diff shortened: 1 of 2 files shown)


base_multi_company/models/multi_company_abstract.py
```diff
@@ -93,6 +93,10 @@ def _multicompany_patch_vals(self, vals):
     def create(self, vals_list):
         """Discard changes in company_id field if company_ids has been given."""
         for vals in vals_list:
+            if "company_id" in vals and "company_ids" not in vals:
+                company_id = vals.pop("company_id")
```
pedrobaeza, line 101: Put here the comment about why doing this.

chienandalu: Done

pedrobaeza (approved):

EmilioPascual (approved):

PR #1037, opened 2026-07-27
[17.0][FIX] partner_multi_company: Prevent cache pollution when changing company_ids on parent partner

carlos-lopez-tecnativa:
Before this fix, modifying the `company_ids` field on a parent partner and propagating the change to its children caused cache pollution. Children would incorrectly retain outdated company values.

**Steps to reproduce:**
1. Create a partner with 2 companies.
2. Create a child of that partner.
3. Remove a company from the parent, the child still preserves the old companies.
4. Add a new company to the parent, the child still preserves the old companies.

**After this commit:**
The companies on the children are now synchronized correctly with the parent.

TT63857
@Tecnativa @pedrobaeza @christian-ramos-tecnativa @pilarvargas-tecnativa @victoralmau could you please review this?

### DIFF
--- a/partner_multi_company/models/res_partner.py
+++ b/partner_multi_company/models/res_partner.py
@@ -32,6 +32,17 @@ class ResPartner(models.Model):
         commercial_fields += ["company_ids"]
         return commercial_fields
 
+    def _children_sync(self, values):
+        if "company_ids" in values:
+            # When the transaction is performed in sudo mode,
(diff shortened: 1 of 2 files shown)


christian-ramos-tecnativa (approved):
LGTM, does this happen in later versions?

pilarvargas-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #901, opened 2025-11-04
[MIG][19.0] partner_multi_company

Vandan-Pandeji-SerpentCS:
Depending Module: https://github.com/OCA/multi-company/pull/889

PR #1038, opened 2026-07-27
[18.0][FIX] partner_multi_company: Prevent cache pollution when changing company_ids on parent partner

carlos-lopez-tecnativa:
FWD-Port of https://github.com/OCA/multi-company/pull/1037
Before this fix, modifying the `company_ids` field on a parent partner and propagating the change to its children caused cache pollution. Children would incorrectly retain outdated company values.

**Steps to reproduce:**
1. Create a partner with 2 companies.
2. Create a child of that partner.
3. Remove a company from the parent — the child still preserves the old companies.
4. Add a new company to the parent — the child still preserves the old companies.

**After this commit:**
The companies on the children are now synchronized correctly with the parent.

TT63857
@Tecnativa @pedrobaeza @christian-ramos-tecnativa @pilarvargas-tecnativa @victoralmau could you please review this?

### DIFF
--- a/partner_multi_company/models/res_partner.py
+++ b/partner_multi_company/models/res_partner.py
@@ -32,6 +32,17 @@ class ResPartner(models.Model):
         commercial_fields += ["company_ids"]
         return commercial_fields
 
+    def _children_sync(self, values):
+        if "company_ids" in values:
+            # When the transaction is performed in sudo mode,
+            # see https://github.com/odoo/odoo/blob/629a563148452091bd1af9267596859b140350bb/odoo/addons/base/models/res_partner.py#L689
+            # the cache becomes inconsistent for the many2many field,
+            # and the child partners are not updated with the new company_ids value.
+            # Instead, they retain the previous value.
+            # Therefore, we need to invalidate the cache to force the update.
+            self.invalidate_recordset(["company_ids"], True)
+        return super()._children_sync(values)
+
     @api.model
     def _amend_company_id(self, vals):
         if "company_ids" in vals:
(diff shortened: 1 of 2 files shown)


partner_multi_company/models/res_partner.py
```diff
@@ -32,6 +32,17 @@ def _commercial_fields(self):
...
 
+    def _children_sync(self, values):
+        if "company_ids" in values:
+            # When the transaction is performed in sudo mode,
+            # see https://github.com/odoo/odoo/blob/629a563148452091bd1af9267596859b140350bb/odoo/addons/base/models/res_partner.py#L689
+            # the cache becomes inconsistent for the many2many field,
+            # and the child partners are not updated with the new company_ids value.
+            # Instead, they retain the previous value.
+            # Therefore, we need to invalidate the cache to force the update.
+            self.invalidate_recordset(["company_ids"], True)
```
pedrobaeza, line 43: Have you tried the same test without applying the patch if it still fails?

carlos-lopez-tecnativa: Yes @pedrobaeza, I tested it in runboat and the error exists. I also tested it in V19 on this instance http://oca-multi-company-19-0-pr901-7f6ed4875049.runboat.odoo-community.org/odoo/contacts/9 from this PR https://github.com/OCA/multi-company/pull/901

pedrobaeza (approved):

christian-ramos-tecnativa (approved):
LGTM but we should wait for the response in https://github.com/OCA/multi-company/pull/979

victoralmau (approved):

BenjaHe (approved):
Same issue and this PR resolve it

PR #1048, opened 2026-08-24
[18.0][IMP] account_invoice_inter_company: Set journal and company explicitly

carlos-lopez-tecnativa:
In strange cases, when a customer invoice is validated and the system tries to create a vendor bill, the user gets the error: `Cannot create a purchase document in a non-purchase journal`

Before this commit, Odoo computed the company and journal. Instead, we set these fields explicitly to prevent them from being recomputed.

TT64262
@Tecnativa @pedrobaeza could you please review this?

### DIFF
--- a/account_invoice_inter_company/models/account_move.py
+++ b/account_invoice_inter_company/models/account_move.py
@@ -142,6 +142,7 @@ class AccountMove(models.Model):
                     "dest_company_id": dest_company.id,
                 }
             )
+        return dest_journal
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

carlos-lopez-tecnativa:
@carlosdauden @sergio-teruel @victoralmau could you please review this?

carlosdauden (approved):

victoralmau (approved):

PR #709, opened 2024-10-23
[18.0][MIG] company_dependent_flag: Migration to 18.0

BertVGroenendael:

PR #1052, opened 2026-09-03
[MIG] company_dependent_flag: Migration to 18.0

SirPyTech:
Migration with rename from `16.0` (https://github.com/OCA/multi-company/tree/627705784e77c702ac6d56126f625052aca029e3/company_dependent_flag).

Supersede https://github.com/OCA/multi-company/pull/709 because the author is inactive from a while.

@legalsylvain @sergiocorato @PaulGoubert @LoisRForgeFlow @bealdav you worked on the linked PR, maybe you'd like to have a look?

I removed the demo Python file because the field `barcode` is already `company_dependent` in module `base`: https://github.com/odoo/odoo/blob/21340b235f461bf43e6ec0b234a41502b4f5dd9c/odoo/addons/base/models/res_partner.py#L311


SirPyTech:
@lmarion-source I see you are working on the migration to `19.0`, would you like to have a look too?

legalsylvain:
Hi. thanks for migrating this module.

Could you provide a fix regarding the place of the icon that is not in the label, but in the "value" space, and without any space between the icon and the value ?

image

legalsylvain:
this should fix another problem, that the font of the value is changed, and is "fontAwsome", that should not be the case.


image

MiquelRForgeFlow (approved):

legalsylvain (changes requested):
see comment.

SirPyTech:
@legalsylvain I added a new commit that should fix what you reported in https://github.com/OCA/multi-company/pull/1052#issuecomment-5527917203 and https://github.com/OCA/multi-company/pull/1052#issuecomment-5527940118, please check

web_field_company_dependent_flag/static/src/form_label.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+    ~ Copyright 2026 Simone Rubino - PyTech
+    ~ License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+-->
+<templates xml:space="preserve">
+    <t id="FormLabel" t-inherit="web.FormLabel" t-inherit-mode="extension">
+        <xpath expr="//sup[@t-if='hasTooltip']" position="after">
+            <i t-if="CompanyDependentClasses" t-att-class="CompanyDependentClasses" />
```
legalsylvain:
```suggestion
            <i t-if="CompanyDependentClasses" t-att-class="CompanyDependentClasses"
            title="Values set here are company-specific." />
```

Proposal / not blocking : As per in https://github.com/odoo/odoo/blob/18.0/addons/web/static/src/views/form/setting/setting.xml#L19

Don't you think ?

legalsylvain (approved):
Thanks for the patch ! 
(tested on runboat, works like a charm).

web_field_company_dependent_flag/static/src/form_label.xml
legalsylvain: In fact, I cherry-picked your fix, and I added some other fixes you can take a look here https://github.com/OCA/multi-company/pull/1053 (and cherry-pick back ! if you want).
title works well in V16.

SirPyTech: I added the title and the corresponding accessible label :+1: please check.

I'll have a look at https://github.com/OCA/multi-company/pull/1053 and let you know there.

carlos-lopez-tecnativa (changes requested):
The rename is not completely correct. If you review the first commit, it still preserves the old name. You should rewrite the entire commit history; otherwise, in the next migration, only the commits with the new name will be included.

I use these commands in such cases:

```
git filter-branch --tree-filter 'if [ -d old_module_name ]; then mv old_module_name new_module_name; fi' HEAD
git rebase origin/branch
```

lmarion-source:
> @lmarion-source I see you are working on the migration to `19.0`, would you like to have a look too?

@AnizR maybe this thread needs to be looked at 👀 maybe some things to retrieve in v19.0

SirPyTech:
> The rename is not completely correct. If you review the first commit, it still preserves the old name. You should rewrite the entire commit history; otherwise, in the next migration, only the commits with the new name will be included.
> 
> I use these commands in such cases:
> 
> ```
> git filter-branch --tree-filter 'if [ -d old_module_name ]; then mv old_module_name new_module_name; fi' HEAD
> git rebase origin/branch
> ```

Hi @carlos-lopez-tecnativa, thank you for having a look!
I know that in the next migration only the commits with the new name will be picked up, but I'd rather accept that then mangle with the commit history more than we already do.
We are already squashing the bot commits in users' commits, it's fine to keep the history clean, but it already makes it harder to check that the history is preserved in a migration.

I have recently noticed the "renamed module" clause in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0, I'd like to know if there has been a discussion to better understand this new guideline.

Moreover, the WARNING at the beginning is not encouraging:
> git filter-branch --tree-filter 'if [ -d company_dependent_flag ]; then mv $module web_field_company_dependent_flag; fi' HEAD
> git rebase origin/18.0
> WARNING: git-filter-branch has a glut of gotchas generating mangled history
> 	 rewrites.  Hit Ctrl-C before proceeding to abort, then use an
> 	 alternative filtering tool such as 'git filter-repo'
> 	 (https://github.com/newren/git-filter-repo/) instead.  See the
> 	 filter-branch manual page for more details; to squelch this warning,
> 	 set FILTER_BRANCH_SQUELCH_WARNING=1.
> Proceeding with filter-branch...

carlos-lopez-tecnativa:
> Hi @carlos-lopez-tecnativa, thank you for having a look!
> I know that in the next migration only the commits with the new name will be picked up, but I'd rather accept that then mangle with the commit history more than we already do.

But the correct approach is for all commits in the history to be carried over to the next version, not only the commits after the module name was changed. This is my point.

> We are already squashing the bot commits in users' commits, it's fine to keep the history clean, but it already makes it harder to check that the history is preserved in a migration.

Squashing the administrative commits according to [this guide](https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests) is a normal process during a migration.

If the command I suggested is obsolete, feel free to use another command or tool, but the point is to rework the entire Git commit history.

SirPyTech:
> > Hi @carlos-lopez-tecnativa, thank you for having a look!
> > I know that in the next migration only the commits with the new name will be picked up, but I'd rather accept that then mangle with the commit history more than we already do.
> 
> But the correct approach is for all commits in the history to be carried over to the next version, not only the commits after the module name was changed. This is my point.

I see your point, my point is about preserving the commit history: I think they are both valid, that's why I asked if there has been a discussion to better understand the new guideline.
Moreover, editing the commits by renaming the module would break them: for instance if you checkout a commit in that edited history any XMLID reference that mentions the old module name wouldn't work anymore.

carlos-lopez-tecnativa:
> > > Hi @carlos-lopez-tecnativa, thank you for having a look!
> > > I know that in the next migration only the commits with the new name will be picked up, but I'd rather accept that then mangle with the commit history more than we already do.
> > 
> > 
> > But the correct approach is for all commits in the history to be carried over to the next version, not only the commits after the module name was changed. This is my point.

> I see your point, my point is about preserving the commit history: I think they are both valid, that's why I asked if there has been a discussion to better understand the new guideline. 

Perhaps @pedrobaeza could provide some additional insight on this.

> Moreover, editing the commits by renaming the module would break them: for instance if you checkout a commit in that edited history any XMLID reference that mentions the old module name wouldn't work anymore.


You don't need to rename the module and XML IDs in all commits. For example:

C1 — Module name: MyModule
C2 — Module name: MyModule
C3 — Module name: MyModule
C4 — Module name: MyNewModule
C5 — Module name: MyNewModule

From the point where you rename the module (ideally in the latest commit or in the migration commit), you change the XML IDs and make all the necessary changes. But from C1 to C3, the code does not change. It is only a Git-level process, and it does not break anything. It only helps ensure that all commits are taken into account in the next migration.

pedrobaeza:
@SirPyTech the whole renaming is done because if not, on 19, you won't get the whole history with the usual method, so this renaming is done preventively at the renaming moment for such purpose, but as Carlos has said, it's only the directory renaming, not the resources inside. That's done on an additional renaming commit once.

SirPyTech:
Thanks @pedrobaeza and @carlos-lopez-tecnativa for the clarifications.
I understand what the procedure does, I have done that procedure a few times already on my own for other purposes.

What I meant is that it might be breaking the edited commits: for example if in a renamed commit the Python code was referencing an external ID `company_dependent_flag.some_resource` and in that commit the module is named `web_field_company_dependent_flag` than the resource would not be found.
If you checkout or inspect that renamed commit, the module does not work nor the code in the commit makes sense because you might think `company_dependent_flag` is a different module.

So we would be carrying a lot of broken commits that can be confusing.
Squashing consecutive commits does not break anything, but this procedure does.

Has been there a discussion about this new guideline? Maybe someone else already highlighted this and this has already been discussed.

pedrobaeza:
This is not a new guideline, but for not breaking the existing one of preserving the whole commit history. About broken commits, that something not real in practice, as you are not checking out a commit before the renaming for debugging anything, counting that this is done on a migration, so any commit prior to the migration won't probably work anyway.

SirPyTech:
Yes no-one executes them for debugging, but they can be used to inspect how the code was, and if you find `company_dependent_flag.some_resource` in a module named `web_field_company_dependent_flag` you might think that's another module.

The point is that the commit history should be preserved as much as possible.
If you want to see the history of how the module evolved in an old version, you could checkout the module in that version's branch.

Moreover: Github allows to see the history of renamed files (not directories/modules), see for instance: image
(from https://github.com/OCA/l10n-italy/commits/18.0/l10n_it_riba_oca/__manifest__.py)

pedrobaeza:
Please, don't theorize. Is there any XML-ID reference right now in this module to prevent you to do it?

SirPyTech:
> Please, don't theorize. Is there any XML-ID reference right now in this module to prevent you to do it?

No there isn't.
I'm theorizing because the procedure is suggested for any module, independently from their content.

Is this massive history rewrite required for merge?

pedrobaeza:
Yes for 2 of the current reviewers.

legalsylvain:
Hi. 

Hum. Annoying point : 

- Well, I understand the point of @pedrobaeza. indeed, history can be lost in the next migration process. (but maybe, it's a problem of [our "usual method"](https://github.com/OCA/multi-company/pull/1052#issuecomment-5621769398)).
- In another hand, I'm agree with @SirPyTech : rewriting history, to write wrong things in the past should be avoided. That is absolutely not good practice, and it would be the first time I’ve seen that in the OCA. (The only manipulation of past history allowed for the time in the OCA is merging some commits. (like weblate commits)). 

**As :**
- the workaround proposed by @pedrobaeza is not correct in a git history point of view
- [as said](https://github.com/OCA/multi-company/pull/1052#issuecomment-5661555810) by @pedrobaeza, it is not an OCA guideline
- the current method done by @SirPyTech has been done for most of other OCA modules renamed

**Conclusion :** 
 this point can not be blocking for merging the current PR.


@pedrobaeza : the point you raise is interesting. Could you open an issue in a place where we can all talk about that topic and compare solution (and related problems) ? 

thanks !

pedrobaeza:
Sorry, but it's the proper method to comply with OCA guidelines. I'm not going to open anything, because I have it clear. You are doing damage to the initial contributors of this module when in 19, their commit history will be lost. I'm not using nor contribute to this module in the past, so I'm not affected. I'm just trying to be fair with all in behalf of OCA, but if you want to make me lose more time, just go ahead with what you want and give explanations to the previous contributors of the module, like @bealdav

legalsylvain:
> Sorry, but it's the proper method to comply with OCA guidelines. 

The main guideline I refer is that one : https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

* Up to 17.0 there is no reference to module renaming procedure
* from 18.0, there is a section that has been written a year ago  (Jul 2, 2025) : https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0/406a6d0e9b2e050e83156229cc1edb2d017b8bfb
 
@pedrobaeza  did you talked with other OCA maintainers, before adding this section ? If yes, where ? thanks !  

@StefanRijnhart, @simahawk : I saw you co-written this documentation. What's your point of view on that topic, and the problem raised by @SirPyTech ?


Quoted section : 

- **If the module is renamed** in the migration, you have to rename the whole commit history. If not done, on the next version, when these commands will be executed again, only the latest commit will be brought. For doing the renaming, execute the following commands:
``` 
git filter-branch --tree-filter 'if [ -d $module ]; then mv $module $new_module_name; fi' HEAD
git rebase origin/18.0
```

> but if you want to make me lose more time

Lately, whenever I say something you don’t agree with, you say I’m wasting your time. I find that quite a passive-aggressive attitude. No, my aim isn’t to waste your time; my aim is to move forward with the PR for the projects I’m the PSC for. Could you adopt a more cordial tone? 

Thanks !

pedrobaeza:
Sylvain, my reluctance is because you always charge in my shoulders to do things (like opening an issue), while this hasn't been a problem before, and only theorizing about possible problems. The migration guide are not guidelines. It's a guide, as its name states. If you don't want to follow that method and lose commit history, do what you consider, but I don't want to ~spend~ invest more time in this.

bealdav:
Hi all, and thanks to @SirPyTech for your work.

On my side I don't know why you renamed this module to `web_company_dependent_flag`. !??
On my side I love simple things, then I choose python for that and it's KISS spirit, and name only should describe what is it for, I think. Other consideration can be documented in readme if required.
This module intentionally use python and no js/html boillerplate, then web prefix make it confusing, I suppose
If you stick to the original name, I think you solve all the other problems, isn't it ?

bealdav:
Thanks to contributors on this PR, this module wasn't perfect because of css class positionning particularly. Some pointed problems and other solve a lot of them, it seems at least. Really nice job. And if historic can be saved, that's perfect.

SirPyTech:
> Hi all, and thanks to @SirPyTech for your work.
> 
> On my side I don't know why you renamed this module to `web_company_dependent_flag`. !??

Thank you for joining us :hugs:
I renamed the module because it was mentioned in the roadmap: https://github.com/OCA/multi-company/blob/627705784e77c702ac6d56126f625052aca029e3/company_dependent_flag/readme/ROADMAP.rst?plain=1#L1

> On my side I love simple things, then I choose python for that and it's KISS spirit, and name only should describe what is it for, I think. Other consideration can be documented in readme if required. This module intentionally use python and no js/html boillerplate, then web prefix make it confusing, I suppose If you stick to the original name, I think you solve all the other problems, isn't it ?

That happened because a review (https://github.com/OCA/multi-company/pull/1052#pullrequestreview-5130843656) highlighted some style issues that I could only fix on the JS side and believe me, I don't like JS either.

> Thanks to contributors on this PR, this module wasn't perfect because of css class positionning particularly. Some pointed problems and other solve a lot of them, it seems at least. Really nice job. And if historic can be saved, that's perfect.

You are welcome, I really did like working on this, even if I had to write some JS 😆️
Unfortunately the most simple thing is to not preserve the history before the rename because with the current migration method, the modules are picked up from the previous version based on their folder's name.
Keeping the whole history implies rewriting the whole commit history and that's what I'm trying to avoid.

bealdav:
Hi thanks for your reply

My pragmatic point of view, I hope ;-)

```
git reset --soft HEAD~2
git add .
git commit "your migration"
```

All the work you've done is with your commit / authors !?

=> next module, you have a life ... ;-)

SirPyTech:
> Hi thanks for your reply
> 
> My pragmatic point of view, I hope ;-)
> 
> ```
> git reset --soft HEAD~2
> git add .
> git commit "your migration"
> ```
> 
> All the work you've done is with your commit / authors !?
> 
> => next module, you have a life ... ;-)

The problem is not this migration's history, as you can see it has all the commits of the module.
The problem will be for the migration from `18.0` to `19.0`: there the module `web_field_company_dependent_flag` will only have 2 commits because the previous commits store the module with its previous name `company_dependent_flag`.

bealdav:
But 19.0 PR/commits can also be replayed from your updated ones ?
cc @lmarion-source ?

I'll let you define the best strategy, I'll not block.

Thanks

HekkiMelody (approved):
Code and functional review, LGTM

Image

PR #1053, opened 2026-09-07
[IMP] web_field_company_dependent_flag: Better icon rendering

legalsylvain:
Otherwise the icon is shown within the value space of the field and also influences the font used.

1) backport of https://github.com/OCA/multi-company/pull/1052/commits/8c18c5b139bce7fa7231f3baaf881411c8457aa8 done by @SirPyTech in the migration (and rename of ``web_field_company_dependent_flag``) ``company_dependent_flag`` module

2) also apply some simplification, overloading ``fields_get`` instead of ``_get_view``.
It makes the code more easy.
This implementation is based on the web_field_tooltip implementation. See : https://github.com/OCA/web/blob/16.0/web_field_tooltip/models/base.py#L28

3) apply same display as in odoo core.
- put a margin to avoid to join text and building icon.
- add test to avoid to display the buidling icon, if user is not member of multicompany group
- add fa-lg as in odoo core

### Before

image

### After

image


company_dependent_flag/models/base.py
```diff
@@ -8,25 +8,19 @@ class Base(models.AbstractModel):
...
-    def _get_view(self, view_id=None, view_type="form", **options):
-        arch, view = super()._get_view(view_id, view_type, **options)
-        if view_type == "form":
-            self._update_company_dependent_css(arch)
-        return arch, view
-
-    def _update_company_dependent_css(self, arch):
-        cpny_dep_fields = [
+    def fields_get(self, allfields=None, attributes=None):
+        result = super().fields_get(allfields=allfields, attributes=attributes)
```
SirPyTech, line 12: **chore**: Moving from `_get_view` to `fields_get` should be in another commit, otherwise this is very different from my original commit https://github.com/OCA/multi-company/commit/8c18c5b139bce7fa7231f3baaf881411c8457aa8 (now https://github.com/OCA/multi-company/pull/1052/changes/e6f1f6682ab99692bcb3651a1a898834f24011db)

company_dependent_flag/__manifest__.py
```diff
@@ -7,9 +7,13 @@
     "maintainer": ["bealdav"],
     "website": "https://github.com/OCA/multi-company",
     "category": "web",
-    "depends": [
-        "base",
-    ],
+    "depends": ["web"],
```
SirPyTech, line 10: **question**: Is this really needed? If so, should I have it in https://github.com/OCA/multi-company/pull/1052 too?

company_dependent_flag/models/base.py
```diff
@@ -8,25 +8,22 @@ class Base(models.AbstractModel):
...
-                field.attrib["class"] = " ".join(set(classes))
+        for field_name in result.keys() & set(company_dependent_fields):
+            result[field_name]["company_dependent_css_class"] = " ".join(set(classes))
+        return result
 
     def _get_company_dependent_css_class(self):
         """Inherit to apply your own class"""
 
-        return ["fa", "fa-building-o", "d-flex", "flex-row"]
+        return ["fa", "fa-lg", "fa-building-o", "company_dependent_field_icon"]
```
SirPyTech: **suggestion**: `fa-lg` is increasing the height of the whole row, that does not look good, can we drop it?

SirPyTech: **praise**: The dedicated class will help when we want to further customize the element 👏

SirPyTech (changes requested):
Please keep the cherry-picked commit as similar as possible to the original commit https://github.com/OCA/multi-company/commit/8c18c5b139bce7fa7231f3baaf881411c8457aa8 (now https://github.com/OCA/multi-company/commit/e6f1f6682ab99692bcb3651a1a898834f24011db).

**praise**: Using `fields_get` instead of `_get_view` feels less clunky 👍 

**issue**: Any idea why I can't see the icon in runboat?
Image
http://oca-multi-company-16-0-pr1053-296ccad38c86.runboat.odoo-community.org/web?debug=assets#id=23&cids=1&menu_id=111&action=249&model=product.template&view_type=form

company_dependent_flag/__manifest__.py
legalsylvain: I don't think if this change is necessary but the module does'nt work without web installed : https://github.com/grap/multi-company/blob/16.0-FIX-company_dependent_flag-SLG/company_dependent_flag/static/src/form_label.esm.js#L6-L7

> If so, should I have it in https://github.com/OCA/multi-company/pull/1052 too?

Yes, ideally.

legalsylvain:
> issue: Any idea why I can't see the icon in runboat?

image

is a good reason ! 

I don't know why, but since some days, I have to manually install modules on OCA runboat...

company_dependent_flag/models/base.py
legalsylvain: I didn't saw ! thanks for the review. I removed fa-lg.

legalsylvain:
> Please keep the cherry-picked commit as similar as possible to the original commit https://github.com/OCA/multi-company/commit/8c18c5b139bce7fa7231f3baaf881411c8457aa8 (now https://github.com/OCA/multi-company/commit/e6f1f6682ab99692bcb3651a1a898834f24011db).

Hum. sorry, I don't know exactly how to split a commit in two commit, when there are other commit on top of this one. Is a blocking problem, as this branch is not the "head" one, and this "problem" will be forgotten in more recent version ?

company_dependent_flag/__manifest__.py
SirPyTech: :+1: added

I don't think I have ever seen an installation without `web` though

company_dependent_flag/models/base.py
SirPyTech: :+1:

SirPyTech (changes requested):
> > **issue**: Any idea why I can't see the icon in runboat?
> 
> image
> 
> is a good reason ! 
> 
> I don't know why, but since some days, I have to manually install modules on OCA runboat...

🤦‍♂️ I should have checked, sorry for the noise, I confirm it works correctly when the module is installed: Image
you can use this image for the README so it comes from `16.0` and shows the tooltip 😉
For your info, no module is installed in runboat when there are rebel modules, that is the case for this repository: https://github.com/OCA/multi-company/blob/85379227a5bdc2086fba52855f450efcae7d1d15/.copier-answers.yml#L19-L21.

> > Please keep the cherry-picked commit as similar as possible to the original commit https://github.com/OCA/multi-company/commit/8c18c5b139bce7fa7231f3baaf881411c8457aa8 (now https://github.com/OCA/multi-company/commit/e6f1f6682ab99692bcb3651a1a898834f24011db).
> 
> Hum. sorry, I don't know exactly how to split a commit in two commit, when there are other commit on top of this one. Is a blocking problem, as this branch is not the "head" one, and this "problem" will be forgotten in more recent version ? 

I'm sorry but I'd like the commit I authored to be accurate with what I did so for me it's blocking.
A small change is fine but switching from `_get_view` to `fields_get` was not the purpose of that commit.

You can add yourself as a co-author and keep it as-is, or you can cherry-pick it again, and then apply your commits on top of it: in case of conflicts just accept your code.
For your info, the right way to split the commit in two is: `git rebase`, choose `edit` for that commit, create the new commits as you like, continue the rebase.

PR #1056, opened 2026-09-10
[19.0][ADD] res_company_representative: new module

NICO-SOLUTIONS:
This Odoo module allows you to manage company representatives and their representative roles.

A representative is a contact (res.partner) assigned to a specific company and representative role. This relationship is managed separately from the partner's parent contact, as a contact being associated with a company does not necessarily mean that they represent the company.

The structured representative data can be used by other modules, for example to display representatives and their roles in external reports or documents.

Features:
- Assign contacts as representatives to companies.
- Assign a representative role to each representative.
- Manage representative roles centrally.
- Define the display order of representatives.
- Define representative sequences per company.
- Prevent duplicate representatives for the same company.
- Display representatives and their roles on the company form.
- Allow the same contact to represent multiple companies independently.


NICO-SOLUTIONS:
@CRogos
Would you mind reviewing this one when you have some time?

res_company_representative/data/data.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2026-TODAY NICO SOLUTIONS - ENGINEERING & IT (<https://www.nico-solutions.de>)
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
+    <record id="role_ceo" model="res.company.representative.role">
+        <field name="name">CEO</field>
+    </record>
+    <record id="role_authorized_officer" model="res.company.representative.role">
+        <field name="name">Authorized Officer</field>
+    </record>
```
CRogos, line 10: Maybe add the following roles:
Board (Im deutschen gibt es auch noch die Unterscheidung "Geschäftsführer" vs. "Vorstand")
Chairman of the board

res_company_representative/models/res_company_representative.py
```diff
@@ -0,0 +1,60 @@
...
+            limit=1,
+        )
+
+        return (last_representative.sequence or 0) + 1
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if not vals.get("company_id"):
+                vals["company_id"] = self.env.company.id
```
CRogos, line 55: Is a company_id realy this necessary? I think there are cases where you may want to define one person for a role for all companies. Normaly Odoo works like, "Is there an entry with comapny_id, use this, otherwise fallback to the entry without company_id"

res_company_representative/tests/test_res_company.py
```diff
@@ -0,0 +1,172 @@
...
+        partner2 = self.env["res.partner"].create(
+            {
+                "name": "Test Partner 2",
+            }
+        )
+        role = self.env["res.company.representative.role"].create(
+            {
+                "name": "Owner",
+            }
+        )
```
CRogos: Move this to setup()?
Maybe also company2 and patner2?

CRogos (approved):
LGTM, minor suggestions.

res_company_representative/data/data.xml
NICO-SOLUTIONS: That’s not necessary. You can create custom roles using the module.

res_company_representative/tests/test_res_company.py
NICO-SOLUTIONS: Role owner is shared by multiple tests, so it makes sense to create it in setUpClass(). The other roles are only used in individual tests and can stay local.

res_company_representative/models/res_company_representative.py
NICO-SOLUTIONS: company_id is intentionally company-specific here. Representatives are always defined for a specific company; i don't want global representatives shared across companies.

The same person can still be a representative for multiple companies. In that case, the same partner_id is used in separate representative records, one for each company.

Therefore, if no company_id is provided when creating a representative, the current company is assigned automatically.

NICO-SOLUTIONS:
@smaddlsoft 
Maybe You can have a Look here as well? As this is the next step to include it in 5008...

smaddlsoft (changes requested):
Functionally, LGTM.

But: Could you make more explicit here and in the readme what the purpose/use case/possible application (e.g. reports) is and why the standard way of child contacts and using the field position is not enough, or why you think its better outside of the res.partner form but instead in the way you have proposed?

NICO-SOLUTIONS:
> Functionally, LGTM.
> 
> But: Could you make more explicit here and in the readme what the purpose/use case/possible application (e.g. reports) is and why the standard way of child contacts and using the field position is not enough, or why you think its better outside of the res.partner form but instead in the way you have proposed?

@smaddlsoft 
Thanks for the feedback. I’ll make this a bit more explicit in the README.

Actually, this approach cannot really be modeled properly using res.partner child contacts and the standard position field. One main use case is having structured representative data available (e.g. for external reports/documents), independently of the partner’s parent contact.
The representative is still a res.partner, but being a representative is a relationship to a specific res.company and role — which is why I think it makes sense to keep it on res.company rather than on the partner itself.

This also allows multiple representatives per company and the same partner to represent different companies independently. Pretty much the reason for the separate model in the first place 😉

PR #1057, opened 2026-09-14
[17.0][FIX]product_multi_company: keep company_ids on variant write

LoisRForgeFlow:
The product form's header includes an invisible `company_id` field, used as a trick to be able to set that field from the view. When the web client saves, it sends both `company_id` and `company_ids` in the same write call.

`product.template.write()` already protects against this collision through `_multicompany_patch_vals`, discarding the redundant `company_id` so its inverse doesn't overwrite `company_ids`.

`product.product` doesn't inherit `multi.company.abstract` and lacks this protection, even though it exposes the same fields (related to the template). Saving from the variant form (e.g. reached from the MRP Area Parameters view) silently reverted `company_ids` to a single company.

Backport of #1055 to 17.0.


LoisRForgeFlow:
Backport, fast tracking

/ocabot merge patch

